Issue #5, opened 2016-01-19
error running jackStraw


roryk:
Hi,

I have an issue running the `jackStraw` method:

```
> seurat.data = jackStraw(seurat.data, num.replicate=200, do.print=FALSE)
Error in matrix(x2[ind], nrow = nrow(x), ncol = ncol(x), byrow = TRUE) : 
  non-numeric matrix extent
```

I saved the `seurat.data` object and stuck it up here: https://dl.dropboxusercontent.com/u/2822886/seurat-error.RData

Any ideas about why this is failing? It has worked on other datasets, I'm not sure what is different about this one.

Puriney:
Could you please share these 3 components from your huge Rdata? 

```r
pca.x <- seurat.error@pca.x
save(pca.x, file = <path>)
pca.rot <- seurat.error@pca.rot
save(pca.rot, file = <path>)
sca.dt <- seurat.error@scale.data
save(sca.dt, file = <path>)
```

roryk:
Hi Yun,

Thanks for responding. Here is an RData object with just those three objects saved. It is still large (170 MB). https://dl.dropboxusercontent.com/u/2822886/seurat-small.RData

Puriney:
Thanks for sharing the file. 

**Solution**
If `prop.freq = 0.016` (or any other number greater than 0.015), your call runs. 

**Why error**
The error is because you have not set the `prop.freq` parameter.It is used to determine the number of permuted 'synthetic' null original variables denoted as `s` in original jackstraw paper. In Seurat implementation, s = (number of variable genes) * `prop.freq`.  

**Going deep**
Your specific data yields extreme case for Seurat which is helpful for the package grow-up. Thanks again for reporting issue. 

By default, `prop.freq` will be 0.01. If your data has less than 200 variable genes, the `prop.freq` will be forced to be up a little bit, to 0.015, so that helps user has enough tests to get robust results. Your data is the latter case which has 131 variable genes, therefore 131 * 0.015 = 1.965 ~ 1 gene is ready to be permuted. In this extreme case the accuracy of p-value is going to be maximized while the algorithm is least efficient, referring to original jackstraw paper. Currently public Seurat implementation does not allow this happen but without warning, thus the error happens. 

---

Hope it helps. Excuse me for replying late as I was busy making :snowman: these days.

fanli-gcb:
Does anyone also get this error running jackStrawPlot?
```
> jackStrawPlot(srt, PCs = 1:12)
[...]

PR #8, opened 2016-02-14
New imputation strategies; Codes are faster. 

Puriney:
## Overview

| Functions | What's New |
| --- | --- |
| addImputedScore | additional imputation strategies |
| fit.gene.k | 10X faster; strict control of biological meanings |
| initial.mapping | 1X faster |
| refined.mapping | 17X faster |
| jackStraw | debug #5 |

---

Changes are imported from the extension repository: https://github.com/Puriney/honfleuR. If this PR were merged, `seurat` will have the new imputation strategies and improved running speed from 1X to 17X. 

As these modifications are exactly same as provided in `honfleuR`, the author could do the following convenient steps to try `honflueR` to see the improvements without changing your own codes. 

**1. Install `honflueR`**

``` r
require(devtools)
install_github("Puriney/honfleuR", subdir = "pkg")
```

**2. Try `honflueR`**

Referring to latest vignette here (https://github.com/Puriney/honfleuR/blob/bd171f0d2016f8f1200d4cf5d2fc3ece6ee26989/report/2016-01-31/2016-01-31-compare.pdf) to see how this PR improves seurat with maintaining original syntax, and without changing the entire codes design.

**3. Merge and enjoy**


mschilli87:
I this still under consideration?

Puriney:
Thanks for bring this up. My PR was based on early-back-to-antique-stage of Seurat. I refactored codes a little bit by removing explicit for loop for example but boosted speed a lot, and made some biological sense, for example the `fit.gene.k` function. But it is unlikely be merged with current version in straightforward way, thus I would close this PR. Thanks for holding my PR until now.

PR #35, opened 2016-12-13
Update to FindMarkers

rpeys:
The FindMarkers code has a useful line that allows a user to pass cell names, instead of an identity class, to the "ident.1" and "ident.2" parameters. As the code is currently written, if a user passes cell names, an error will be thrown because of "cells.1=WhichCells(object,ident.1)" before that useful line of code is ever reached. I have changed the order of the two lines and added an if/else to fix the bug.


andrewwbutler:
Thanks for fixing this!

Issue #33, opened 2016-12-13
Issue with FindMarkers


kevin199011:
I was running Seurat with some targeted sequencing data. My targeted sequencing data contains around 6000 cells and 300 genes. However, when I came to the step of finding differential markers, I have an error message:

Error in apply(object@data[genes.use, cells.1], 1, function(x) return(length(x[x >  : 
  dim(X) must have a positive length

I don't have this issue when running with example PBMC data or my own transcriptome data. I'm wondering is it because the total number of genes are too small? If so, how can I fix this issue? Thank you!

Kevin

andrewwbutler:
Hi Kevin,

I would try lowering the cutoffs you're using. You may have weaker markers in your data.

PR #53, opened 2017-03-29
FindMarkers bug

timoast:
I noticed a bug in the `FindMarkers()` function where if any clusters had only 1 cell, this would cause an error (issue #33 might be related):

```r
dat <- readRDS("seurat_obj.Rdata")
WhichCells(dat, 40)  # an example cluster with only 1 cell
```
```
"CAGGTGCGTGCACTTA"
```
```r
FindMarkers(dat, ident.1 = 40)
```
```
Error in apply(object@data[genes.use, cells.1], 1, function(x) return(length(x[x >  :
  dim(X) must have a positive length
```

I believe this is because when the number of cells is 1, iterating over rows in `apply` fails because `dim(cells.1)` is `NULL`. The change I made just coerces the subset of cells that is iterated over by `apply` to a matrix to avoid this problem.

To me, the expected result of setting `min.cells` would be to skip any clusters where the total number of cells is less than `min.cells`, since you can't have more than `min.cells` cells expressing any gene if the total number of cells in the cluster is less than `min.cells`, so I also added an additional check for that. Finding markers for one cell also tends to take a really long time and doesn't really make that much sense, so I think it's good to be able to skip those clusters by setting `min.cells`.

Tim


satijalab:
It's good point. I've made a few changes as well that'll I'll add to this PR if you give me write permissions to this branch. Instead of coercing to a matrix, I use the "drop = F" option for subsetting the matrix when the number of cells in the cluster is 1. This should save a good bit a time/space when that matrix gets big. I also removed the extraneous parameters from FindAllMarkers as mentioned in #54 .

andrewwbutler:
It's a good point. I've made a few changes that I'll incorporate into this PR if you give me write permissions to the branch. I'd rather go with adding "drop = F" to the matrix subsetting rather than coercing as this should save time/space for big datasets. I also removed the extraneous parameters from FindAllMarkers (referenced in #54 ).

PR #64, opened 2017-04-25
Changed pcTopGenes -> PCTopGenes

mukundvarma:
ValidateClusters was still using the deprecated version of the function


andrewwbutler:
Thanks for the fix!

PR #97, opened 2017-07-24
Update seurat.R

mashranga:
I think ggplot 2 has some automated bandwidth smoothing and it need to be controlled. Otherwise the violinplot is not possible to explain !!

### DIFF
--- a/R/seurat.R
+++ b/R/seurat.R
@@ -3971,7 +3971,7 @@ plot.Vln=function(gene,data,cell.ident,ylab.max=12,do.ret=FALSE,do.sort=FALSE,si
   }
   if (y.log) data.melt$value=data.melt$value+1
   p=ggplot(data.melt,aes(factor(ident),value))
-  p2=p + geom_violin(scale="width",adjust=adjust.use,trim=TRUE,aes(fill=factor(ident)))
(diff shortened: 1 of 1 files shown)


mschilli87:
If you already switch from automatic determination of the 'best' bandwidth to a fixed one, why not make it a user contrallable parameter (defaulting to `0.5` like you suggest)?

mashranga:
This correction is in plot.Vln function which is not exported  and plot.Vln is called by VlnPlot function . But in VlnPlot function , there is no parameter to give that value !!

mashranga:
I have asked for some explanation to R-help list !! 
here is the post link: 
https://stat.ethz.ch/pipermail/r-help/2017-July/448297.html

mschilli87:
So again: If you switch from the self-adjusting solution to the fixed one (I'm all with you by the way), why not add this as a parameter (in `VlnPlot` with default `0.5`, passed down to `plot.Vln`, which wouldn't even need any default)?

mashranga:
Thanks for your advice . I have already add this solution (fixed-adjusting bw = 0.5) to my analysis ..but I wondered about that self- adjusting solution.  Will this self-adjusting solution is always appropriate ? Or I am making any confusion on my understanding !!   

You can have a look into this .
https://stat.ethz.ch/pipermail/r-help/2017-July/448297.html

PR #102, opened 2017-07-28
When save.raw = T, delete object@raw.data after modifying it

farcasia:
This fixes a small bug when save.raw is set to true: attempting to modify object@raw.data after it is set to NULL.


mojaveazure:
Thank you for this fix!

PR #129, opened 2017-08-14
Fix error in CalcVarExpRatio

timoast:
Hi,

I ran into a subscript out of bounds error when running `CalcVarExpRatio` (which calls `CalcLDProj`, where the problem occured). The problem was, when there are missing genes when looking up the gene loadings matrix for a given dimension reduction method, the later selection (`[genes.use,]`) fails, and also causes a later problem when trying to multiply matricies of incorrect shapes. I made a couple of small changes to avoid these problems.

I also added the ability to pass arguments through `RunCCA` to `MergeSeurat`, as I found you sometimes have cell name conflicts and need to specify `add.cell.id`.

Tim

### DIFF
--- a/R/dimensional_reduction.R
+++ b/R/dimensional_reduction.R
@@ -430,7 +430,8 @@ RunCCA <- function(
   num.cc = 20,
   genes.use,
   scale.data = TRUE,
-  rescale.groups = FALSE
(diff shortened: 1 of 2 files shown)


andrewwbutler:
Thanks Tim!

PR #132, opened 2017-08-16
Update to WhichCells()

rpeys:
The option to input a "subset.name" argument has not been working until now, instead returning the error "Identity : [ident] not found". Fixed bug and simplified code.


rpeys:
mm, I see now that "subset.name" needs to be used in conjuction with "accept.value." Perhaps that can be made clearer in the documentation. I had been thinking I was supposed to use "subset.name" = "metadata.column.name" together with the "ident" = "desired.metadata.value" argument to pull cell names according to metadata.

PR #140, opened 2017-08-25
unquote variable for use in dplyr::gather

timoast:
Ran into this error when running FeatureHeatmap:

```r
dat <- readRDS("aligned.Rdata")
FeatureHeatmap(dat, features.plot = "XCP1")
Error in FUN(X[[i]], ...) : object 'features.plot' not found
```

Simple 2 character fix, unquoting `features.plot` in the `FeatureHeatmap` function.

Although, this function was working for me previously and I haven't been able to work out what changed in Seurat to produce the error, so someone may want to verify they get the same error when running FeatureHeatmap before merging.


timoast:
Looks like this has been fixed in 2.0.1

Issue #137, opened 2017-08-23
Error running BuildClusterTree


rmtsoa:
Hi:

Thanks for putting this tutorial together, it is a great tool. 
I have a question about BuildClusterTree, and the following is a quick summary of what I did:

I ran the RunPCA(Lot351101, pc.genes = Lot351101@var.genes, do.print = T, pcs.print = 1:5, genes.print = 30), and got the results stored in Lot351101@dr

However, as I tried to run BuildClusterTree:
Lot351101 <- BuildClusterTree(Lot351101,do.reorder = TRUE,reorder.numeric = TRUE,pcs.use = 1:20)

This is the output and error I got:
[1] "Finished averaging RNA for cluster 351101Cluster1"
[1] "Finished averaging RNA for cluster 351101Cluster2"
[1] "Finished averaging RNA for cluster 351101Cluster3"
[1] "Finished averaging RNA for cluster 351101Cluster4"
[1] "Finished averaging RNA for cluster 351101Cluster5"
[1] "Finished averaging RNA for cluster 351101Cluster6"
[1] "Finished averaging RNA for cluster 351101Cluster7"
[1] "Finished averaging RNA for cluster 351101Cluster8"
[1] "Finished averaging RNA for cluster 351101Cluster9"
[1] "Finished averaging RNA for cluster 351101Cluster10"
[1] "Finished averaging RNA for cluster 351101Cluster11"
[1] "Finished averaging RNA for cluster 351101Cluster12"
 Show Traceback
Error in is.data.frame(x) : no slot of name "pca.rot" for this object of class "seurat"

I could not find 'pca.rot' slot anywhere, any suggestion?

Thanks,
Rosemarie

> sessionInfo()
R version 3.4.1 (2017-06-30)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS Sierra 10.12.6

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.4/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[...]

PR #178, opened 2017-10-15
Fix error in AveragePCA() and BuildClusterTree() when using PCA space. 

yueqiw:
Hi ,

I run into an error when running `BuildClusterTree(object, pcs.use=1:20, ...)` in the PCA space, as previously reported in Issue #137 . 

I fixed the code to make it work. Now it uses the averaged PCA cell embeddings from each cluster to build the tree. Whether the euclidean distance in the PCA space is weighted by eigenvalues depends on the `weight.by.var` argument in `RunPCA()`. 

In addition, there is another error when running `AveragePCA()`:
` number of rows of result is not a multiple of vector length (arg 2)`
This is also fixed in the PR.


andrewwbutler:
Looks great! Thanks for the fix. Could you make the PR into the develop branch though rather than into master?

PR #180, opened 2017-10-17
fix default DE test from 'bimod' to 'wilcox'

julietchen:
change default test.use to reflect changed documentation in
FindAllMarkers, FindMarkersNode, FindAllMarkersNode


andrewwbutler:
Thanks for the fix!

Issue #188, opened 2017-10-23
AverageExpression crashes when only one gene is provided


b-dawes:
Hello,

Running AverageExpression with only one gene in genes.use gives the error message:
```
Error in apply(X = data.use[genes.assay, temp.cells], MARGIN = 1, FUN = fxn.average) : 
  dim(X) must have a positive length
```

I get the same error whether I provide just a string or wrap the string in c() or list().

For reference, I am running the development branch of 2.1.

Best,
Brian

PR #189, opened 2017-10-25
fix for issue #188

rpeys:
Allows AverageExpression() to work even when only one gene is provided. By adding "drop = F" the matrix structure of data.temp is maintained even though it only has one row (default would be for its structure to change from a matrix to a vector), allowing data.temp to be successfully passed as an argument to the apply function.

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -587,7 +587,7 @@ AverageExpression <- function(
       }
       if (length(x = temp.cells) >1 ) {
         data.temp <- apply(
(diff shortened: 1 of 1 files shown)


rpeys:
Thanks for the blog post! Updated.

mojaveazure:
@rpeys I've switched the base branch from master to develop. Next time, please ensure that your pull request is merging into develop, not master.

andrewwbutler:
Thanks for the fix!

PR #201, opened 2017-10-31
Fix #197 if do.reorder is true

### LINKED ISSUE #197: BuildClusterTree silently
Hello,

This is just a small quibble, but it looks like there's no way to run BuildClusterTree silently. For example, running BuildClusterTree prints:
```
[1] "Finished averaging RNA for cluster 1"
[1] "Finished averaging RNA for cluster 2"
[1] "Finished averaging RNA for cluster 3"
[1] "Finished averaging RNA for cluster 4"
[1] "Finished averaging RNA for cluster 5"
[1] "Finished averaging RNA for cluster 6"
[1] "Finished averaging RNA for cluster 7"
[1] "Finished averaging RNA for cluster 8"
[1] "Finished averaging RNA for cluster 9"
```

I'm running seurat in a pipelined script and want to keep my output log clean. I found a work-around by wrapping the function call in `invisible(capture.output())` but having an option built into the function would be much nicer looking.

It looks like the output is being printed by the AverageExpression call which has an option for show.progress. This could probably be fixed by having a show.progress option in BuildClusterTree that gets passed to AverageExpression.

Brian

b-dawes:


mojaveazure:
Thanks for the fix!

Issue #45, opened 2017-02-17
progress output clutters knitr output


mschilli87:
When running Seurat through `knitr` in an `Rmd` file, Seurat's progress output clutters the resulting output document.

This is caused by Seurat printing its diagnostic output to `STDOUT` instead of `STDERR` as common.

Also see [this answer of the `knitr` dev on the subject](http://stackoverflow.com/a/20024043/2451238).

mschilli87:
There are several new instances of progress bars and print statements cluttering `knit` output.
It seems I cannot re-open this issue, but instead of creating a new one, I'll just open another pull request with a fix similar to the previously accepted a019766e8cfb9088f71dcd5aa9b7ab5eb47d094b.

PR #46, opened 2017-02-17
moved progress output from STDOUT to STDERR (knitr compatibility)

mschilli87:
fixes #45

PR #226, opened 2017-11-27
moved progress output from STDOUT to STDERR (again)

mschilli87:
This is needed for knitr compatibility and fixes issue #45.

This was done before (see pull request #46, commit a019766e8cfb9088f71dcd5aa9b7ab5eb47d094b), but since then, new progress output to STDOUT was added.

I'm testing my branch right now but as all I do is replacing `print` by `cat` and adding `file = stderr()` to those `cat` or `txtProgressBar` calls, I doubt this affects anything else.


mschilli87:
Local test after `devtools::install_github("mschilli87/seurat@knitr_fix")` finished successfully.
Everything like before, but no progress in the `knitr` output. :smile:

mojaveazure:
Is there a reason why setting `verbose = FALSE`, `display.progress = FALSE`, or `print.output = FALSE` to function calls doesn't work? Admittedly, we could do a better job in standardizing the argument name, but the current build of Seurat should be just fine with making documents using knitr.

mschilli87:
1. To name one particular example: [`AlignSubspace` doesn't have any of the arguments you mentioned](https://github.com/satijalab/seurat/blob/401c7944f32ae213a99e7a93dcbb9d5846f2dd3c/R/dimensional_reduction.R#L896-L901) (maybe I missed another synonym), but [calls `ScaleData` without passing `...`](https://github.com/satijalab/seurat/blob/401c7944f32ae213a99e7a93dcbb9d5846f2dd3c/R/dimensional_reduction.R#L930) and thus without the chance to set [its `display.progress` argument](https://github.com/satijalab/seurat/blob/401c7944f32ae213a99e7a93dcbb9d5846f2dd3c/R/preprocessing.R#L468)  to `FALSE`.
 2. Also, as you already mentioned, those functions having an argument to control this behaviour, use various names which makes it very hard to use.
 3. Some function already print progress to STDERR, others to STDOUT, which adds another layer of inconsistency & thus unpredictablity.
 4. [Printing `"ERROR"` to STDOUT](https://github.com/satijalab/seurat/blob/cec7cb95c73fd6d605723e9af9a1f96eda5635de/R/spatial.R#L31-L34) just feels wrong. :wink:
 5. I don't see how sending progress output to STDERR could break anything, but it clearly helps with `knitr` (even if it 'just' saves specifying all the arguments you mentioned).
 6. This way, I can see the progress in the R console (which is helpful/informative and would be lost by using your sulution) but it won't end up in my output file (where it simply doesn't belong, IMHO).
 7. I figured providing a fix analogous to a one previously accepted in the same repo instead of just complaining that 'it does not work' would be the better solution. :wink:

But obviously, any other solution making it convinient to run Seurat in `knitr` without seing its progress output in the output document would be fine with me (I just probably won't provide a pull request myself). :wink:

mojaveazure:
Thanks Marcel! Sorry this took so long to get merged in.

PR #229, opened 2017-11-27
change x-axis label rot to 45 degree

rpeys:
For the x-axis labels in VlnPlot: ability to rotate labels is important for making them fit, but my personal opinion is that a 45 degree rotation is more legible than the 90 degree rotation (which requires viewers to tilt their heads to read)

### DIFF
--- a/R/plotting_internal.R
+++ b/R/plotting_internal.R
@@ -683,7 +683,7 @@ SingleVlnPlot <- function(
     plot <- plot + scale_fill_manual(values = cols.use)
   }
   if (x.lab.rot) {
-    plot <- plot + theme(axis.text.x = element_text(angle = 90, vjust = 0.5))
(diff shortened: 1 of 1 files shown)


mschilli87:
And I prefer to rotate the whole plot by swapping the x & y axis (adding `coord_flip()`). :wink:

One great feature off `ggplot` is that there is no need to please everyone with the default:
If you call `VlnPlot` with `do.return=TRUE`, you will get the `ggplot` object to modify it yourself:

```r
p <- VlnPlot(seurat.object, features, do.return=TRUE)
print(p)
print(p + theme(axis.text.x=element_text(angle=45, hjust=1))
print(p + coord_flip())
```

rpeys:
Thanks, mschilli87. That's a great point about using do.return=TRUE. 
Even so, I wanted to suggest this updated default so that the default image produced is the one least likely to need further tweaking.

PR #278, opened 2018-01-22
Fix spatial@final.prob variable name

gokceneraslan:
Replace `finalprob` with `final.prob` in spatial.R


mojaveazure:
Thanks for catching this!

PR #306, opened 2018-02-11
Plot colour gradient legend

leonfodoulian:
Hi,

I added a function that allows to plot a colour gradient legend from a palette of colours in hexadecimal form. This function is called `GetGradientLegend()` and will be used in the `SplitDotPlotGG()` function (pull request for the update of that function will shortly be posted). The proposed change is an alternative to produce a similar graph as that found in [Figure 2D of the paper describing the alignement performed by Seurat's `RunCCA()` and `RunMultiCCA()` functions](https://www.biorxiv.org/content/biorxiv/early/2017/07/18/164889.full.pdf). 

Best,
Leon


leonfodoulian:
I have slightly updated the function, as I realised that `mapply()` didn't pass to `GetGradientLegend()` one of the input.

PR #307, opened 2018-02-11
Plot gradient legends in SplitDotPlotGG()

leonfodoulian:
Hi again,

I updated the `SplitDotPlotGG()` function, such that now one can plot gradient legends for each of the 2 conditions, hence bypassing the need to pass by Adobe Illustrator to generate such gradient legends. This pull request is linked to [#306](https://github.com/satijalab/seurat/pull/306). 

Along the way, I also found that the `cols.use` argument was not passed to the function. The function previously produced gradient colours from grey to blue for the first condition, and from grey to red for the second condition. This is now fixed in the new version of the function. 

Best,
Leon


leonfodoulian:
I have slightly updated the function, as I realised that `mapply()` didn't pass to `GetGradientLegend()` one of the input.

PR #392, opened 2018-03-29
Output numeric prop.test p-values from JackStrawPlot()

leonfodoulian:
Hi,

This fix outputs in the `data` slot from the `ggplot2` object the numeric p-values from the `prop.test` performed by `Seurat::JackStrawPlot()`. This facilitates the indexing of PCs for downstream analysis using automated code lines.

Best,
Leon

### DIFF
--- a/R/jackstraw_internal.R
+++ b/R/jackstraw_internal.R
@@ -4,7 +4,8 @@ jackstraw.data <- setClass(
   slots = list(
     emperical.p.value = "matrix",
     fake.pc.scores = "matrix",
-    emperical.p.value.full = "matrix"
+    emperical.p.value.full = "matrix",
+    overall.p.values = "matrix"
   )
 )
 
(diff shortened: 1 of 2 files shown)


andrewwbutler:
Hi Leon,

It's a good idea to return these values but I think it might make more sense to associate them with the object. I created a slot in the jackstraw class so you can now pull them from `object@dr$pca@jackstraw@overall.p.values` for use in downstream analysis.

leonfodoulian:
Hi Andrew,

I thought of that option too, but since these values are computed through a 'plotting' function, I thought that it could have made sense to return them with the plotting data. But yes, I agree that returning these values in a slot in the Seurat object makes much more sense.

Thanks!

Best,
Leon

andrewwbutler:
Yeah, this part of the code should be restructured at some point to separate the computation from the plotting (similar to what we did for the old `MeanVarPlot`). Slotting it in for now should allow for that refactoring in the future without you having to go back and change your automation scripts :)

PR #444, opened 2018-04-27
Warn when increasing num.cores

mschilli87:
When setting `do.par = TRUE` but `num.cores = 1`, `JackStraw` interprets this to use half of the available cores.
While this seems to be intended it is not documented and might cause confusion, so I figured a warning would be fair.
The same is already done when specifying more cores than available, so this also increases consistency.


mojaveazure:
Thanks Marcel!

PR #466, opened 2018-05-09
Avoid Compilation Errors by Fixing Makevars

evolvedmicrobe:
MacOS X versions of R have compilation issues with fortran code.  In particular, the Makeconf file distributed with R points to a fortran library that either doesn't exist or may be out of date.  The problem is described here:

https://thecoatlessprofessor.com/programming/rcpp-rcpparmadillo-and-os-x-mavericks--lgfortran-and--lquadmath-error/

Installing Seurat from source can therefore lead to a compile error depending on the toolings and versions of R/gfortran installed.

However, Seurat does not have any fortran dependencies and so does not need to link to fortran.  By removing this line, we can avoid compilation errors.

Also, no folder named inst/include exists, so there is no need to reference it.


mojaveazure:
Thanks Nigel! One thing we ask is that you set the base branch to `develop` for future pull requests, as we don't accept pull requests to `master`.

evolvedmicrobe:
Thanks for merging this so quickly Paul! And sorry for not seeing dev early, in the future will be sure to base off that.

Thanks again!
Nigel

andrewwbutler:
In preparing the next CRAN release, we are seeing the following warning when running R CMD check:

`* checking for portable use of $(BLAS_LIBS) and $(LAPACK_LIBS) ... WARNING
  apparently PKG_LIBS is missing $(FLIBS) in 'src/Makevars'`

It looks like the link to fortran is going to be required by CRAN. See [Writing R Extensions](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Using-Makevars). 

> For portability, the macros BLAS_LIBS and FLIBS should always be included after LAPACK_LIBS (and in that order). 

For that reason, I've put it back in ab52a6a00b26396454aba9569cbc4d586a3e2ea7 .

PR #468, opened 2018-05-09
Update JackStrawPlot to return Seurat object instead of plot

leonfodoulian:
Hi,

Related to PR #392. However, the fix in #392 did not return a `Seurat` object. In this case, `Seurat::JackStrawPlot()` returns a `Seurat` object, with `object@dr$pca@jackstraw@overall.p.values` representing PC p-values and `object@dr$pca@misc$jackstraw.plot` storing the `ggplot2` plot.

Please do make sure also to update the vignette on your webpage!

Best,
Leon


andrewwbutler:
Good point. Thanks Leon!

Issue #485, opened 2018-05-21
Conversion to SingleCellExperiment from Seurat objects


wolfganghuber:
Hi, I posted a problem here: https://support.bioconductor.org/p/109152/
I hope it's OK for you that I do it like this.
Thanks for the great effort, and kind regards 
Wolfgang Huber

PR #487, opened 2018-05-22
These changes are intended to fix issue 485

wolfganghuber:
https://github.com/satijalab/seurat/issues/485
In addition, added `stringsAsFactors = FALSE` in several places to try to avoid converting gene names into factor
Also, `seq_len(n)` is better than `1:n` (esp for cases where n<1)


andrewwbutler:
Thanks for the fix!

wolfganghuber:
Thanks Andrew for the quick reaction!

PR #498, opened 2018-05-24
Add PHATE dimensionality reduction

scottgigante:
Hi! We spoke a short while ago about integrating PHATE into Seurat.

This won't pass Travis-CI yet because the phateR package is pending manual inspection for upload to CRAN but I thought I would submit this PR to check that the API suits what you're expecting. I'll post a message on here when phateR is on CRAN proper.

Scott


mojaveazure:
To get this to pass Travis checks, please update the CI configuration YAMLs to use phateR from GitHub. For Travis, see [this example](https://github.com/satijalab/seurat/blob/master/.travis.yml#L19) in our Travis YAML, or [this one](https://github.com/mojaveazure/loomR/blob/develop/.travis.yml) in the YAML for loomR. For AppVeyor, see [this example](https://github.com/gaborcsardi/ask/blob/master/appveyor.yml#L16) from ask.

scottgigante:
Hi Paul, phateR is currently under review at CRAN - if you're happy with the API as I have written it I'm happy to wait until PHATE arrives on CRAN to push this through.

mojaveazure:
Hi Scott, I understand that, but the CI's won't be retriggered once phateR goes to CRAN. We use the CIs as a way to check the API in addition to our own checks, so configuring the CIs to use the GitHub version of the package (which I'm assuming is the same as the one you submitted to CRAN) will help us move this along quicker. Plus, it also checks that everything works properly under systems that we don't have ourselves (eg. Windows). That is why we ask that you modify the YAMLs to pull phateR from GitHub for the time being.

Otherwise, you can wait for phateR to be available on the CRAN mirrors and open a new PR with the changes in this one.

mojaveazure:
For Travis, set up an install script (similar to [this one](https://github.com/mojaveazure/loomR/blob/develop/travis_setup.sh)) and use the `before_install` section to run it. For AppVeyor, [here](https://github.com/ogrisel/python-appveyor-demo/blob/master/appveyor.yml) is an example in setting up a Python dependency. I think all you'll need is to install Python, Pip, and PHATE, not every check that the example provides.

andrewwbutler:
Hi Scott, 

It might be easier to just handle the PHATE dependency similar to the way we did for other dependencies not on CRAN. For example:

https://github.com/satijalab/seurat/blob/96d07d80bc4b6513b93e9c10d8a9d57ae7016f9f/R/differential_expression.R#L1198-L1201

Or see `RunUMAP` for an example if you want to use the python version of PHATE. 

https://github.com/satijalab/seurat/blob/96d07d80bc4b6513b93e9c10d8a9d57ae7016f9f/R/dimensional_reduction.R#L1082-L1084

scottgigante:
Hi Andrew,

Thanks for the suggestion. All of this dance I'm doing is to try to make the `\examples` in `RunPHATE` run on CI. Technically phateR runs with reticulate under the hood, so the examples are run with `if (reticulate::py_module_available("phate")` wrappers, so if python isn't installed there won't be any errors. It's an option to just let this happen, but of course I would prefer the checks to pass than be skipped.

phateR will be on CRAN shortly so there is no need to have it installed from github in the long run.

scottgigante:
Looking at it now it appears testing many versions of Python is a) too much burden on your CI and b) unnecessary, so I'll pull back the Appveyor test to just python3. All it needs to test really is that the interface with phate is working - the rest we can do on our end.

scottgigante:
Looks like appveyor had a 503 service unavailable for some reason totally unrelated to the last commit. I'm going to close and re-open the PR.

scottgigante:
Okay, phateR is now on CRAN and installs successfully on both my Windows and Linux machines, and the checks were passing here prior to removing the github dependencies. I'm pretty sure the problems with the builds now are to do with HDF5?

andrewwbutler:
Yeah, the problems were due to HDF5. We added a helper function in develop to read h5 files and those dependencies broke travis. It's fixed now but also introduced a merge conflict here (sorry). I think the cleanest way to resolve this one is to move the python setup commands you have under `before_install:` into the `travis_setup.sh` script I added to get a more recent hdf5 version.

andrewwbutler:
Thanks Scott!

scottgigante:
Not a problem! Thank you!

scottgigante:
@andrewwbutler copying the travis file for use in another project I noticed that I didn't actually include the critical line, `pip install phate`, in the `travis_setup.sh` when I did the merge. I can open another PR or you can add it, either way.

andrewwbutler:
No worries. Just added it.

PR #500, opened 2018-05-24
Fast residual

evolvedmicrobe:
While profiling some analysis, I noticed that a lot of time was being spent in this function, and put some optimizations into it.  For a test case where this results in the ScaleData function only taking half as much time when called to regress on 2 variables using 2 cores for a dataset with 7.3K cells and 20K genes.  The speed improvements come from two main changes:

1 – **Avoids allocating several large empty character vectors in the inner loop.** Instead, it creates one empty_char vector and repeatedly reuses it, avoiding a lot of memory allocation overhead.

2 -- **Reuses the QR Decomposition** - Rather than recompute the QR decomposition for each regression, it reuses one calculated once.

After these first two commits, some other minor changes are included which were flagged in profiling.

I confirmed the ScaleData command returned the same values with these changes in place as it did originally.

### DIFF
--- a/R/preprocessing_internal.R
+++ b/R/preprocessing_internal.R
@@ -108,6 +108,7 @@ RegressOutResid <- function(
   data.resid <- foreach(i = 1:max.bin, .combine = "c", .options.snow = opts) %dopar% {
     genes.bin.regress <- rownames(x = data.use)[bin.ind == i]
     gene.expr <- as.matrix(x = data.use[genes.bin.regress, , drop = FALSE])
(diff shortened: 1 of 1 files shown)


andrewwbutler:
This is great! Really like the idea of reusing the QR.

evolvedmicrobe:
@andrewwbutler Thanks for accepting these requests! I've been playing around with the Seurat package and am working to help put some tutorials together for people, when I see some quick optimizations I hope you don't mind if I keep sending them your way for inclusion.  Am very excited about this package.

Warm wishes,
Nigel

PR #502, opened 2018-05-25
Performance Optimization of Sparse Matrix Access and RNG Generation

evolvedmicrobe:
Some changes to speed up the `LogNormalize` and `SampleUMI` functions.  These changes for SampleUMI changed the execution time from 24 seconds down to 0.4 seconds. The two changes are:

**Use scalar runif** Calling runif to return one value is expensive, as the underlying code must generate an R NumericVector, populate the vector, then access the scalar value in this vector, which involves a lot of R allocator overhead.

Rcpp provides a much lighter interface where you can just generate a single double value, rather than a whole vector (see: http://gallery.rcpp.org/articles/random-number-generation/).

Since the code now runs an example 7K cells and 20K gene dataset in less than half a second, it also seemed sensible not to display a progress bar by default.

**coeffRef(it.row(), it.col()) -> it.valueRef()** 

Performance optimization.  In a few of the edited places, the code is iterating over the elements of a sparse matrix using an iterator.  However, when it came time to update the sparsematrix at the given position, rather then directly assign the value, the code called `coeffRef` to assign it.

This is inefficient because `coeffRef` involves an expensive binary search to find position `i,j` in the sparse matrix, even though the memory location of `i,j` is already known to the iterator on the stack.  To directly use the known value, we can just call `.valueRef()` on the iterator instead of looking for the position again.

A discussion of this difference is given in the `Iterating over the nonzero coefficients` section of this tutorial: https://eigen.tuxfamily.org/dox/group__TutorialSparse.html

Benchmarking showed this shaved 13% off the time for NormalizeData, and dropped the time for `SampleUMI` from 0.8 seconds to 0.4 seconds after the change listed above was implemented.

### DIFF
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -651,7 +651,7 @@ SampleUMI <- function(
   data,
   max.umi = 1000,
   upsample = FALSE,
-  progress.bar = TRUE
+  progress.bar = FALSE
 ) {
   data <- as(data, "dgCMatrix")
   if (length(x = max.umi) == 1) {
--- a/man/SampleUMI.Rd
+++ b/man/SampleUMI.Rd
@@ -4,7 +4,7 @@
 \alias{SampleUMI}
 \title{Sample UMI}
 \usage{
-SampleUMI(data, max.umi = 1000, upsample = FALSE, progress.bar = TRUE)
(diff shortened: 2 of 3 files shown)


andrewwbutler:
Nice! Thanks again!

PR #503, opened 2018-05-26
VlnPlot without jittered dots

QiangTU:
if `point.size.use == 0`, VlnPlot will only plot violin plots without jittered dots.


andrewwbutler:
Thanks! We had been just setting `point.size.use = -1` which also removes the points but this is definitely more intuitive.

Issue #505, opened 2018-05-26
Duplicate cell names - RunMultiCCA


ptrcalmeida:
Hi, I was able to add the unique prefixes to each of the 4 indivual 10X datasets that I'm working with prior to CreateSeuratObject using the following, as suggested in #270 :
colnames(x = s1.data) <- paste('S1', colnames(x = s1.data), sep = '_')

It works great: A1_AAACCTGCACGACGAA

However, I still get an error when running RunMultiCCA:
lung.integrated <-RunMultiCCA(ob.list, genes.use=genes.use, num.ccs = 1)

Error in MergeSeurat(object1 = combined.object, object2 = object.list[[i]], :
Duplicate cell names, please provide 'add.cell.id1' and/or 'add.cell.id2' for unique names

Help, please? More info attached. Thank you!

[RunMultiCCA error.txt](https://github.com/satijalab/seurat/files/2042012/RunMultiCCA.error.txt)

leonfodoulian:
Hi,

This issue is a duplicate of issue #400. I have already provided two code lines to check whether you have duplicated cell names in #400.

Best,
Leon

ptrcalmeida:
Thank you, Leon!

Issue #400, opened 2018-04-05
Creating cell IDs for different datasets for RunMultiCCA


nicodemus88:
Hi to the seurat team,

I am trying to run RunMultiCCA() using 4 datasets. I checked back on the multiple dataset tutorial and past issues ( issue #270 and #290 ) and I know I have to create unique names for cells from different datasets when creating the initial Seurat object.

However, what I couldn't find is how do I do that? It's not mentioned in the tutorial. I browsed the documentations for CreateSeuratObject but I'm not sure how to do that. Do I use the 'names.field' and 'names.delim' parameter?

I tried specifying the 'names.field' parameter to different name for different datasets but I got several warning messages. 

```R
s1 <- CreateSeuratObject(raw.data = s1.data, min.cells = 3, min.genes = 200, project = "S1", names.field = "S1")
There were 50 or more warnings (use warnings() to see the first 50)
> warnings()
Warning messages:
1: In FUN(X[[i]], ...) : NAs introduced by coercion
2: In FUN(X[[i]], ...) : NAs introduced by coercion
3: In FUN(X[[i]], ...) : NAs introduced by coercion
4: In FUN(X[[i]], ...) : NAs introduced by coercion
5: In FUN(X[[i]], ...) : NAs introduced by coercion
```

Can you guide me on how do I insert unique cell identifiers for each dataset during the initial Seurat object creation?

Hope to hear back from you soon. Thank you.

nicodemus88:
@mojaveazure Thank you for your reply. I managed to run RunMultiCCA now.
I will close this topic now.

ptrcalmeida:
Hi, I was able to add the unique prefixes to each of the 4 indivual 10X datasets that I'm working with prior to CreateSeuratObject using the following, as suggested in #270 :
colnames(x = s1.data) <- paste('S1', colnames(x = s1.data), sep = '_')

It works great: A1_AAACCTGCACGACGAA

However, I still get an error when running RunMultiCCA: 
lung.integrated <-RunMultiCCA(ob.list, genes.use=genes.use, num.ccs = 1)

**Error in MergeSeurat(object1 = combined.object, object2 = object.list[[i]],  : 
  Duplicate cell names, please provide 'add.cell.id1' and/or 'add.cell.id2' for unique names**

Help, please? More info attached. Thank you!

[RunMultiCCA error.txt](https://github.com/satijalab/seurat/files/2042011/RunMultiCCA.error.txt)

[...]

Issue #290, opened 2018-01-28
Error in RunMultiCCA due to duplicate cell names


del2007:
Hi, 

When attempting to integrate 4 datasets, I get the error below due to common cell names between the sets.  
Where should I input the add.cell.id1 - add.cell.id4? On the RunMultiCCA line? The merge obj function is built into it I see...

I thought about recreating the seurat objects if there is no other solution? How can I rename the cell ids while I do this? 

Thank you!

> norms.integrated <- RunMultiCCA(ob.list, genes.use = genes.use, num.ccs = 15)
[1] "Computing CC 1"
[1] "Computing CC 2"
[1] "Computing CC 3"
[1] "Computing CC 4"
[1] "Computing CC 5"
[1] "Computing CC 6"
[1] "Computing CC 7"
[1] "Computing CC 8"
[1] "Computing CC 9"
[1] "Computing CC 10"
[1] "Computing CC 11"
[1] "Computing CC 12"
[1] "Computing CC 13"
[1] "Computing CC 14"
[1] "Computing CC 15"
Error in MergeSeurat(object1 = combined.object, object2 = object.list[[i]],  : 
  Duplicate cell names, please provide 'add.cell.id1' and/or 'add.cell.id2' for unique names

leonfodoulian:
Hi,

I don't think there is such an option to `RunMultiCCA()`. This question has already been asked, in [issue #270](https://github.com/satijalab/seurat/issues/270). I guess you should give your cells unique names when creating your initial Seurat objects.

Hope this helps!

Best,
Leon

yingxixu:
Did you solve this problem? all my 3 databases are 10X, so cell.id are same, I still don't know how to add"add.cell.id“...

Issue #270, opened 2018-01-18
RunMultiCCA() question


cimbusch:
Hello,

great the RunMultiCCA() is now available!

When running RunCCA() is was setting add.cell.id1 and add.cell.id2 to avoid duplicate cell names. With RunMultiCCA() this option does not make so much sense anymore, but there should be still an option to create unique cell identifiers? Where can I inject a prefix to the cell ids, maybe directly when creating the the Seurat object: CreateSeuratObject() ?

Cheers,
  Charles

andrewwbutler:
We'll provide an example workflow for multiple alignment on the website in the very near future. One way to handle this though would be to prepend your unique identifiers to the cell names of the initial input matrix you pass to `CreateSeuratObject`.

cimbusch:
Ok, I see. Yeah, just appending it to the initial input matrix is working. Thanks.

cimbusch:
Looking forward to an example workflow, the downstream part seems to be a bit different:

> data.union <- RunMultiCCA(samples_list, genes.use = hvg.union)
[1] "Computing CC 1"
Performing log-normalization
0%   10   20   30   40   50   60   70   80   90   100%
[----|----|----|----|----|----|----|----|----|----|
**************************************************|
[1] "Scaling data matrix"
  |=====================================================================================================================================================| 100%
> 
> # visualize results of CCA
> # plot CC1 versus CC2 and look at a violin plot
> p1 <- DimPlot(data.union, reduction.use = "cca", group.by = "sample", pt.size = 0.5, do.return = T)
Error in `[.data.frame`(data.plot, cells.use, dim.codes) : 
  undefined columns selected
> DimHeatmap(data.union, reduction.type = "cca", cells.use = 500, dim.use = 1:9, do.balanced = T)
Error in DimTopGenes(object = object, dim.use = ndim, reduction.type = reduction.type,  : 
  Gene loadings for cca with use.full=FALSE have not been computed

ghoshal:
Hi, 

I am having issues with RunMultiCCA as well. 

hvg.D04 <- rownames(x = head(x = D04@hvg.info, n = 5000))
hvg.D05 <- rownames(x = head(x = D05@hvg.info, n = 5000))
hvg.D06 <- rownames(x = head(x = D06@hvg.info, n = 5000))
hvg.union <- Reduce(union, list(hvg.D04, hvg.D05, hvg.D06))

# lastly, we set the 'protocol' in each dataset for easy identification
[...]

PR #510, opened 2018-05-30
Improve setting of cell names

lazappi:
With more people combining datasets for analysis with `RunMultiCCA` fixing duplicate cell names is becoming more of a problem as seen in #270, #290, #400, #505. To address this I have made the following changes:

* Add a new `RenameCells` function that renames cells either by adding a prefix or replacing with a new vector of names
* Modified `MergeSeurat` to use the new `RenameCells` function (to avoid duplication)
* Add a new `add.cell.ids` parameter to `RunMultiCCA` which applies `RenameCells` to each object in `object.list`
* Add new documentation and tests


lazappi:
Merged develop to hopefully get Travis working after #511.

andrewwbutler:
Thanks Luke! Very helpful function and should reduce confusion when analyzing multiple datasets.

nbahlis:
Hello, 
when using RenameCells I get the error "invalid dimnames given for "dgCMatrix" object "
Does RenameCells work with sparse matrices?

thanks

lazappi:
@nbahlis I think it might be better to discuss this as an issue rather than on this pull request so I have opened #523

PR #513, opened 2018-06-01
Fixed conversion error from seurat with no dr to anndata [WIP]

ivirshup:
Hi! 

I've found a bug in the conversion methods, conversion to `anndata` from `seurat` fails if there are no dimensionality reductions:

```R
> suppressPackageStartupMessages(library(Seurat))
> nodr <- pbmc_small
> nodr@dr <- list()
> Convert(nodr, "anndata")
/usr/local/opt/python/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
Error in py_call_impl(callable, dots$args, dots$keywords) : 
  IndexError: list index out of range

Detailed traceback: 
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/anndata/base.py", line 636, in __init__
    filename=filename, filemode=filemode)
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/anndata/base.py", line 825, in _init_as_actual
    obsm = utils.convert_dictionary_to_structured_array(obsm)
  File "/usr/local/opt/python/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/anndata/utils.py", line 74, in convert_dictionary_to_structured_array
    arr = np.zeros((len(cols[0]),), dtype)

> Convert(pbmc_small, "anndata")
AnnData object with n_obs × n_vars = 80 × 230 
    obs: 'n_genes', 'n_counts', 'orig_ident', 'res_0_8', 'res_1'
    var: 'gene.mean', 'gene.dispersion', 'gene.dispersion.scaled'
    obsm: 'X_pca', 'X_tsne'
```

It looks like the `anndata` constructor for doesn't like getting an empty dictionary for it's `obsm` argument, but is fine with receiving it's null default, which this pull request implements. I also added a couple "this shouldn't throw an error" tests.

While fixing this, I've run into a few more issues with the converters and would like to put more fixes and some general tests into this pull request. To do that, I'd need some input on what the expectations of the conversion methods are and maybe some explanation for a few lines of code. Let me know if that's something you'd be interested in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,9 @@ before_install:
   - chmod +x travis_setup.sh
   - ./travis_setup.sh
 
+before_script:
+  - R -e "reticulate::py_config()"
+
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ before_install:
   - ./travis_setup.sh
 
 before_script:
+  - Rscript py_install.R
   - R -e "reticulate::py_config()"
 
 addons:
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -237,7 +237,7 @@ Convert.seurat <- function(
           reduction.type = dr
         ))
       }
-      obsm <- dict(obsm)
+      obsm <- if (!identical(obsm, list())) dict(obsm) else NULL
       meta_data <- from@meta.data
       if ("nUMI" %in% colnames(x = meta_data)) {
         colnames(x = meta_data) <- gsub(
--- a/R/dimensional_reduction.R
+++ b/R/dimensional_reduction.R
@@ -1119,6 +1119,11 @@ RunDiffusion <- function(
 #' (URL: http://doi.org/10.1101/120378),
 #' <URL: https://www.biorxiv.org/content/early/2017/12/01/120378>.
 #' @examples
+#' system2(conda_binary("auto"), args=c("list"))
+#' reticulate::conda_list()
+#; reticulate::use_condaenv("base")
+#' system2(conda_binary("auto"), args=c("list"))
+#' reticulate::conda_list()
 #' if (reticulate::py_module_available("phate")) {
 #'
 #' # Load data
--- a/R/dimensional_reduction.R
+++ b/R/dimensional_reduction.R
@@ -1119,9 +1119,10 @@ RunDiffusion <- function(
 #' (URL: http://doi.org/10.1101/120378),
 #' <URL: https://www.biorxiv.org/content/early/2017/12/01/120378>.
 #' @examples
+#' library(reticulate)
 #' system2(conda_binary("auto"), args=c("list"))
 #' reticulate::conda_list()
-#; reticulate::use_condaenv("base")
+#' reticulate::use_condaenv("base")
 #' system2(conda_binary("auto"), args=c("list"))
 #' reticulate::conda_list()
 #' if (reticulate::py_module_available("phate")) {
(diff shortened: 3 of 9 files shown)


ivirshup:
Forgot to add python dependencies for travis and appveyor, I'll be giving that a shot now.

andrewwbutler:
Hi Isaac,

Good catch! Please do add any other fixes for bugs that you find. We'd be happy to try to clarify any expectations/code chunks if it's helpful.

Our general workflow is to merge changes into the develop branch first, merging into master only when doing new releases. We actually have some additional code there for installing python dependencies on travis/appveyor which has been added since our last release (hence not on master). For travis, you could probably just add on the pip installs to the end of the `travis_setup.sh` script.

ivirshup:
Alright, I've rebased and made progress on the builds (macOS and x64 windows builds are working). It looks like there are some issues with `reticulate` seeing `anndata`. Are there any currently running tests using python?

A few questions about the conversion tests:

* What should I be testing against? There's the `pbmc_small` object, but it's not necessarily from the current version of seurat. Should I be building my own, like in `test_seurat_object.R`?
* Do you have a working test case for converting from `anndata` to `seurat`? I've been getting errors from `py_to_r` being called on an R data frame which I believe `reticulate` is auto-converting from a pandas data frame.

andrewwbutler:
Unfortunately, we don't currently have any tests using python. I did find [this vignette](https://cran.r-project.org/web/packages/reticulate/vignettes/package.html) when googling the problem. Seems like a reasonable solution for skipping tests when the module isn't available (which may be the case on CRAN servers). However, we can probably get it working for these CI tools.  There is an example in `RunPHATE` but again there's a check here to only run if the module gets successfully installed:
https://github.com/satijalab/seurat/blob/c67472cb047d139faf46ea5a659301cb542e3d91/R/dimensional_reduction.R#L1109-L1115

For the conversion tests:

* I would recommend building a new object for this set of tests as that will ensure that it's always consistent with the most recent Seurat version (even if we forget to update `pbmc_small`).
* Both `pbmc_small` and the [3k pbmc object](https://www.dropbox.com/s/kwd3kcxkmpzqg6w/pbmc3k_final.rds?dl=0) from the online tutorial work on my end. Granted those both have non-empty `dr` slots but that would be easy to modify/test as you showed above.

ivirshup:
Hey, just got back to this. 

Right now I'm trying to make sure calls to python are definitely tested. I don't think the examples haven't been running on linux, which I'm trying to look into now. I'm wondering if it would be worth it to try and handle as much of the python environment as possible through `reticulate` (like that vignette suggests), since that's the installation which needs to work. 

* Sorry about trashing your build queue, but I can't get the whole build to run though my personal travis account, I think due to the long build time and not having a cache. Any tips on getting that running?
* Just to be sure I understand what you're able to run, does `Convert(Convert(pbmc_small, "anndata"), "seurat")` work on your machine?

andrewwbutler:
Don't worry about the build queue. That's what it's there for. One way to potentially speed up build times on linux is to install some from binary packages. See the `r_binary_packages` field [here](https://docs.travis-ci.com/user/languages/r/#additional-dependency-fields).

 The double `Convert` does not work for me either.

ivirshup:
I've (finally)  managed to get the travis builds going!

Do you have a working case of `Convert({anndata_object}, "seurat")`? I'd hate to start editing it and lose functionality.

ivirshup:
Great, that looks like it works.

I'm thinking I might factor this out into a couple pull requests since I'm pretty busy for the next couple weeks. I'm thinking the first PR could just be getting the CI testing working for python dependencies. The only remaining issue blocking this is appveyor, where 32-bit builds are failing. When I use the same `appveyor.yml` file on a different repo, no 32-bit build is even attempted (I think this comes from the `R_ARCH: x64` line). Any idea what's up with that?

PR #534, opened 2018-06-08
Don't lose NA with cusom color scale

olgabot:
Currently, in `TSNEPlot` all cells labeled "NA" disappear with a custom color scale (`cols.use` is not `NULL`). This can be remedied with:

```R
p = TSNEPlot(tiss)
p = p + scale_color_manual(values=colors.use, na.value="grey50")
```

This PR adds the `na.value` option to dimplot and propagates it down to the `scale_color_manual` call


mojaveazure:
Thanks Olga!

PR #561, opened 2018-06-22
Remove All PKG_LIBS Flags

evolvedmicrobe:
A previous pull request (https://github.com/satijalab/seurat/pull/466) removed the unused `$(FLIBS)` flag from the Makevars file, which was a mistake because there are circumstances where linking to `$(LAPACK_LIBS) $(BLAS_LIBS)` depends on fortran and so can create compile errors.  This new commit removes all of these unused flags, rather than just one.  Thus avoiding unneccesary linkage, fortran errors AND `R CMD CHECK` errors.

For background, in older versions of Rcpp, these compile flags were required.  However, Rcpp has since been updated to avoid this issue.  Their FAQ (http://dirk.eddelbuettel.com/code/rcpp/Rcpp-FAQ.pdf) describes the issue

    2.15. **What about the new ‘no-linking’ feature.** Starting with Rcpp
    0.11.0, functionality provided by Rcpp and used by packages
    built with Rcpp accessed via the registration facility offered by
    R (and which is used by lme4 and Matrix, as well as by xts and
    zoo). This requires no effort from the user / programmer, and
    even frees us from explicit linking instruction. In most cases, the
    files src/Makevars and src/Makevars.win can now be removed.
    Exceptions are the use of RcppArmadillo (which needs an entry
    PKG_LIBS=$(LAPACK_LIBS) $(BLAS_LIBS) $(FLIBS)) and
    packages linking to external libraries they use.
    But for most packages using Rcpp, only two things are required:

    • an entry in DESCRIPTION such as Imports: Rcpp (which
    may be versioned as in Imports: Rcpp (>= 0.11.0)), and

    • an entry in NAMESPACE to ensure Rcpp is correctly instantiated,
    for example importFrom(Rcpp, evalCpp). The name of the symbol does not really matter.

Rcpp version 0.11.0 was released years before the current version of R required by Seurat (3.4.0), so although this requirement is implicit, this commit also specifies the version requirement.

Sorry for the mistake last time!


mojaveazure:
Thanks Nigel!

PR #573, opened 2018-06-27
Transform `data` also for single cells in AverageExpression

### LINKED ISSUE #571: AverageExpression discrepancy
I think there is a discrepancy in the values returned by AverageExpression when using the "data" slot when there are some identities which contain only a single cell.  When there is a single cell for an identity, and the "data" slot is selected, the log normalized values are returned corresponding to that single cell. For all other identities (with more than one cell), the function uses mean(expm1(x)) which gives the non-log normalized averages. 

This is related to #193

mschilli87:
This fixes a discrepancy in AverageExpression:
When `use.scale` and `use.raw` are both FALSE (default), identities with
several cells were transformed back (exp(x) - 1) before averaging while single
cell identities were neither averaged (no need) nor transformed (discrepancy).
This commit fixes this by transforming the data data conditionally in the
scenario described above.
It uses `if(!(use.scale | use.raw))` assuming this would be faster to evaluate
than `if(use.slot == "data")` or applying `mean` to a single value to safe the
conditional altogether. This performance assumption, however, was not tested.

fixes #571

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -592,6 +592,10 @@ AverageExpression <- function(
       genes.assay <- unique(x = intersect(x = genes.assay, y = rownames(x = data.use)))
       if (length(x = temp.cells) == 1) {
         data.temp <- (data.use[genes.assay, temp.cells])
+        # transform data if needed (alternative: apply fxn.average to single value above)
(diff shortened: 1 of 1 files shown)


andrewwbutler:
Thanks Marcel!

PR #584, opened 2018-07-02
Fix error when cell has exactly same value the two scores

lpantano:
Hi,

thanks for develop this awesome package! We use it a lot. 

When using `CellCycleScoring `, we (I and @marypiper) found a weird case where the initial scores of some samples got the same values and it was returning NA values for the score for all the samples for cellcycle column.

We think the problem is here: `c(first, second)[which(x = scores == max(scores))]` that will return two values if the cell happens to have the same value.

We assume that `assignments` should be a vector not a list, but if the returned value from `apply` is more than 1 element, it would be a list and it would get a bad `data.frame` conversion, where columns are samples instead of rows.

Choosing the first value in case more than one is giving, solves the issue:

```
c(first, second)[which(x = scores == max(scores))][1]
```

Sorry if the message is confused, let us know what you think and if it is right to make this fix.

Thanks for your time in advance!


lpantano:
although this will inflate the first type, maybe another solution is better.

lpantano:
Thanks! undecided has a lot of more sense!

PR #593, opened 2018-07-05
Update PHATE API for phateR v0.2.8

scottgigante:
I've also modified the HDF5 install to reduce the log length in travis.


mojaveazure:
Thanks Scott!

Issue #547, opened 2018-06-15
Feature Request: GetIdent()


wyattmcdonnell:
Hi there!

The `SetAllIdent()` function is super useful, and it would be great if at some point in the future a similar function `GetIdent()` could be incorporated. Especially as larger datasets such as those from HCA begin to roll out, such a function would be good for users to check that they're matching the right identity before making a huge plot, doing differential expression, etc. 

Skeleton code for this function could be as simple as an alias to something like this: 

```R
unique(seurat.object@ident)
```

Or more complex by using output from above command, and seeing which vector in `seurat.object@meta.data` has unique levels that match. 

Thank you so much!

mt1022:
I am considering writing a simple `GetIdent` function. Here is how it looks like:
```r
GetIdent <- function(object, uniq = TRUE, cells.use = NULL){
    ident <- object@ident
    if(!is.null(cells.use)){
        ident <- ident[names(ident) %in% cells.use]
    }
    if(uniq){
        unique(ident)
    }else{
        ident
    }
}
```
This function will return unique `ident` by default and allows specifying the subset of cells to look at. Are there any other features that should be incorporated?

wyattmcdonnell:
Hi @mt1022 thanks for considering this! I don't have any other thoughts on extra things to incorporate, but maybe @mojaveazure, @andrewwbutler, or others do?

mt1022:
Fine. If no other requirements, I'll try to make a PR on open souce Fridy.

PR #596, opened 2018-07-06
add GetIdent function

mt1022:
I write a simple `GetIdent` function for this feature request (#547). This function will return unique identity values by default and allows specifying the subset of cells to look at. This function is documented and exported. I tested it locally and it works fine.


mojaveazure:
Thanks a bunch!

PR #599, opened 2018-07-07
Remove Java Dependency

evolvedmicrobe:
Seurat currently uses the Java package, `ModularityOptimizer.jar,` available from http://www.ludowaltman.nl/slm/ to perform clustering, and invokes this java application by writing an output file and then using a system call.

Although the use of Java enables code reuse and allows any errors in the clustering process to take place in a separate process that has robust memory and error management, the dependency on Java also creates installation difficulties.  Distributing Seurat then requires a Java installation (which can consume a fair amount of space in things like docker images), and also requires appropriate setting of the PATH environmental variable (which is confusing for neophytes).

Because the ModularityOptimizer Java code is 1) largely written in “C syntax” and thus easily translated to C++, 2) Does not appear to still be under development (last update 8/2015) and 3) Is the only Java code used by the Seurat package, this commit replaces this Java dependency by re-writing the ~1K lines of Java code with C++11/Rcpp as is used elsewhere in the package.

In translating the code, the interface was maintained, such that the although the programming languages changed, the results are identical.  As a consequence,  rather than rewriting the program to be more idiomatic Rcpp/C++, both the output and the code was kept as close to the original Java as possible.  For example, in order to maintain compatibility, the Java random number generator was implemented in C++, and `std::stable_sort` was used in place of `std::sort` to match the output results of the Java code exactly.  This code also avoids the need to first write a temporary file that is taken as input to the Java program and then deleted.  However, the interface still accepts a name for this (no-op) file operation.

In order to test and verify that the C++ and Java code return the same results, the C++ code added also includes a mirror of the original command line interface.  This version of the program can be compiled with a command such as:

    clang++ -O3 -std=c++11 -DSTANDALONE -Wall -g ModularityOptimizer.cpp

And can then be called in an identical fashion to the original Java version to verify compatibility or identify any issues in the input/output.

As expected, when revisiting a codebase, some brittle behavior was identified in the original Java code.  The Rcpp interface thus does differ in that validation checks are added to avoid errors that would crash the original Java program and produce error messages instead.

Testing was performed to verify the Java code and C++ code produces the same output on a range of possible inputs.  A subset of the tested scenarios was added as unit tests (the tests taking << 1 second time) to ensure validity going forward.


evolvedmicrobe:
Hi Guys,

I encountered some issues today and put this together.  Looks like it hit some hiccups, will check them out in the next few days and then make this pull active again, for now, it's a `[WIP]`.

Cheers,
Nigel

satijalab:
Hi Nigel - Again, this is fantastic, thanks so much! Let us know when its ready, and we look forward to trying it out.

best,
Rahul

evolvedmicrobe:
Hi Rahul,

Great! Really glad you guys are open to this, I think it will help out a lot on the installation/distribution front.  I made a few more edits and added some more tests, so it's ready for review.

Cheers,
Nigel

andrewwbutler:
Thanks Nigel! Definitely agree that this will be a great help with installations/distribution.

PR #610, opened 2018-07-11
Reduce Memory Usage in Standardize Function

evolvedmicrobe:
Inside the function RunMultiCCA, there is the following line:

    if(standardize){
        for (i in 1:num.sets){
          mat.list[[i]] <- Standardize(mat.list[[i]], display_progress = F)
        }
      }

Which in turn calls the Rcpp function `Standardize`.  It appears the matrix located in `mat.list` can be large and so this function has some very expensive memory allocations as it does the following:

1 - Copies the original matrix into a `Eigen::MatrixXd` type to pass as an argument (2X memory usage)
2 - Makes another `Eigen::MatrixXd` matrix of the same size to store results in.
3 - Finally, copies that result `Eigen::MatrixXd` into a `NumericMatrix` type to pass back into R.

As a result, the code can contain up to 4 times the size of the original matrix at once.

This pull reduces this memory consumption by making the following changes.  First, rather than copying the matrix to pass it in, we use a Eigen::Map<> to simply use the original underlying data without modifying it.  Next, rather than create an EigenMatrix that is copied to a NumericMatrix, it uses the NumericMatrix type directly to store the data at the end.  This avoids significant memory consumption for large matrices, and helps avoid out of memory errors. 

In a simple test of the following, the time dropped from to 10.8 (s) to 2.6 (s) on my MacBook, with the reduced memory usage matching the expectation.

    library(Seurat)
    n = 21000
    q = matrix(nrow=n, ncol=n, 1)
    system.time({g = Seurat:::Standardize(q)})


andrewwbutler:
Thanks again Nigel!

Issue #618, opened 2018-07-16
Error in ScaleData


vthaker:
Hi there,
We are using Seurat to analyze a dataset of 13,000 cells from mouse brain prepared in 9 batches. 
We have ~2658 variable genes in the @var.genes. Could you please advise on overcoming the error we encounter. Thanks
```R
objectname <- ScaleData(object = name, vars.to.regress = c("percent.mito", "orig.ident", "nUMI"), genes.use = name@var.genes)

[1] "Regressing out percent.mito" "Regressing out orig.ident"   "Regressing out nUMI"        
  |===================================================================================| 100%
Time Elapsed:  4.00979118347168 mins
[1] "Scaling data matrix"
  |                                                                                   |   0%Error in scaled.data[genes.use[my.inds], ] <- data.scale : 
  number of items to replace is not a multiple of replacement length

sessionInfo()
R version 3.4.3 (2017-11-30)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS High Sierra 10.13.4

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.4/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] bindrcpp_0.2.2 dplyr_0.7.4    Seurat_2.3.0   Matrix_1.2-14  cowplot_0.9.2  ggplot2_2.2.1 

loaded via a namespace (and not attached):
  [1] diffusionMap_1.1-0   Rtsne_0.13           VGAM_1.0-5           colorspace_1.3-2    
  [5] ggridges_0.5.0       class_7.3-14         modeltools_0.2-21    mclust_5.4          
  [9] htmlTable_1.11.2     base64enc_0.1-3      proxy_0.4-22         rstudioapi_0.7      
 [13] DRR_0.0.3            flexmix_2.3-14       prodlim_2018.04.18   mvtnorm_1.0-7       
 [17] lubridate_1.7.4      ranger_0.9.0         codetools_0.2-15     splines_3.4.3       
[...]

PR #619, opened 2018-07-17
Warn on NA data being used for regression

evolvedmicrobe:
If the data used for the regression contains NA values, this will lead to errors downstream.  For example, if one row contains an NA value, it will be removed and create a size mismatch when a data.frame is later constructed using the residuals of all rows, or if a linear model is used, it will create a problem when the QR decomposition has different rows than the value being regressed against.  This notifies the user of the underlying issue before a harder to debug error message appears.

Motivated by:
https://github.com/satijalab/seurat/issues/618


andrewwbutler:
Thanks Nigel!

Issue #626, opened 2018-07-18
RunMultiCCA() fails to detect duplicated cell names early on


HenrikBengtsson:
There's a [thinko](https://github.com/satijalab/seurat/blob/master/R/dimensional_reduction.R#L835-L839) in how `RunMultiCCA()` tries to protect against duplicatd cell names early on;

```r
  names.list <- lapply(object.list, slot, name = "cell.names")
  names.intersect <- Reduce(intersect, names.list)
  if(length(names.intersect) > 0) {
    stop("duplicate cell names detected, please set 'add.cell.ids'")
  }
```

For instance, if `intersect(object.list[[1]], object.list[[2]])` is empty, then it doesn't matter what `object.list[[3]]` etc. contains.

satijalab:
Closing this issue as we are now moving away from the use of RunMultiCCA - we do improved error checking in the v3 workflow - thanks to your comment

PR #639, opened 2018-07-21
BUG FIX: RunMultiCCA() failed to detect duplicated cell names [#626]

HenrikBengtsson:
cf. Issue #626


mschilli87:
Isn't this still a valid bug? Why won't it be fixed if there is a fix around?

mschilli87:
@satijalab: see https://github.com/satijalab/seurat/pull/725#issuecomment-437897720 :wink:

Issue #622, opened 2018-07-17
Suerat Installation erroring out


rarvindk:
[R_studio_341_2.txt](https://github.com/satijalab/seurat/files/2203151/R_studio_341_2.txt)

Hello Suerat Team,

We are trying to install “Suerat” package on one of the HPC environment. We are constantly running into errors. I’m attaching a txt file with the capture of the screen while installing the Seurat package with in R 3.4.1. Could pls let us know if there is anything we are missing or your assistance is greatly appreciated.

romanhaa:
My personal advice is to use either conda environments or (better) containers. I can't really tell what the problem is from the `txt` file you uploaded, but there could be all sorts of library incompatibilities that are hard to overcome on HPC if you are not an admin (I've had my experience with this). So, check out conda environments which are easy to set up and overcome most library problems. However, some libraries kept making trouble, including `cairo`. I used conda environment quite extensively for a while on our HPC but eventually switched to Docker + Singularity because it gives me full control. Since then, I've had 0 issues installing any R package. In a nutshell, you create your Docker container on your local machine and then convert that to a Singularity image (not necessary if you have a Docker license for Linux machines I think).

Hope this helps.

rarvindk:
Hello,

Thanks for your time and support. Greatly appreciate you assist. This is
the client environment and we have very limited authority to change the
environment for now. I'll check on condo environment and see what best
assist can I provide.

-Arvind

HenrikBengtsson:
We ran into the same problem where we / users couldn't easily install [hdf5r](http://cran.r-project.org/web/packages/hdf5r) due to lacking or outdated hdf5 system libraries.  
If you don't need to read 10X HDF5 files using `Seurat::Read10X_h5()` then a workaround is to install the fake hdf5r package that I've created - see https://github.com/UCSF-TI/fake-hdf5r

To Seurat devels: By moving `hdf5r` from Imports to Suggests and update `Seurat::Read10X_h5()` to:
```r
  requireNamespace("hdf5r", quietly = TRUE) || stop("Please install the 'hdf5r' package")
  infile <- hdf5r::H5File$new(filename)
```
you'll make 'hdf5r' optional and helps users with this particular issue.

sunwu93:
Others and myself also encountered this problem trying to update seurat on our HPC systems. Thanks @HenrikBengtsson, did the trick!

HenrikBengtsson:
Ping

evolvedmicrobe:
[...]

PR #641, opened 2018-07-21
Make hdf5r optional, i.e. move from Imports to Suggests

HenrikBengtsson:
cf. Issue https://github.com/satijalab/seurat/issues/622#issuecomment-406057861


evolvedmicrobe:
I also think this is a very good idea.  It might be worth considering removing the hdf5 installation from the testing script at the same time, as the hdf5 dependency also makes the CI tests take a long time.

HenrikBengtsson:
ping

Issue #647, opened 2018-07-23
RunCCA doesn't check for compatible input and has incomplete documentation


TomKellyGenetics:
I tried to use `RunCCA` to combine 2 datasets and perform canonical correlation analysis to evaluate batch effects. However, I received the following error message:

```
> RunCCA(object1, object2, group.by = "group")
0 valid genes in genes.use
```

This error was not helpful. Giving `genes.use = union(object1@var.genes, object1@var.genes)` did not resolve this issue nor did re-running `FindVariableGenes` on each object.

This issue arose because I had neglected to run `ScaleData` on an object. This also produces this error message. I was able to identify and troubleshoot this from viewing the source code, since the default `genes.use` is compared to `possible.genes` derived from `rownames(object@scale.data)` (which does not exist when the data has not yet been scaled) rather than `rownames(object@data)` . 

This error message is therefore terse and is unhelpful considering that many less inexperienced R users perform Seurat analysis since the package is so popular. I would suggest to include more checks for compatible data inputs and warning or error messages if these data inputs are missing. While the vignette includes the appropriate data scaling, it may be helpful to mention that this requires a scaled object by default in the R documentation. I would also suggest to mention in the `RunMultiCCA` documentation that `RunCCA` is recommended if only 2 datasets are being compared. 

It would also be helpful to make "group" be the default for `group.by` (for `RunCCA`) and `grouping.var` (for `AlignSubspace`). It is not clear why there is no default for these, why the group arguments differ between them, or why `RunCCA` suggests to "add.cell.ids"  (an argument in `RunMultiCCA` and `MergeSeurat` but not this command) when a grouping variable would resolve this problem and is already supported by the Seurat object Metadata.

Overall these are minor issues but addressing them would make performing CCA easier.

satijalab:
Thanks for the feedback - it would be wonderful if you would consider submitting a PR with modified documentation, and/or compatibility checks. 

Hopefully, when you were able to run the alignment procedure, it worked well for you.

TomKellyGenetics:
I'm still experiencing some difficulties but I think that may be due to the experiments that we are still optimising. My colleague has performed this analysis on another project and recommends this package enthusiastically. Overall, I think it is very good but I encountered some minor issues running this analysis and have suggested changes with consideration for the popularity of the package and that this would cause more difficulties for inexperienced R users.

yueqiw:
@satijalab I had experienced similar issue to this one. 

[...]

PR #649, opened 2018-07-24
Enhance user experience with RunCCA and AlignSubspace

TomKellyGenetics:
Implements changes suggested in Issue #647

Deprecates `grouping.var` for consistent `group.by` in both commands. Adds checks for compatible input data (Seurat Objects) with `scale.data` and `meta.data` slots.

Changes have not been made to `RunMultiCCA` and `MergeSeurat` but are suggested for consistency (e.g., `group.by` or `group1` instead of `add.cell.id1`).


TomKellyGenetics:
Since these arguments did not previously have default behaviour, these changes should be backwards compatible with previous versions of Seurat and not change the output of existing scripts.

PR #695, opened 2018-08-13
Use custom gene names and cells names

lambdamoses:
There're several things I have changed here, for the loom branch. First, the original loom branch often assumes that the gene names are stored in "row_attrs/gene_names" and cell names are stored in "col_attrs/cell_names". Actually the functions `GetAssayData` and `GetDimReduction` have arguments that allow us to specify custom names for datasets for gene names and cell names, but other functions that call them do not give us the opportunity to specify and assume the default names. I got a dataset where the cell names are in "col_attrs/CellID" and I want to use transcript accessions in "row_attrs/Accession" rather than names since there're multiple isoforms of the same gene, so I added the `gene.names` and `cell.names` arguments to `RunPCA` and `RunTSNE` so we can use custom datasets names.

Another important thing I changed is to make `FindVariableGenes.loom` store parameters as attributes to `row_attrs/var_genes`, for I think that's the most sensible place to store such parameters. I have made some other changes to fix what I perceived to be bugs. See the commits and diffs. Now the modified version of Seurat for loom works for my purposes. There's another thing to note: I often got warnings from loom when running PCA that batch.next and batch.scan are deprecated and use map and apply instead. I did not fix this since it doesn't seem to affect results, but in order to get the new version of Seurat to CRAN, you'd better fix this.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -110,7 +110,7 @@ Collate:
     'utilities_generics.R'
     'utilities_internal.R'
     'zfRenderSeurat.R'
-RoxygenNote: 6.0.1
+RoxygenNote: 6.1.0
 Suggests:
     testthat,
     S4Vectors,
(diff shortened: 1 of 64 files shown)


dcollins15:
Thanks for using Seurat!

PR #708, opened 2018-08-18
Prepare Seurat for CellRanger 3.0

evolvedmicrobe:
10X is releasing a new version of CellRanger that is changing the output format.  This pull request makes Seurat forward compatible with the new version.  In particular, the following changes are made:

**HDF5 Format** - The format of this file is changing.  As Seurat is making HDF5r optional, the `Read10X_h5` function will no longer be the preferred way to load data, and we have added an error message if newer (and incompatible files are loaded.)

**Text File Formats** - In order to save disk space, the sparse matrix and barcode text files will now be gzipped.  As R automatically identifies and correctly reads gzipped files, no changes were needed to account for this other than appending a suffix when necessary.  Additionally, in order to account for experiments that have "multimodal" datasets, the `gene.tsv` will instead become the `features.tsv` file.  This file will contain an additional column describing the type of feature referred to in that row of the matrix.

In order to help users follow the [multimodal vignette](https://satijalab.org/seurat/multimodal_vignette.html) and analyze this type of data, rather than return one matrix, with multimodal data a list of matrices are returned, with the name of each element in the list corresponding to the type of data.  As this new data is distinct from the older types of data shown in past tutorials, we print a message when this occurs so the user is aware that they have different data types.

Very small test data files and associated tests were added to verify the expected behavior.


satijalab:
Hi Nigel - thanks very (very!) much we reached out to you on your gmail to discuss this and additional future plans offline- let us know if you didn't receive or if there is a better address to reach you at.

evolvedmicrobe:
Hi @satijalab, sorry for the slow response, I just wrote back from my more "official" email address after rebooting following a family vacation.  Hope to continue the dialog.

Warm wishes,
Nigel

PR #725, opened 2018-08-22
Multi cca improvements

ekernf01:
Hi Seurat devs, 

I have made some changes to the `RunMultiCCA` function. If you'd like to incorporate these changes, I think they will bring `RunMultiCCA` better in line with Seurat's typical ways of doing things. 

- `object.list` can now be a Seurat object, and for this case, I added a `group.by` arg to specify groups.
- The data scaling after creating `combined.object` takes up tons of memory, and the resulting scaled data isn't immediately necessary. I changed it to scale only the genes given in `genes.use`.
- Since CCA's latent components are ranked and they explain different amounts of variation in the data, I added a `weight.by.var` option mimicking `RunPCA`.
- I added unit tests for the new way of calling `RunMultiCCA`.

Hope this works well for you! Let me know if there's any small changes that would make this request more acceptable to you.
Best regards,
Eric


mschilli87:
@satijalab: I suggest you reconsider your bot settings - A bot that closes good community contributions because the maintainer didn't find the time to review it yet is kind of the worst case scenario.
Same is true for valid open issues.
How about introducing a tag 'waiting for reply' or something like that which you manually add when *you* are waiting for an answer and have the but close stale issue/PRs *with that tag only*. Ideally, as soon as the issue/PR has *any* reply, the bot could remove the tag because the ball is back in your hands then. :wink:

dcollins15:
Thanks for using Seurat!

PR #735, opened 2018-08-26
Seurat implementation of ALRA

linqiaozhi:
@JunZhao1990 and I implemented Adaptively-thresholded Low Rank Approximation (ALRA), a [method](https://www.biorxiv.org/content/early/2018/08/22/397588) for imputation of dropped out expression values in scRNA-seq data, for Seurat v3.0. 

We felt that it made the most sense to include the RunALRA functions in preprocessing.R, as imputation could be considered a kind of preprocessing. The most important parameter for ALRA is the approximate rank of the matrix, which can be chosen automatically using the statistics of the spacings between singular values, as described in the paper. Some users may want to inspect the spectrum themselves, and hence we also provide an ALRAChooseKPlot() function in the visualization.R file. 

A simple example of usage is provided in the documentation for RunALRA(). 

Please let us know if there is any way we can improve the code or make it more in keeping with the conventions established for Seurat. We did our best to follow the practices on the wiki and also to look at other functions, but as this is our first contribution to Seurat, there is likely to be room for improvement.


JunZhao1990:
It looks like our PR fails the CI checks. For example, we have the below error:
image

We think it is probably due to an incompatibility between Seurat 3.0 and the tests in 'Seurat-Ex.R'. Specifically, the AddMetaData() function in Seurat 3.0 can't be applied to an old Seurat object. We also notice that several recent commits to Seurat 3.0 have the same issue (https://ci.appveyor.com/project/satijalab/seurat/build/1.0.485).

andrewwbutler:
Hi @linqiaozhi and @JunZhao1990,

Thanks for putting this together! You were right about some of our examples causing problems with the CI's. Most were caused by trying to use the new object structure with the old testing object (`pbmc_small`). We've updated our examples and `pbmc_small` so those should be working now. 

I think there may still be a couple issues with your examples (probably due to `pbmc_small` being such a small dataset) but hopefully it should be easy to just adjust some of the parameters to get them to run.

linqiaozhi:
Great, thanks @andrewwbutler! 

Yes, we did not realize that pbmc_small was so small (80 samples by 230 genes)! Our default settings assume that the matrix is at least 100 by 100, and it gives an error otherwise. We can change the parameters (K, noise_start) in the example for it to run, but we are wondering if that would be confusing to people. For the vast majority of datasets, those parameters need not be changed, and if we have them set differently in the example (i.e. visible upon ?RunALRA), people might be confused.

Are there any other test datasets the system has access to? For example, the pbmc3k dataset is pretty small but seemed like a nice example for ALRA. Or perhaps we should wrap this example in \notrun tags so we don't get this error and write a separate test with the parameters set such that pbmc_small will work? 

Of course, you guys know best--if you think it important that the example uses pbmc_small, then we can change the parameters in the example, no problem. Just trying to think about what will make it least confusing for users.

linqiaozhi:
Actually, we were discussing, and we think a better solution is to change our code to automatically set K and noise_start based on the dimension, instead of hard-coding them. It will use the same defaults if the matrix is larger than 100 by 100, but if it's less, then it will use parameters based on the smallest dimension, and just give a warning. 

This way the example can run without having to specify the parameters, but the defaults stay the same for all except the smallest matrices.

So, we will go ahead and make those changes and then commit.

andrewwbutler:
I like the parameter auto-setting solution! Thanks for updating the examples!

PR #788, opened 2018-09-21
Fixes to `AlignSubspace()` function

robertamezquita:
- Added to @importFrom pbapply the pbsapply function
- fixed reference to `sdi` reference (ScaleData() was returning the Seurat object, not a matrix; subsequent reassignment of `NA` values now works
- fixed a typo in reference to objects[[i]] (was missing an "s")

### DIFF
--- a/R/integration.R
+++ b/R/integration.R
@@ -48,7 +48,7 @@ globalVariables(
 #'
 #' @importFrom dtw dtw
 #' @importFrom stats density
-#' @importFrom pbapply pbapply
+#' @importFrom pbapply pbapply pbsapply
 #' @importFrom graphics plot lines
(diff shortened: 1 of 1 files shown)


robertamezquita:
#789

mschilli87:
@robertamezquita: Why did you close the PR? Where these things fixed already elsewhere (where?) or are they not considered bugs anymore (why?)?

robertamezquita:
Hi @mschilli87 the developers of Seurat closed the issue I had submitted #789 because they will be replacing it wholesale with new methods that are not yet public. For now, the current version Seurat 2.3 has the AlignSubspace() function that works.

Issue #449, opened 2018-04-30
maximal number of DLLs reached 


topherconley:
If you could find a way to not attach so many libraries so that I would not get this error:

> Error in dyn.load(file, DLLpath = DLLpath, ...) : 
  unable to load shared object '/usr/local/lib/R/3.4/site-library/yaml/libs/yaml.so':
  `maximal number of DLLs reached...

There has got to be a genuine solution to this problem. At present, combining Seurat with other packages attached to the workspace does not play nicely. Is there something you can do in managing the NAMESPACE or something so this doesn't happen?

mojaveazure:
Unfortunately, we cannot decrease the number of DLLs Seurat loads. However, you can increase the number of DLLs you have loaded. In you're `~/.Renviron` file, set `R_MAX_NUM_DLLS` to be something greater than 100 (the default), eg. `R_MAX_NUM_DLLS=200`

topherconley:
Thanks!

mschilli87:
@mojaveazure:
> Unfortunately, we cannot decrease the number of DLLs Seurat loads.

Maybe [the import package](https://github.com/smbache/import) could be used to use functions from other package within Seurat without pulling them in completely.
As a disclaimer I should add that I have never tried this and probably even when not attaching the packages namespace to to the global one that wouldn't help here as the DLL might still need to be loaded (I have no clue how this works internally in R). But is there really no way to statically link in part of a packge into another one in R?

PR #790, opened 2018-09-22
Feat/subset by barcode distribution

robertamezquita:
new feature - subset by barcode distribution

- calculates inflection point of barcode distribution (the "knee") at which to subset a seurat object; performs this calculation with respect to a grouping variable, such as sample origin (currently `orig.ident`)
- user can specify the column to use for the calculation (umi.column = 'nCount_RNA' by default)
- user can specify barcode rank range at which to perform the calculation (threshold.low, threshold.high)
- user can diagnose at various stages with returnable ggplot2 plots, the data.frame with calculated inflection, and also simply retrieve the cells which passed the inflection filter to perform their own subsetting as needed


robertamezquita:
Additionally also added some of the tidyverse utilities I made use of, such as the pipe and other libraries

R/preprocessing.R
```diff
@@ -1526,3 +1526,165 @@ RegressOutNB <- function(
...
+SubsetByBarcodeDistribution <- function(object,
+                                       umi.column = 'nCount_RNA',
+                                       threshold.low = NULL,
+                                       threshold.high = NULL,
+                                       show.plot = FALSE,                                       
+                                       return.plot = FALSE,
+                                       return.cells = FALSE,
+                                       return.inflections = FALSE) {
+    ## Make sure there is a column for "cell" to subset the object by
+    if (!("cell" %in% colnames(object@meta.data))) {
```
mojaveazure, line 1582: Object cell names can be found with `colnames(x = object)`

R/preprocessing.R
```diff
@@ -1526,3 +1526,165 @@ RegressOutNB <- function(
...
+#'
+#' @return Return Seurat object subsetted. Appends to the workflow slot 'nCount_RNA_threshold'
+#' with a `data.frame` showing the calculated thresholds per sample group that determine the passing cells.
+#'
+#' @importFrom dplyr mutate arrange group_by summarise select inner_join filter
+#' @importFrom tidyr nest
+#' @importFrom purrr map_int map2_dbl map2
+#' @importFrom ggplot2 ggplot geom_line theme_bw geom_vline
+#' @importFrom rlang sym
+#' @import tibble
```
mojaveazure, line 1562: One thing we're working on with Seurat v3 is reducing general `@import` in favor of explicit `@importFrom`; could you specify what from tibble you're importing here?

R/preprocessing.R
```diff
@@ -1526,3 +1526,165 @@ RegressOutNB <- function(
...
+#'
+#' @rdname Preprocessing
```
mojaveazure, line 1564: The rdname should be the same name as the function

R/preprocessing.R
```diff
@@ -1526,3 +1526,165 @@ RegressOutNB <- function(
...
+                                       umi.column = 'nCount_RNA',
+                                       threshold.low = NULL,
+                                       threshold.high = NULL,
+                                       show.plot = FALSE,                                       
+                                       return.plot = FALSE,
+                                       return.cells = FALSE,
+                                       return.inflections = FALSE) {
+    ## Make sure there is a column for "cell" to subset the object by
+    if (!("cell" %in% colnames(object@meta.data))) {
+        object@meta.data$cell <- rownames(object@meta.data)
```
mojaveazure, line 1583: Please use the `[[` and `[[<-` operators for getting and setting meta data

R/preprocessing.R
```diff
@@ -1526,3 +1526,165 @@ RegressOutNB <- function(
...
+
+    ## Check that umi.column exists in meta.data
+    if (!(umi.column %in% colnames(object@meta.data))) {
+        stop("`umi.column` specified not present in Seurat object provided")
+    }
+
+    ## Grab meta data
+    tmp_meta <- object@meta.data %>%
+        as.data.frame() %>%
+        dplyr::group_by(orig.ident) %>%
```
mojaveazure, line 1597: Anything that was declared with `@importFrom` doesn't need the `::` to call it

R/preprocessing.R
```diff
@@ -1526,3 +1526,165 @@ RegressOutNB <- function(
...
+                dplyr::summarise(
+                    inflection = which(rawdiff == min(rawdiff[threshold.low:threshold.high], na.rm = TRUE))
+                )
+            y[[1]]
+        })) %>%
+        dplyr::select(-data)
+
+    ## Calculate inflection point plot for QC
+    if (show.plot == TRUE | return.plot == TRUE) {
+        g <- tmp_meta %>%
```
mojaveazure, line 1626: Could you separate the plotting aspect into a separate function?

R/preprocessing.R
```diff
@@ -1526,3 +1526,165 @@ RegressOutNB <- function(
...
+                x <- dplyr::filter(d, rank < i)
+                min(x[umi.column], na.rm = TRUE) # return the min nCount threshold used
+            }),
+            cells_pass = purrr::map2(data, inflection, function(d, i) {
+                x <- dplyr::filter(d, rank < i)
+                return(x$cell) # return passing cells
+            }))
+
+    ## Break: returns the calculated inflection points data and passing cells
+    if (return.inflections) {
```
mojaveazure, line 1664: Functions should strive for a single return value that isn't dependent on a parameter, could you put all extra things in object meta data or tools?

R/preprocessing.R
```diff
@@ -1526,3 +1526,165 @@ RegressOutNB <- function(
...
+    ## Get vector of which cells to keep
+    cells_pass <- unlist(cells_filtered$cells_pass)
+
+    ## Break: returns the names of the cells that pass only
+    if (return.cells == TRUE) {
+        return(cells_pass)
+    }
+
+    ## Subset the object object
+    object <- SubsetData(object, cells = cells_pass)
```
mojaveazure, line 1677: `SubsetData` will be deprecated at some point in the near future, could you use `subset` or the `[` operator instead? Please see `?subset.Seruat` for details

R/utils-pipe.R
```diff
@@ -0,0 +1,11 @@
+#' Pipe operator
```
mojaveazure, line 1: I'm not sure what this is for, could you explain what the purpose of this

robertamezquita:
Thank you for the thorough code review, I really appreciate it! I will work on resolving these various issues, and will follow-up with regards to the individual chunks shortly, but just wanted to say thanks in advance!

R/preprocessing.R
robertamezquita: Great, will do. One question re: the use of `tibble` though. The reason I use it is to organize some data into "list-columns" for easier access/annotation of data, which requires the use of tibble to view neatly (aka, without it printing a bunch of lines in an unreadable way). It seems like Seurat v3 doesn't really use the `tibble` format in general, so should I just get rid of this dependency and use `data.frame` instead? I personally prefer `tibble` to `data.frame` (better printing, list-columns are great), but will go with what you think is best.

R/preprocessing.R
robertamezquita: Could I get an example usage? I'm not familiar with this..thanks!

R/preprocessing.R
robertamezquita: For this case, since it is returning the inflection point cut-off per sample, would it be more appropriate then to return it to tools (in meta data, it would be pretty repetitious to have it be appended on a per cell basis).

R/preprocessing.R
robertamezquita: Do you think that directly subsetting the data and returning the subsetted object is most appropriate for the function? Or, should it be that it returns the `meta.data` slot with an appended column `cells_pass` with a boolean of whether it is above the "knee" or not? 

If so, is there a way to make it so that the new working data is the subsetted version (and thus doesn't lose the "bad cells")?

Probably easiest to just return the subsetted Seurat object that removes the "bad cells", but wanted to check in on this

R/utils-pipe.R
robertamezquita: This is just what's added by the `usethis::use_pipe()` function to enable pipe functionality throughout an R package. We can remove this since we are removing the pipe operator from the function.

R/preprocessing.R
mojaveazure: If the function is designed to subset (since it's called `SubsetByBarcodeFraction`, it sounds like a subsetting function to me), then it should return a subsetted object, following how Seurat subsets objects. `SubsetData`, for example, subsets all aspects of the Seurat object down to certain features and/or cells.

If the function was to simply calculate the inflection point, but not subset, then it should be titled differently. I personally think that as long as you're adding something to object meta data that users can subset on themselves, you can leave subsetting up to the user.

R/preprocessing.R
mojaveazure: Do whichever you think is best. The tools slot is designed to allow developers a list with free space to put whatever they want. We currently don't have setters/accessors for the tools slot, so use the `slot` function to access it. We'll take care of creating setters/accessors and updating code in Seurat to use such helper functions once we build them.

R/preprocessing.R
mojaveazure:
```R
colnames(x = pbmc_small[[]])
pbmc_small[['nCount_RNA']]
pbmc_small[['groups']] <- sample(x = c('g1', 'g2'), size = ncol(x = pbmc_small), replace = TRUE)
pbmc_small[['groups', drop = TRUE]]
```

R/preprocessing.R
mojaveazure: One of the criticisms of Seurat v2 was that we depended on and loaded too many dependencies, especially those that are compiled (see #449 or search our issues tracker for 'dll'). With Seurat v3, we're trying to reduce the number of dependencies, especially those that are compiled. [Tibble](https://cran.r-project.org/package=tibble) (and much of the Tidyverse) is compiled, so if you can get away with using `data.frame`s or structured lists, that'd be great.

If you feel that you absolutely need tibbles, then include it using the explicit `@importFrom` rather than the general `@import`

PR #820, opened 2018-10-02
new refactored code for barcode inflections calc/plot/subset #790

robertamezquita:
Refactored code per @mojaveazure comments to #790 - created a new branch that was more limited in scope in terms of code changes to first get feedback. New family of functions is:

CalculateBarcodeInflections
PlotBarcodeInflections
SubsetByBarcodeInflections


mojaveazure:
Hi Robert,

Thanks for the refactor. I've made a couple of stylistic changes to match Seurat's code style (still working on a style guide) as well as other minor changes:

For all functions, I've removed all references to the `@` accessor for getting slot information. We ask that developers not use the `@` as it allows us to restructure the Seurat object without needing to worry about code breaking. We provide accessor and setter functions for most aspects of the Seurat object and for slots that we don't provide functions for, we ask that the `slot` function be used. For example, object metadata can be accessed using the `[[` operator (eg. `object[[c(group.column, barcode.column)]]`. We currently don't provide an accessor/setter for the tools slot, so the use of `slot` works just fine here.

Another thing to remember is that all functions found in neither Seurat nor `base` must be imported. In this case, `aggregate` and `ave` are found in `stats`, and must be imported using the `@importFrom` tag in the documentation.

The `tools` slot is open to all developers to store data in. However, one shouldn't just overwrite it without ensuring all other tools are copied over. One way to get around this is `slot(object = object, name = 'tools')$CalculateBarcodeInflections <- info`; this way you're simply appending to the tools list or overwriting a previous call to `CalculateBarcodeInflections`, but not overwriting another function's storage of data in the tools slot.

I've renamed `PlotBarcodeInflections` to `BarcodeInflectionsPlot` to keep in style with Seurat's plotting functions as well as moved it to `visualization.R` with the other plotting functions. Greater detail about file organization can be found [here](https://github.com/satijalab/seurat/wiki/Source-Code-File-Structure-and-Organization). Likewise, I've [sorted ](https://github.com/satijalab/seurat/wiki/Source-Code-File-Structure-and-Organization#code-sorting) `preprocessing.R` so that `CalculateBarcodeInflections` and `SubsetByBarcodeInflections` are in their proper section and order.

For subsetting a Seurat object, the `subset` function has the arguments `x` for the Seurat object, `subset` for an expression of feature expression or metadata variables (eg. `MS4A1 > 3`, `group == 'g1'`), and `select` for a static list of cells or identity classes to keep. Since we're subsetting on a list of cell names, we can take a shortcut and call `object[, cells_pass]`, similar to matrix or data frame subsetting.

I've also required that all functions must be explicitly called rather than chaining by several functions together. This change means that users have to be aware when running your functions.

robertamezquita:
Thank you so much Paul for all the tips and advice. Sorry for the errors on my part regarding accessors/setters and for cleaning up the code, I greatly appreciate you taking the time to do all that and will do my best to remember all this for any subsequent PRs! 

One thing I did want to point out is that with ggplot2, the `aes_string` function has been soft-deprecated, so in lieu of that I had shifted over to a tidyeval form of accepting arguments programmatically, specifically using the `rlang::sym` and `rlang::!!` (bang-bang) functions to pass arguments. Just wanted to point it out just in case, but for consistency I agree that using `aes_string` is the better solution for now.

mojaveazure:
I'm aware of the `aes_string` deprecation, but am hesitant to use `rlang` more than necessary within Seurat. We've had [issues](https://github.com/satijalab/seurat/issues?utf8=%E2%9C%93&q=is%3Aissue+dll+) with Seurat trying to load too many DLLs, which stems from our massive dependency list and the majority of those dependencies relying on compiled code. As such, as much as possible, we try to remove dependencies that have a compiled component to them so that we avoid this issue.

If you don't have any other objections or questions, I'll merge this in to Seurat.

robertamezquita:
No objections at all, this is all great. Thanks!

mojaveazure:
Wonderful, thanks for the new functionality!

XavTek:
Hi,
I am trying to use this function.
What is the best way to do it?
Thanks

PR #908, opened 2018-11-01
Release/3.0 adding ReadAlevin

k3yavi:
>Hi Seurat team,
First of all thanks for the awesome tool, it has helped us and our user base a lot in analyzing the scRNA-seq data.
We have recently developed a dscRNA-seq quantification pipeline with name Alevin (preprint here), and we frequently encounter feature requests from our user base regarding the preprocessing of the reads for subsampling/batch correction like COMBINE-lab/salmon#305. We generally recommend users to go through this tutorial to parse the Alevin output and perform bias correction through Seurat package downstream of Alevin. In lieu of that to simplify the clustering pipeline by one more step we'd love to make the following pull request to parse the Alevin output and would really appreciate, if possible, to optimize the committed parsing function and merge to master in your next release.
Alevin dumps both csv and a more efficient binary format. I'd appreciate if you see something horribly wrong in the committed parsing code and would appreciate your feedback.

Second time the charm!

### DIFF
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -357,7 +357,7 @@ LogNormalize <- function(data, scale.factor = 1e4, verbose = TRUE) {
 #' Load in data from Alevin pipeline
 #'
 #' Enables easy loading of csv format matrix provided by Alevin
-# ran with `--dumpCsvCounts` flags.
+#' ran with `--dumpCsvCounts` flags.
 #'
 #' @param base.path Directory containing the alevin/quant_mat*
 #' files provided by Alevin.
@@ -365,6 +365,8 @@ LogNormalize <- function(data, scale.factor = 1e4, verbose = TRUE) {
 #' @return Returns a matrix with rows and columns labeled
 #'
 #' @export
+#'
+#' @author Avi Srivastava
 #'
 #' @examples
 #' \dontrun{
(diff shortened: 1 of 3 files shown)


R/preprocessing.R
```diff
@@ -354,6 +354,113 @@ LogNormalize <- function(data, scale.factor = 1e4, verbose = TRUE) {
...
+#'
+#' @export
+#'
+#' @author Avi Srivastava
+#'
+#' @examples
+#' \dontrun{
+#' data_dir <- 'path/to/output/directory'
+#' list.files(data_dir) # Should show alevin/quants_mat* files
+#' expression_matrix <- ReadAlevin(base.path = data_dir)
```
mojaveazure: For `ReadAlevinCsv`, the example should showcase `ReadAlevinCsv`

R/preprocessing.R
```diff
@@ -354,6 +354,113 @@ LogNormalize <- function(data, scale.factor = 1e4, verbose = TRUE) {
   return(norm.data)
 }
 
+#' Load in data from Alevin pipeline
+#'
+#' Enables easy loading of csv format matrix provided by Alevin
+#' ran with `--dumpCsvCounts` flags.
+#'
+#' @param data.dir Directory containing the alevin/quant_mat*
```
mojaveazure: Documented parameters don't match function definition on line 379

R/preprocessing.R
```diff
@@ -354,6 +354,113 @@ LogNormalize <- function(data, scale.factor = 1e4, verbose = TRUE) {
...
+  rownames(x = matrix) <- gene.names
+  matrix[is.na(x = matrix)] <- 0
+  return(matrix)
+}
+
+#' Load in data from Alevin pipeline
+#'
+#' Enables easy loading of binary format matrix provided by Alevin
+#'
+#' @param data.dir Directory containing the alevin/quant_mat*
```
mojaveazure: Documented parameters don't match function definition on line 426

k3yavi:
Hi @mojaveazure , apologies for the slow response.
Thanks for reviewing the pull request, I have updated the requested syntactical changes.
re: combining the function: The initial reasoning for keeping the two functions separate was, as there were two functions for `Read10x`, separate for text and compressed format; I was trying to be consistent but a more technical reason was that the binary matrix parser of Alevin might not be portable across different `endianness` (currently fixed as little and works well for x86 architecture) and might need some patch later sometime in future, while CSV reader is not expected to change. If you think keeping the two functions into one is a better idea I would be happy to combine them but just wanted to give heads up of our thoughts.

mojaveazure:
:+1: That reason is good enough for me. Thanks for the new functionality!

Issue #486, opened 2018-05-22
umap installation problem 


hmassalha:
Hi,
I am trying to install the umap feature based on your instructions and lmcinnes/umap without any success. I installed the 'umap-learn' using _pip_, however, Seurat still not recognizing the package.
Any help/suggestion would be very appreciated. (I am using the last version of Seurat)
Thanks, HM

mojaveazure:
What was the exact command you used to install UMAP?

In Python, what happens when you run
```python
>>> import umap
```

In R, what's the output of
```R
> reticulate::py_config()
```

In the shell, what's the output of
```bash
echo $PYTHONPATH
```

hmassalha:
I installed anaconda and then run the pip command. I assume something went bad with the installation of the package. Sorry in advance if I did something stupid. I have no experience in Python. Thanks. HM 

The output of `> reticulate::py_config()` in R is: Error: Specified conda binary 'C:\Users\Hassan\AppData\Local\CONTIN~1\ANACON~1/Scripts/conda.exe' does not exist.

The output of `>>> import umap` in Python is: 
`Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2018.1.3\helpers\pydev\_pydev_bundle\pydev_import_hook.py", line 19, in do_import
    module = self._system_import(name, *args, **kwargs)
ImportError: No module named umap`

No output for `echo $PYTHONPATH` in shell.

mojaveazure:
I don't think you can use Pip with Anaconda as Anaconda likes to do things its own way and become the dominant Python installation. I recommend getting rid of Anaconda, but if you need it for some reason, then use `conda` (Anaconda's package manager) to install UMAP (assuming UMAP is available on `conda`) or installing it by hand using Anaconda Python.

sansense:
I think I have got a solution to this problem. Anaconda uses its own _pip_, and the default installation by pip (on Anaconda) goes to _~/.local/anaconda3/lib/python3.6/site-packages_  on my machine. Reticulate is not seeing modules at this path

Instead, I installed UMAP at the default pip (without anaconda) path (~/.local/lib/python3.6/site-packages ), which solved the problem

`pip install -t ~/.local/lib/python3.6/site-packages umap-learn`

PR #949, opened 2018-11-16
ExportToCellbrowser function

mxposed:
This function exports Seurat object to UCSC Cellbrowser compatible format. Optionally, it runs cellbrowser itself to build static html files and open them in the browser.

cc @maximilianh


R/utilities.R
```diff
@@ -1148,3 +1148,185 @@ ToNumeric <- function(x){
...
+#' @param embeddings vector of embedding names to export
+#' @param markers.file path to file with marker genes
+#' @param cluster.field name of the metadata field containing cell cluster
+#' @param cb.dir in which dir to create UCSC cell browser content root
+#' @param port \emph{experimental} on which port to run UCSC cell browser after export
+#'
+#' @export
+#'
+#' @examples
+#' ExportToCellbrowser(object = pbmc_small, dataset.name = "PBMC", dir = "out")
```
andrewwbutler: For examples that require python packages, we generally recommend wrapping them in `/dontrun{}` so that the CI/CRAN checks will pass.

R/utilities.R
```diff
@@ -1148,3 +1148,185 @@ ToNumeric <- function(x){
...
+  dataset.name = NULL,
+  meta.fields = NULL,
+  meta.fields.names = NULL,
+  embeddings = c("tsne"),
+  markers.file = NULL,
+  cluster.field = NULL,
+  port = NULL,
+  cb.dir = NULL
+) {
+  if (is.null(dir)) {
```
andrewwbutler: For arguments that are required (e.g. `dir` and `dataset.name`), we recommend leaving the default empty (like you have for `object`) so that it's clear to users that they need to supply them. Then, you don't have to explicitly check if they're `NULL` later on in the code.

R/utilities.R
```diff
@@ -1148,3 +1148,185 @@ ToNumeric <- function(x){
...
+    }
+  }
+  if (!is.null(port) && is.null(cb.dir)) {
+    stop("cb.dir parameter is needed when port is set")
+  }
+  if (!dir.exists(dir)) {
+    dir.create(dir)
+  }
+  if (!dir.exists(dir)) {
+    stop(sprintf("Output director %s cannot be created or is a file", dir))
```
andrewwbutler: Functions like `stop` will automatically concatenate strings. We would recommend writing this as `stop("Output directory ", dir, " cannot be created or is a file")`

R/utilities.R
```diff
@@ -1148,3 +1148,185 @@ ToNumeric <- function(x){
...
+    cluster.field,
+    enum.string,
+    markers.string,
+    coords.string
+  )
+  cat(config, file=file.path(dir, "cellbrowser.conf"))
+  message("Prepared cellbrowser directory ", dir)
+
+  if (!is.null(cb.dir) && require(reticulate)) {
+    library(reticulate)
```
andrewwbutler: `library` shouldn't be used within a function. Instead, import only the required functions in the roxygen documentation. E.g `@importFrom reticulate py_module_available`

R/utilities.R
```diff
@@ -1148,3 +1148,185 @@ ToNumeric <- function(x){
...
+    if (!py_module_available("cellbrowser")) {
+      py_install("cellbrowser") # see https://rstudio.github.io/reticulate/articles/python_packages.html
```
andrewwbutler: If the module isn't available, we would prefer that the function stops with a message rather than trying to install with `py_install`. Users tend to have varied preferred installation methods and would prefer that they handle the install themselves.

R/utilities.R
```diff
@@ -1148,3 +1148,185 @@ ToNumeric <- function(x){
...
+    }
+  }
+}
+
+#' Stop Cellbrowser web server
+#'
+#' @export
+#'
+#' @examples
+#' StopCellbrowser()
```
andrewwbutler: Wrapping in `\dontrun{}` is recommended.

R/utilities.R
```diff
@@ -1148,3 +1148,185 @@ ToNumeric <- function(x){
...
+#'
+StopCellbrowser <- function() {
+    library(reticulate)
```
andrewwbutler: Similarly, please use the `importFrom` tag within the roxygen documentation instead of `library`

mxposed:
Thank you for the comments! I addressed the issues in the latest commit. Please let me know if I should change anything else

R/utilities.R
maximilianh: Thanks for the detailed review!

I agree with you on this on, but there is one problem that is solved by installing the package through reticulate: users may have multiple pythons (2, 3, the one that came with the OS, one they built themselves, one through conda, etc) and they may not be sure which one is the python that is used by R. So by installing the package through reticulate, we are sure that we're installing into the correct Python. I agree that it shouldn't do the install automatically, but maybe we can add another parameter do.install=FALSE that the user can set to get the installation done? If the package is not there, we can stop with a message that tells the user to use pip or alternative set do.install=TRUE? Would you be OK with that?

R/utilities.R
```diff
@@ -1299,20 +1298,23 @@ coords=%s'
...
   message("Prepared cellbrowser directory ", dir)
 
-  if (!is.null(cb.dir) && require(reticulate)) {
-    library(reticulate)
+  if (!is.null(cb.dir)) {
     if (!py_module_available("cellbrowser")) {
-      py_install("cellbrowser") # see https://rstudio.github.io/reticulate/articles/python_packages.html
+      stop(
+        "Python package `cellbrowser` is required to prepare and run ",
+        "Cellbrowser. Please, install it with `pip install cellbrowser`."
```
maximilianh: This may work for many people. It would be great if this was always correct, but unfortunately installing modules in python can be quite complicated, depending on the OS. Reticulate uses the first python on the path, which may be python3. If the OS has a pip, and pip3 and a pip2, then typing "pip" may install cellbrowser, but not into the python that reticulate uses.  So I would suggest adding another option, e.g. do.install that takes care of the install by running py_install("cellbrowser"). In addition, we could remind the user that they can always do require(reticulate) + use_python(pathname) to use a particular Python version that they like or a virtual env in that python with use_virtualenv("myenv") or a conda environment with use_condaenv(name). I know this is getting quite complicated, but it is likely that we have users here that are not Python fans or alternatively may have Python fans here that don't understand R well, and few people know reticulate, so I think it's good to make this part as easy as possible for them. If they stumble here, they may give up entirely...

mojaveazure: Hi Maximilian,

Reticulate facilitates using alternative Python installations via [`use_python`](https://www.rdocumentation.org/packages/reticulate/versions/1.10/topics/use_python), so if a user uses an alternative Python they can set reticulate to use that particular version of Python. In our experience, reticulate can switch between CPython 2.X and 3.X with relative ease, so we have little concern for pip installations. We have also noticed that reticulate has issues with Anaconda Python, so we encourage our users to use [CPython over Anaconda as well as use the system's version of Pip over conda](https://github.com/satijalab/seurat/issues/486#issuecomment-391457067) when using a Seurat function that calls out to Python. Using CPython and Pip installed from the system package manager or from [python.org](https://www.python.org) to install modules generally allow Seurat and reticulate to work without any extra effort on the user's end.

Moreover, CRAN generally frowns upon a package modifying the system outside of using temporary files. As per [CRAN's policies](https://cran.r-project.org/web/packages/policies.html):

> The code and examples provided in a package should never do anything which might be regarded as malicious or anti-social. The following are illustrative examples from past experience.
> - [...]
> - Packages should not write in the user’s home filespace (including clipboards), nor anywhere else on the file system apart from the R session’s temporary directory (or during installation in the location pointed to by TMPDIR: and such usage should be cleaned up). Installing into the system’s R installation (e.g., scripts to its bin directory) is not allowed.

As such, we are hesitant to include code that messes with a user's system, such as installing Python packages, and will instead help our users use CPython and Pip to install Python modules used by Seurat. Users savvy enough to use virtual or conda environments should be able to look through reticulate's documentation and figure out how to configure reticulate to work with their Python setup.

maximilianh: This makes sense, I didn't know this CRAN rule. I'll remove the py_install call. thanks!

R/utilities.R
```diff
@@ -509,6 +509,191 @@ ExpMean <- function(x) {
...
+#' @param vars vector of metadata fields to export
+#' @param vars.names list that defines metadata field names after the export.
+#' Should map metadata column name to export name
+#' @param reductions vector of reduction names to export
+#' @param markers.file path to file with marker genes
+#' @param cluster.field name of the metadata field containing cell cluster
+#' @param cb.dir in which dir to create UCSC cell browser content root
+#' @param port \emph{experimental} on which port to run UCSC cell browser after export
+#'
+#' @return ...
```
mojaveazure: It seems as though this function creates some kind of output (not in R, but on the filesystem). Could you document this somewhere?

R/utilities.R
```diff
@@ -509,6 +509,191 @@ ExpMean <- function(x) {
...
+#' @examples
+#' \dontrun{
+#' ExportToCellbrowser(object = pbmc_small, dataset.name = "PBMC", dir = "out")
+#' }
+#'
+ExportToCellbrowser <- function(
+  object,
+  dir,
+  dataset.name,
+  vars = NULL,
```
mojaveazure: I'm wondering if `vars` and `vars.names` could be replaced by an optionally-named vector using the `...`, similar to how `RenameIdents` works (see https://github.com/satijalab/seurat/blob/release/3.0/R/objects.R#L2316:2341). As currently done, either all variables would have to be remapped or none would be remapped (or you'd have to order the variables so that the remapped ones wold come first).

With my proposed change, the `...` would be the metadata fields to pull, and the names would be the mapping. If no names are provided, the `names` attribute of `...` would be `NULL`. If only some are named, you could easily test this using `nchar`. This could be done like so:

```R
vars <- c(...)
names(x = vars) <- names(x = vars) %||% vars
names(x = vars) <- sapply(
  X = 1:length(x = vars),
  FUN = function(i) {
    return(ifelse(test = nchar(x = names(x = vars)[i]) > 0, yes = names(x = vars[i]), no = vars[i]))
  }
)
```

What do you think?

R/utilities.R
```diff
@@ -509,6 +509,191 @@ ExpMean <- function(x) {
...
+#'
+#' @examples
+#' \dontrun{
+#' ExportToCellbrowser(object = pbmc_small, dataset.name = "PBMC", dir = "out")
+#' }
+#'
+ExportToCellbrowser <- function(
+  object,
+  dir,
+  dataset.name,
```
mojaveazure: I'm wondering if a sensible default would be the project name of the Seurat object? This could be set in the function definition with `dataset.name = slot(object = object, name = 'project')` or as null in the function definition with `dataset.name <- dataset.name %||% slot(object = object, name = 'project')` in the function body.

R/utilities.R
```diff
@@ -509,6 +509,191 @@ ExpMean <- function(x) {
...
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' ExportToCellbrowser(object = pbmc_small, dataset.name = "PBMC", dir = "out")
+#' }
+#'
+ExportToCellbrowser <- function(
+  object,
+  dir,
```
mojaveazure, line 556: I'm wondering if `tempdir` could be a sensible default?

R/utilities.R
```diff
@@ -509,6 +509,191 @@ ExpMean <- function(x) {
...
+  dataset.name,
+  vars = NULL,
+  vars.names = NULL,
+  reductions = "tsne",
+  markers.file = NULL,
+  cluster.field = NULL,
+  port = NULL,
+  cb.dir = NULL
```
mojaveazure: I'm not sure why this needs to be separate from `dir`, but I don't know cell browser so it could very well need to be different. Could you ensure that the docs are clear as to why these need to be different?

R/utilities.R
```diff
@@ -509,6 +509,191 @@ ExpMean <- function(x) {
...
+    )
+    conf <- sprintf(
+      '{"file": "%s.coords.tsv", "shortLabel": "Seurat %1$s"}',
+      reduction
+    )
+    embeddings.conf <- c(embeddings.conf, conf)
+  }
+  # Export metadata
+  df <- data.frame(row.names = rownames(x = object[[]]))
+  for (field in vars) {
```
mojaveazure: Could this for-loop be replaced by a call to `FetchData`, especially as passing `'ident'` will cause `FetchData` to pull the active identity classes.

R/utilities.R
mxposed: Thank you for this suggestion. Currently, var names are exported to the header row of `metadata.tsv` table, where they can contain almost any character, and then they are shown in the browser website as properties of cells.
Do I get it right that with this proposed strategy we would be limited to character strings that are valid in R to be names of the arguments?

mojaveazure: Not necessarily, it depends on if you quote named arguments or not. R variable names have to follow certain order, and `data.frame` does weird things with names, but once you fetch the the metadata variables you can replace the column names with the names of `vars` and R won't complain. I've written a dummy function that takes the `...` I've proposed and creates a data frame with the mapped names as column names.

```R
f <- function(...) {
  vars <- c(...)
  names(x = vars) <- names(x = vars) %||% vars
  names(x = vars) <- sapply(
    X = 1:length(x = vars),
    FUN = function(i) {
      return(ifelse(
        test = nchar(x = names(x = vars)[i]) > 0,
        yes = names(x = vars)[i],
        no = vars[i]
      ))
    }
  )
  df <- data.frame(as.list(x = vars)) # Replace this with actually fetching the metadata variables
  colnames(x = df) <- names(x = vars) # This is where you set the column names of the final data frame to have any character string
  return(df)
}
```

If you play with it, you'll see that R won't object to random character strings.

```R
> f(1, '1_a' = 3, c.a = 'potato', "%$#" = 5)
  1 1_a    c.a %$#
1 1   3 potato   5
```

Unquoted argument names have to follow R's argument naming conventions. Quoted argument names, however, can contain any valid character string (I haven't found a limit yet), including escape characters (such newlines) that cell browser recognizes.

maximilianh:
Sorry for a question from a complete R beginner: what would the function
call look like if we support the "..." notation?

mxposed:
> Sorry for a question from a complete R beginner: what would the function call look like if we support the "..." notation?

The function call would be like `ExportToCellbrowser(obj, dir = "dir", dataset.name = "PBMC_small", "Cluster" = "ident", "Number of genes" = "nFeature_RNA")`

I think this is cool. Paul, thank you for the snippets and explanation

maximilianh:
Oh I see, yes this indeed looks cool. Hope it's not difficult to implement,
for now, we're mostly trying to get a first version going.

mxposed:
@maximilianh @mojaveazure Which way do you think is better? `"nFeature_RNA" = "Number of genes"` or `"Number of genes" = "nFeature_RNA"`?

maximilianh:
I guess I'd take the first option, but I don't think it matters. Feel free
to decide or use sample(1:2, 1) to get an answer. :-)

mojaveazure:
@maximilianh @mxposed Either. `RenameIdents`, the function in Seurat that works in this manner uses the former style (`old = new`), but that's because and old/new remapping is required. For this, though, the remapping isn't required so the latter makes more sense to me. However, it'd then be backwards from `RenameIdents`, and one thing we've strived for is consistency in Seurat.

The examples I've provided all operate under the assumption of the latter (`new = old`). If you use the first option, then you'd fetch the metadata using `names(x = vars)` and reassign column names as `colnames(x = df) <- vars`; everything else would remain the same.

maximilianh:
Thanks to you guys for the detailed code review!

PR #1051, opened 2019-01-12
Implement the Leiden clustering algorithm in FindClusters

TomKellyGenetics:
As discussed via twitter.

Calls the python implementation of the Leiden algorithm with reticulate and returns the resulting partition membership to Suerat idents.

The algorithm requires the "leidenalg" python module to be installed via pip or conda in a similar manner to the RunUMAP. See the Python implementation for details on functionality. Currently, this runs the Leiden algorithm with default settings but passing arguments to the Python function would be possible.

https://github.com/vtraag/leidenalg

Merges the functionality of the Leiden R package:

https://github.com/TomKellyGenetics/leiden

Note that this function passes iGraph objects to disk to pass the SNN data to the Python format. This may violate the requirements for a CRAN package. Assistance to resolve this issue would be appreciated.

PR #1060, opened 2019-01-14
Implement the Leiden clustering algorithm in FindClusters

TomKellyGenetics:
Implements Leiden as discussed in #1051


R/clustering.R
```diff
@@ -476,3 +480,42 @@ RunModularityClustering <- function(
...
+  if (!py_module_available(module = 'leidenalg')) {
+    stop("Cannot find Leiden algorithm, please install through pip (e.g. pip install leidenalg).")
+  }
+  
+  #import python modules with reticulate
+  leidenalg <- reticulate::import("leidenalg")
+  ig <- reticulate::import("igraph")
+  
+  #export graph structure to call from Python
+  snn_graph <- object@snn
```
mojaveazure: Seurat v3 no longer has an `snn` slot; instead, we store neighbor graphs in the `graphs` slot (see our [Developer's Guide](https://github.com/satijalab/seurat/wiki/Seurat) for more details). Please update to ensure that you're pulling the correct slot information.

R/clustering.R
```diff
@@ -476,3 +480,42 @@ RunModularityClustering <- function(
...
+
+# Set up Leiden clustering algorithm
+# @name Leiden State Matrix
+# @rdname leiden
+#
+# @title Run Leiden clustering algorithm
+#
+# @description Implements the Leiden clustering algorithm in R using reticulate to run the Python version. Requires the python "leidenalg" and "igraph" modules to be installed. Returns a vector of partition indices.
+#
+# @param object A seurat class object.
```
mojaveazure: It looks like you're calling `RunLeiden` from `FindClusters.default` instead of `FindClusters.Seurat`; `FindClusters.default` never touches a `Seurat` object; it usually works on matrices (the SNN graph) or data.frames instead. As such, it's getting passed the matrix already, not a `Seurat` object. This is similar to how `RunModularityClustering` works. Can you update your documentation and function to work on the matrix instead of the `Seurat` object?

R/clustering.R
```diff
@@ -476,3 +480,42 @@ RunModularityClustering <- function(
...
+# @importFrom igraph graph_from_adjacency_matrix write.graph
+# @export
+
+RunLeiden <- function(object, ...){
+  if (!py_module_available(module = 'leidenalg')) {
+    stop("Cannot find Leiden algorithm, please install through pip (e.g. pip install leidenalg).")
+  }
+  
+  #import python modules with reticulate
+  leidenalg <- reticulate::import("leidenalg")
```
mojaveazure: The `package::function` construct should be replaced with `#' @importFrom` comments in the documentation and simple function calls in the code.

R/clustering.R
```diff
@@ -476,3 +480,42 @@ RunModularityClustering <- function(
...
+#
+# @title Run Leiden clustering algorithm
+#
+# @description Implements the Leiden clustering algorithm in R using reticulate to run the Python version. Requires the python "leidenalg" and "igraph" modules to be installed. Returns a vector of partition indices.
+#
+# @param object A seurat class object.
+# @param ... Parameters to pass to the Python leidenalg function.
+#
+# @keywords graph network igraph mvtnorm simulation
+# @import reticulate
```
mojaveazure: `@import` and `@importFrom` Roxygen comments should always have the `#'` prefix so that they get run when `devtools::document` is run. This updates our NAMESPACE file with our dependencies. Please ensure that these comments are seen by Roxygen.

Also, could you make sure that you update the DESCRIPTION file with any dependencies you've added but aren't listed? Finally, we're trying to cut down on blanket `@import` statements; could you ensure that you're using `@importFrom` and list out which parts from each package we're importing and using?

R/clustering.R
```diff
@@ -476,3 +480,42 @@ RunModularityClustering <- function(
...
+  
+  #import python modules with reticulate
+  leidenalg <- reticulate::import("leidenalg")
+  ig <- reticulate::import("igraph")
+  
+  #export graph structure to call from Python
+  snn_graph <- object@snn
+  
+  #passes iGraph object to working directory to pass to python
+  igraph::write.graph(snn_graph, file = ".graph.edges", format = "graphml")
```
mojaveazure: Please use `tempfile` to make a temporary file in the system's temporary directory. Writing in locations that aren't a system's temporary directory violates [CRAN's policies](https://cran.r-project.org/web/packages/policies.html)

> - Packages should not write in the user’s home filespace (including clipboards), nor anywhere else on the file system apart from the R session’s temporary directory (or during installation in the location pointed to by TMPDIR: and such usage should be cleaned up). Installing into the system’s R installation (e.g., scripts to its bin directory) is not allowed.

R/clustering.R
```diff
@@ -50,18 +50,22 @@ FindClusters.default <- function(
...
+            n.start = n.start,
+            n.iter = n.iter,
+            random.seed = random.seed,
+            print.output = verbose,
+            temp.file.location = temp.file.location,
+            edge.file.name = edge.file.name
+          )
+        } else if (algorithm== 4 || tolower(algorithm) == "leiden"){
+          ids <- RunLeiden(object)
+        }
```
mojaveazure, line 87: Have an `else` block to throw an error if the user specifies an unrecognized algorithm. This will be a lot nicer than R's default error messages.

R/clustering.R
TomKellyGenetics: A temporary file is no longer necessary as discussed: https://github.com/TomKellyGenetics/leiden/issues/1

R/clustering.R
TomKellyGenetics: This has been implemented

R/clustering.R
TomKellyGenetics: I think this is not necessary if `FindClusters.default` runs on an SNN matrix.

R/clustering.R
TomKellyGenetics: This is good to know. This is more consistent with the original reticulate based package: https://github.com/TomKellyGenetics/leiden

R/clustering.R
TomKellyGenetics: These have been removed

R/clustering.R
TomKellyGenetics: I think I've done this correctly but I'm not sure how roxygen handles internal functions. I'm unsure if importing only these specific functions from reticulate are sufficient for the Python interface to work. This may need to be tested more extensively.

TomKellyGenetics:
Hi Paul,

Thanks for your pointers. I hope this update addresses the problems you've raised. I think this improves the function.

One remaining problem (raised by one of my colleagues) is how to pass arguments to the Python leidenalg `ModularityVertexPartition` function. We are still working on this as discussed here: https://github.com/TomKellyGenetics/leiden/issues/2

R/clustering.R
mojaveazure: `@import` makes the entire namespace available for the function being documented, `@importFrom` makes only specific functions available. In Python terms, this is akin to `from package import *` versus `from package import function1, function2`; this doesn't mean that the functions imported don't see everything from their original package, just that we don't. Therefore, just importing the functions you call from reticulate doesn't mean that the Python objects you create in R don't have access to the full reticulate package. Those objects still have access to all necessary code to enable them to function.

We prefer the use of `@importFrom` for declaring dependencies for two reasons: being explicit and being lazy. Using `@importFrom` over `@import` requires that we declare each and every function from a dependency package, which means we can keep track of where our code comes from (`base` and Seurat-defined functions aren't declared, everything else is) and which version of each function we're running. It also means we can reduce namespace clash within Seurat, as several packages name functions the same (eg. `Matrix::colSums` and `base::colSums`).

We could require that everyone uses the `package::function` construct (analogous to Python's `module.function` construct), but then we lose Roxygen's automatic dependency declaration in the NAMESPACE file, and it's very easy to turn a NAMESPACE into a list of imports for the DESCRIPTION (all stuff required for CRAN). Since we're too lazy to comb through Seurat and keep track of every unique dependency, we let Roxygen do it for us with the `@importFrom` tags.

Hopefully this makes sense and assures your concerns about use of this construct to declare dependencies.

mojaveazure:
Hi Tom,

Thanks for addressing those changes. I've made a couple of changes regarding style, documentation, and using `switch` instead of `if/else` for the Leiden method used (it's faster). Could you take one last look over to see if I broke anything that our test didn't catch? Also, does the Python `leidenalg` module install the `igraph` module as well? If not, could you be sure to check for it using `py_module_available` the same way we check for `leidenalg`?

As for that issue with passing extra arguments to `ModularityVertexPartition`, I can't say I know how to overcome this issue. I'll leave this PR open until I hear from you that you're ready for me to merge it in.

TomKellyGenetics:
Hi Paul,

Thanks for your feedback and changes to the imports. I see why that is recommended but I'm not 100% if reticulate needs more functions to handle passing objects to python (hopefully these are imported with the `import` and `r_to_py` functions).

I've checked that it will run on seurat objects including `pbmc_small`. This version should pass arguments to the python functions as well. As I understand it, `NULL` in R will get passed to `None` in Python as default values. I have not tested whether non-default values (e.g., vectors) can be passed here. This won't pass new arguments to the python version but the list of parameters is not exhaustive (although they are not documented well for the R function yet).

I've checked the install of `leidenalg` in a fresh conda environment. It installs the `igraph` python library as required for check with `py_module_available` for `leidenalg` should be sufficient. The `setup.py` for this module includes compiling the C and python versions of `igraph`. The R version of `igraph` is no longer required (as discussed here: https://github.com/TomKellyGenetics/leiden/issues/1).

mojaveazure:
Thanks!

TomKellyGenetics:
No problem, happy to help.

Just an added note, I've made some additional changes to the separate "leiden" package and have submitted this to CRAN. This function works on an adjacency or SNN matrix in much the same manner as `RunLeiden` so it could be called as a dependency instead (i.e., import function `leiden` from package `leiden`). At the moment, the main difference is better handling of weighted matrices and some minor tweaks for performance. I don't think is pressing to update the Seurat version but it calling the function from the package may be an easier way to keep up to date. Alternatively, I can send another PR if any major changes come up. 

https://github.com/TomKellyGenetics/leiden

PR #1127, opened 2019-02-08
Update fftRtsne to be compatible with FIt-SNE v1.1.0

linqiaozhi:
FIt-SNE [v1.1.0](https://github.com/KlugerLab/FIt-SNE/releases/tag/v1.1.0) was released today, which is incompatible with Seurat 3.0. I made minor changes to the Seurat v3.0 built-in FIt-SNE wrapper (fftRtsne) to make it compatible.


andrewwbutler:
Hi George,

Thanks for updating the wrapper and congrats on the publication as well! Looking at the changes, it seems they're fairly minimal (at least on the wrapper end of things). For backwards compatibility purposes, could we have it check for the version of FIt-SNE and then run the appropriate code based on what it finds? We could also display a message encouraging people to update their FIt-SNE install if it finds an older version.

linqiaozhi:
Hey @andrewwbutler, thanks for the suggestion. I went ahead and added backwards compatibility for v1.0.0.  The main functionality change between v1.0 and v1.1 is that the latter supports adjusting the "degree of freedom" of the t-distribution modeling low-dimensional distances. With this new commit, if the user attempts to change that parameter but they are not using v1.1, then an error will be shown.

linqiaozhi:
Hm...looks like one of the checks failed, but it's not clear why. I think this is the error:

> Error in contrib.url(repos, "source") : 
>   trying to use CRAN without setting a mirror
> Calls: install.packages -> contrib.url
> Execution halted
> Command exited with code 1
> 7z a failure.zip *.Rcheck\*

How does that relate to my most recent (very minor) changes? They passed in my original commit...

andrewwbutler:
Thanks for adding that! I think the build fail is probably something on appveyor's side of things but I'll look into it.

PR #1236, opened 2019-03-15
Rename `percent.mito` to `fraction.mito`

### LINKED ISSUE #1235: percent.mito is not a percentage
From what I gather, the value returned by `'percent.mito'` is not actually a percentage. In the [tutorial](https://satijalab.org/seurat/pbmc3k_tutorial.html), you filter cells which have less than 0.05 "percent" mitochondrial expression - I think you actually mean 5% which implies that the percentage ranges from 0 to 1. This is confusing and misleading.

mschilli87:
The data range is 0 - 1, not 0 - 100.
Thus, when specifying `percent.mito=1` to ask for 1 %, in fact the interpretation is in 100 %.

Instead of changing the behaviour, the name is adjusted to break as little existing code (and in a controlled manner only, resulting in an error message instead of silently changing the results).

All changes in this commit were introduced by the following command:

```sh
find -type f -not -path ./.git/\* \
  -exec sed -i 's/percent\.mito/fraction.mito/g' "{}" \;
```

This commit fixes #1235.

### DIFF
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -872,7 +872,7 @@ Read10X_h5 <- function(filename, use.names = TRUE) {
 #' is set to NULL; default is 1.3
 #' @param return.dev.residuals Place deviance residuals instead of Pearson residuals in scale.data slot; default is FALSE
 #' @param vars.to.regress Variables to regress out in a second non-regularized linear
-#' regression. For example, percent.mito. Default is NULL
+#' regression. For example, fraction.mito. Default is NULL
 #' @param do.scale Whether to scale residuals to have unit variance; default is FALSE
 #' @param do.center Whether to center residuals to have mean zero; default is TRUE
 #' @param clip.range Range to clip the residuals to; default is \code{c(-sqrt(n/30), sqrt(n/30))},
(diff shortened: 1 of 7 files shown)


TomKellyGenetics:
Won't this change break a lot of existing scripts and pipelines calling Seurat? Filtering by mitochondrial reads is a routine step in many analyses.

I'm not sure this pull request does what is described. It does not handle the case where "percent.mito" is still entered (as legacy code). It is advised by R developers to deprecate existing functionality rather than removing it entirely. As far as I can see, these changes, only affect the documentation and not the code. The inputs expected are not changed and warnings are not explicitly given.

mschilli87:
@TomKellyGenetics: Thx for the feedback. If you have a look at https://github.com/satijalab/seurat/issues/1235 you'll find I agree to a large extend.

TomKellyGenetics:
Thanks for clarifying. If this only affects the documentation and vignette, then I think it is not a problem to change it. This appears to be an example of variable that could be used to filter out cells and any other could be used in it's place. In that case, the example could be changed without breaking any existing code (and appears to be being addressed already). Considering your discussions there, I agree with you: the variable name should be changed to `fraction.mito` rather than altering the definition of `percent.mito` to avoid confusion with earlier versions with the same name.

> @andrewwbutler
> I agree that changing the variable name to `fraction.mito` in the vignette is probably the best bet here. However, this isn’t really a Seurat level API change as that’s just a variable name we define (albeit imprecisely) in a vignette. Users should still be able to create `percent.mito` and store that in metadata if they want so I don’t think a warning is necessary here. We’ll make sure this is updated for all v3 versions of the vignettes.

andrewwbutler:
Thanks for the input. We've added a function called `PercentageFeatureSet` in v3 that now handles that calculation (and actually returns a percentage). Hopefully this clears up any confusion going forward.

PR #1262, opened 2019-03-21
Correct the way RunTSNE() deal with distance objects

chenyenchung:
Hi there, 
    This edit is in response to an issue where when a dist object is provided to `RunTSNE()` in `distance.matrix`, an error will be triggered. A reproducible example (with v3.0) is:

```
library(Seurat)
dummyexpMat <- t(iris[1:4])
colnames(dummyexpMat) <- paste0("cell", seq(ncol(dummyexpMat)))
row.names(dummyexpMat) <- paste0("gene", seq(nrow(dummyexpMat)))

obj <- CreateSeuratObject(counts = dummyexpMat)
distMat <- dist(t(dummyexpMat))

obj <- RunTSNE(obj, distance.matrix = distMat, perplexity = 1, 
               check_duplicates = FALSE)

### Triggers this error message:
#> Error in Rtsne.default(X = object, dims = dim.embed, ...) : 
#>   formal argument "is_distance" matched by multiple actual arguments
```

It seems to be a result of providing `is_distance = TRUE` twice. The changes in this PR assumes the follows and stop `RunTSNE()` from passing `is_distance`:
1. `RunTSNE.dist` eventually converts a `dist` object to a matrix
2. `Rtsne::Rtsne` has a method for class `dist` which sets `is_dist` automatically

Thank you very much for your time.

Best,
Yen-Chung Chen

--
Edit:
Thanks a lot, Paul! A unit test is added to make sure it is working consistently.


mojaveazure:
Thanks for catching and fixing this bug!

chenyenchung:
Thank you for the help and guidance :)

Issue #1021, opened 2018-12-20
Convert AnnData to Seurat fails with raw h5ad


BenjaminDoran:
https://github.com/satijalab/seurat/blob/37a3750f599fc4c0e1226aced130441a8b250077/R/conversion.R#L364
This line fails on some unprocessed data because the dgCmatrix does not have the correct dimensions
sparseMatrix by default assumes the dimensions of the adata.raw.x matrix by using the max values of the indices. This is not always the case as when converting unprocessed data. where the last couple of rows might be all zeros.

adding 
`
dims = c(py_to_r(from$raw$X$shape[1]), py_to_r(from$raw$X$shape[0]))
`
to the sparse matrix initialization call should fix this particular error

BenjaminDoran:
```
raw.data.matrix <- sparseMatrix(
    i = as.numeric(x = from$raw$X$indices),
    p = as.numeric(x = from$raw$X$indptr),
    x = as.numeric(x = from$raw$X$data),
    dims = c(py_to_r(from$raw$X$shape[1]), py_to_r(from$raw$X$shape[0])),
    index1 = FALSE
  )
```

satijalab:
Thanks very much! I am flagging @mojaveazure to follow up.

mojaveazure:
Reading Anndata h5ad files is present in Seurat v3 with `ReadH5AD` (a la #774 @cartal); writing h5ad files is still WIP

BenjaminDoran:
After checking the `ReadH5AD` function I found a the same issue with reading unfiltered data. The S4 DgCmatrix has the wrong dimensions reading in directly from the h5 file which prevents us from putting in the rownames and column names (var, obs) from the h5ad file.

This should be fixed by replacing 
```
  x <- if (is(object = hfile[['X']], class2 = 'H5Group')) {
    as.sparse(x = hfile[['X']])
  } else {
    hfile[['X']][, ]
  }
```
with
```
  x <- if (is(object = hfile[['X']], class2 = 'H5Group')) {
    as.sparse(x = hfile[['X']])
    x@Dim <- c(length(hfile[["var"]][]$index), length(hfile[["obs"]][]$index))
  } else {
    hfile[['X']][, ]
  }
```

smanne07:
Hi,
I am getting the following error when using  _ ReadH5AD_ in Seuratv3.

`Error in dimnamesGets(x, value) : 
  invalid dimnames given for “dgCMatrix” object
[...]

PR #1332, opened 2019-04-06
Reads H5AD files with zeroed cols

BenjaminDoran:
`as.sparse()` returns `indptr` as wrong size if some columns or rows are zeroed. setting the `Dim` slot after allows the rownames to be set and the seurat object to be created. But the dgcmatrix is not able to be computed on because 'p' is still the wrong length. 

Explicitly creating the sparse matrix with the `dims` argument set fixes issue #1021 

This may also need to be done for `WriteH5AD()`


mojaveazure:
Thanks again Benjamin!

smanne07:
Thank you

Issue #982, opened 2018-12-04
Is it possible to figure out PC variance explained in Seurat?


kysbbubbu:
Hello 

I have simple question.

Is it possible to figure out PC variance explained for each PCs (ex. PC1, PC2, ...)in Seurat?

Or should i manually figure out ?

Thank you

sansense:
You can easily get the sdev of the PCs from the SeratObject:

```
pca = SeuratObj@dr$pca
eigValues = (pca@sdev) ^2 ## 
varExplained = eigValues / sum(eigValues)
```

note that `sum(varExplained)` will be 1 (= Total variance)

zeehio:
Sorry if this message sounds blunt, but this answer is wrong.

One way to show the error is to extract the first Principal Component `RunPCA(..., npcs=1)` and follow @sansense instructions. According to those instructions you would be extracting 100% of the variance (and you are not).

`sum(eigValues)` will only be the total variance of the data when `pca@sdev` includes all the components. If you extract the first 50 components in `RunPCA(..., npcs = 50)` then `sum(eigValues)` will contain only the variance captured by those first 50 components and not the total variance of the data.

plijnzaad:
minor point: in Seurat3 the `pca@sdev` slot is called `stdev`

VitorAguiar:
@zeehio What if you use only the top 2000 variable genes for PCA? Is it still appropriate to compute total variance on the full gene expression matrix?

aandreas13:
@VitorAguiar It shouldn't matter, should it? The solution is using the 'scale.data' slot to compute the total variance. By default, scale.data is only scaling the variable genes anyway. If you are using the the 'counts' or 'data' slot, it will pose a problem indeed.

Apart from that, I realized that Seurat RunPCA computed the total variance for us already. It's in the seu_obj@reductions$pca@misc$total.variance and it agrees with the above answer.

VitorAguiar:
Hi @aandreas13, I usually don't follow the default and I scale all genes. 

Thanks for pointing out that the total variance is included in the Seurat Object. I will take a look.

aandreas13:
[...]

PR #1335, opened 2019-04-08
Let RunPCA report total variance

zeehio:
When doing a PCA it is common to report the % explained variance of the represented principal components. However, the Seurat package does not report that value.

When someone asked in an issue [Is it possible to figure out PC variance explained in Seurat?](https://github.com/satijalab/seurat/issues/982), the first answer was that it could trivially be done by summing all the `sdev^2` returned by Seurat. This is wrong, because it only accounts for the computed PCs, and not all the variance in the matrix. In other words, it is wrong because `RunPCA(..., npcs = 1)` would always return "100% of the variance explained" if we followed that approach.

This PR adds the `total_variance` to the `misc` list of the `DimReduc` object in the `RunPCA` function. I have used your custom C++ `RowVar` function (for performance) and I have added a test to ensure the validity of the results. After this PR, the total variance can be obtained with:  `slot(pca_result[["pca"]], "misc")$total_variance`

Thanks for your work and time, looking forward to the final 3.0 release!

### DIFF
--- a/R/dimensional_reduction.R
+++ b/R/dimensional_reduction.R
@@ -812,7 +812,7 @@ RunPCA.default <- function(
   if (rev.pca) {
     npcs <- min(npcs, ncol(x = object) - 1)
     pca.results <- irlba(A = object, nv = npcs, ...)
-    total_variance <- sum(RowVar(t(object)))
+    total.variance <- sum(RowVar(x = t(x = object)))
     sdev <- pca.results$d/sqrt(max(1, nrow(x = object) - 1))
     if (weight.by.var) {
       feature.loadings <- pca.results$u %*% diag(pca.results$d)
@@ -825,7 +825,7 @@ RunPCA.default <- function(
     if (approx) {
       npcs <- min(npcs, nrow(x = object) - 1)
       pca.results <- irlba(A = t(x = object), nv = npcs, ...)
-      total_variance <- sum(RowVar(object))
+      total.variance <- sum(RowVar(x = object))
       feature.loadings <- pca.results$v
       sdev <- pca.results$d/sqrt(max(1, ncol(object) - 1))
       if (weight.by.var) {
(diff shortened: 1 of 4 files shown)


andrewwbutler:
Thanks for the PR! I agree that this is useful info to store to make computing % variance explained easier for users.

Firedrops:
Thanks for the contribution!

I have tried this with the [PBMC 3k Tutorial](https://satijalab.org/seurat/articles/pbmc3k_tutorial.html), but
1. It looks like referencing the object has changed, `$total_variance` is now `$total.variance`
2. The output is an absolute value, 1733.667 in my case. How can I "translate" this into a %?

Rohit-Satyam:
Hi @Firedrops and @zeehio, I was wondering if you were able to figure out how you can translate variance into percentage? I was going through this tutorial by [HBCtraining](https://hbctraining.github.io/scRNA-seq/lessons/sc_exercises_clustering_analysis.html) and do you recommend this way of ranking PCs?? Also the `total.variance` obtained from the code below is different from what I get from `slot(SeuratObj[["pca"]], "misc")$total.variance`. 

```r
mat <- Seurat::GetAssayData(SeuratObj, assay = "RNA", slot = "scale.data")
pca <- SeuratObj[["pca"]]
total_variance <- sum(matrixStats::rowVars(mat))
eigValues = (pca@sdev)^2  ## EigenValues
varExplained = eigValues / total_variance
```

Also, do you recommend to calculate the `varExplained` as follows now when this new feature of reporting `total.variance` has been included?

```r
pca <- SeuratObj[["pca"]]
total_variance <- slot(SeuratObj[["pca"]], "misc")$total.variance
eigValues = (pca@sdev)^2  ## EigenValues
varExplained = eigValues / total_variance
```

bug1303:
I know it's an older post, but I just came across the same situation, and then found this post, so I share my findings here.

`RunPCA()`, if you don't specify `features` will be run using the variable features for the Assay.
Similarly, `ScaleData()` by default scales/centers the variable features only. 

If you run all with default parameters, they should give the same result. However, it looks like you scaled the data for all features (`ScaleData(SeuratObj, features = rownames(SeuratObj))`) and ran PCA on the variable features only. That explains the discrepancy of the two values.  

```
mat <- Seurat::GetAssayData(SeuratObj, assay = "RNA", slot = "scale.data")
total_variance <- sum(matrixStats::rowVars(mat))

mat2 <- mat[SeuratObj@assays$RNA@var.features, ]
total_variance2 <- sum(matrixStats::rowVars(mat2))
```

The latter should be the same as stored in `SeuratObj@reductions$pca@misc$total.variance`.

Issue #1388, opened 2019-04-17
Read10X failing


BSharmi:
Hello,

I am using Read10X to read from a directory with the following 3 files as needed: barcodes.tv, genes.tv, matrix.mtx

When I use Read10X I get the following error - Error in `[.data.frame`(feature.names, , gene.column) :   undefined columns selected

The matrix.mtx file looks like this - 

%%MatrixMarket matrix coordinate real general
%
17308	13313	13782382
1	337	0.697960
1	943	1.709912
1	2191	1.814328
1	2376	0.084899
1	2819	0.659472
1	3031	0.892236
1	3167	0.288861

genes.tv looks like this - 

0610005C13Rik
0610007N19Rik
0610007P14Rik
0610009B22Rik
0610009D07Rik
0610009E02Rik
0610009L18Rik
0610009O20Rik
0610010F05Rik
0610010K14Rik

barcodes.tv looks like this - 

0-1a_TACTGGTCCCCC
0-1a_AAGTCTTACACT
0-1a_CGTAGTTATTAG
0-1a_TGAATTTTTTCT
0-1a_TTTCTAATCGGA
0-1a_CAACCTTATGCA
0-1a_CGTCCTTGAGAT
0-1a_CAACAAGCTAAA
0-1a_TTCCGGCGGACC
0-1a_CTCACGAGCGTA

I would greatly appreciate if someone could help me point what I missed.

Thank you,
Sharmi

evolvedmicrobe:
It appears there is no second column in your `genes.tsv` file, which is causing this error.  Can you retry the command with the `gene.column` argument set to `1` and let me know if it works? e.g.

`Read10X("same_directory", gene.column=1)`

BSharmi:
I just tried it after posting and it worked. Thank you so much!

evolvedmicrobe:
Great! I'll submit a PR to issue a more informative error message about this later.

July233:
very thanks.I settled the problem too!

bagoods:
Awesome, this helped me too thanks!

xiandxing:
Thanks, a nice answer!

DexinYang1998:
thank you a lot

justinesjw:
Thank you! it helped me :)

Hu-Clala:
Thank you soooooooooo much!

PR #1394, opened 2019-04-18
New Read10X Error Message

evolvedmicrobe:
Removed some commented out code and added an error message to help avoid: https://github.com/satijalab/seurat/issues/1388


andrewwbutler:
Thanks Nigel!

PR #1551, opened 2019-05-16
Fixed Description of return.only.var.genes default

mauriziopaul:
Fixed the description of the default setting for return.only.var.genes (currently it says FALSE, but the setting is TRUE).

# Note


mauriziopaul:
Hi Christoph, thanks for responding, and yes, it turns out I was misreading what the first "default is" referred to. I reverted that change, and made one edit on a separate default message. Hope this helps.

PR #1568, opened 2019-05-20
Adding global title option to CombinePlots

milanmlft:
Hi Seurat devs,

I just added a small feature that might be of interest: adding global title to the plot grid created by `CombinePlots`. It just uses `plot_grid` from `cowplot` to add a title on top of the aggregated plots.

Please let me know what you think!


mojaveazure:
Hi Milan,

While this is really cool, I'm curious as to what the benefit of this over recommending [`ggpubr::annotate_figure`](https://rpkgs.datanovia.com/ggpubr/reference/annotate_figure.html).

```R
ggpubr::annotate_figure(
  p = FeaturePlot(object = pbmc_small, features = c('MS4A1', 'LYZ')),
  top = ggpubr::text_grob(label = 'Plot Title', face = 'bold')
)
```

Sure, it's convenient to have everything in a single function, but yielding to already existing external functions allows greater customization of these plot titles than we could ever hope to provide in `CombinePlots` (for example, your `CombinePlots` requires a centered, bolded title but yielding to an external function would allow the user any possible title using any possible font face at any possible position). Moreover, if we did add `suptitle` as a parameter to `CombinePlots`, we'd need to ensure that all functions that use `CombinePlots` are able to pass it, not just `FeaturePlot`.

Our goal is to one day deprecate `CombinePlots` and use only external packages for combining and managing multiple ggplot objects (hopefully, [patchwork](https://github.com/thomasp85/patchwork) or a package offering similar capabilities will go to CRAN in the near future), so the less code we have in and dependent on `CombinePlots`, the better. I'm inclined to leave this functionality to other packages, but if you feel strongly about including it (or anything we can do to make this external functionality better-known to the user), I'm happy to discuss it.

milanmlft:
Hi Paul,

Thanks for the feedback. You make a very good point and to be honest, I didn't know yet about the `ggpubr::annotate_figure` function... so thanks for that!

Indeed, my main reason for including this was for convenience but as you say, using external packages would allow far greater customization so that makes more sense. I completely understand if you don't want to include this functionality in Seurat, especially if you plan to deprecate `CombinePlots`, so please feel free to close this PR.

Still, it might indeed be a good idea to recommend the use of `ggpubr::annotate_figure` somewhere in the documentation or vignettes, since more people might be looking for this feature.

Cheers!

Issue #1737, opened 2019-06-24
FeaturePlot fails with cells argument


alexvpickering:
The error is `arguments imply differing number of rows: 84, 173`.

Ran this after data integration (but don't think it is specific to):
```
Seurat::FeaturePlot(scseq, gene, cells = cells, reduction = 'umap')
```
This issue appears to be in `FetchData` where the idents are not being subsetted according to cells:

```R
# Pull identities
  if ('ident' %in% vars && !'ident' %in% colnames(x = object[[]])) {
    data.fetched[['ident']] <- Idents(object = object)
  }
```

This seems to fix it: 

```R
# Pull identities
  if ('ident' %in% vars && !'ident' %in% colnames(x = object[[]])) {
    data.fetched[['ident']] <- Idents(object = object)[cells]
  }
```

Shouldn't introduce any new bugs because `cells` is either all or a subset of column names:

```R
cells <- cells %||% colnames(x = object)
```

I'll submit a PR.

mojaveazure:
Fixed in #1738

PR #1738, opened 2019-06-24
fixes satijalab/seurat#1737

alexvpickering:
idents were not being subsetted by `cells` argument to `FetchData`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 3.0.3.9004
+Version: 3.0.3.9005
 Date: 2019-06-25
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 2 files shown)


mojaveazure:
Thanks Alex!

DirtyHarry80:
I have a similar problem.
Have defined my cells which I want to focus on from the Seurat object in form of a vector. I did prepare two vectors using CellSelector and another one directly by FetchData function. When I run TSNEPlot function, I can nicely set the argument "cells" as equal to my vector. Plotting works. When I want to do the same in FeaturePlot, now highlighting a gene, I get an error message:

Error in (function (..., row.names = NULL, check.rows = FALSE, check.names = TRUE,  : 
  arguments imply differing number of rows: 33, 8444

Any idea?

Issue #1048, opened 2019-01-10
Customize colors for DoHeatmap function in v3


MJ-Yang:
Hello, 
I really appreciate your efforts to make such useful tool for single cell RNA seq.
I've been using v2 for the whole time and now adjusting to v3.

I've found that DoHeatmap function has changed a bit. Previously, we were able to set colors manually through col.low ~col .high, but I've found it's missing in v3. Is it deprecated?

If so, how can I set colors manually for DoHeatmap function?

andrewwbutler:
Hi,

Since `DoHeatmap` returns a ggplot object in Seurat v3, you can manipulate the colors by specifying the color scale. For example

```R
library(ggplot2)
DoHeatmap(object = pbmc_small) + scale_fill_gradientn(colors = c("blue", "white", "red"))
```

You can also easily apply color schemes provided by other packages, such the viridis color palettes.
```R
library(viridis)
DoHeatmap(object = pbmc_small) + scale_fill_viridis()
```

lima1:
Hi, I think it's currently not possible to change the color of group bar, right? 

DoHeatmap(object = pbmc_small) + scale_color_manual(values=cbPalette)

andrewwbutler:
No, unfortunately it's not currently possible to change those colors though we would welcome a PR for that feature if anyone's interested in implementing.

Dragonmasterx87:
Single a google search defaults to this I thought i'd add this in case someone would be interested 

[DoHeatmap](https://www.rdocumentation.org/packages/Seurat/versions/2.3.4/topics/DoHeatmap)

`group.colors = c("red", "blue")` adds colors to the bar.

lwhitmore:
Does anyone know if you can sort the groups (columns) in the heatmap specifically in doheatmap?

kaizen89:
@lwhitmore You can do this by changing the order of levels of the variable that you want to sort by.
```
levels(factor(object@variable_to_sort))
> levels: 1,2,3
object@variable_to_sort <- factor(object@variable_to_sort, levels=c(3,2,1))
levels(object@variable_to_sort)
levels: 3,2,1
```

Landau1994:
Follow the @andrewwbutler idea  , we can do the following
```R
mapal <- colorRampPalette(RColorBrewer::brewer.pal(11,"RdBu"))(256)
DoHeatmap(pbmc_small, 
          angle = 90,size = 3)+
  scale_fill_gradientn(colours = rev(mapal))
```

[...]

PR #1815, opened 2019-07-11
custom group colors DoHeatmap/SingleRasterMap

dpschreiner:
refer to issue  #1048

this works for me. if you customize your heatmap colors (expression) and use space between groups, then you have to add another argument, i.e.

scale_fill_gradientn(colors = c("blue", "white", "red"), **na.value = 'white'**)


dpschreiner:
oh, it looks like it failed due to the example. that was foolish - i didn't know it would try to execute it.  was trying to be helpful!

mojaveazure:
Hi David,

Thanks for the PR! I've made some changes, generally moving the logic into `DoHeatmap` rather than spread across multiple functions. I've also added the ability for `DoHeatmap` to fill in colors if too few were specified in `group.colors`. Let me know if you have any questions/comments/concerns about these changes and once resolved, we'll get it merged in.

Dragonmasterx87:
Hi @mojaveazure do you have any idea how to change the color of the label bar on top of the heatmap? Please if you could suggest a line of code which could allow for the implementation of something like HTML codes, so that we have multiple color options for the label.

Cheers!
Fahd

Dragonmasterx87:
My apologies I just saw [this](https://rdrr.io/cran/Seurat/man/DoHeatmap.html) 

🤦‍♂

Issue #1645, opened 2019-06-06
Leiden algorithm cast SNN to dense


andrewhill157:
FindClusters has an option for the leiden algorithm, but as far as I can tell it casts the adjacency matrix to a dense matrix prior to generating the graph. I assume this is required given the way leidenalg is being called currently, but do you have any suggestions for large datasets where this cast to dense is more prohibitive? No worries if not, and realize might be lower on priority list, just wanted to check!

andrewwbutler:
It's a good point. @TomKellyGenetics do you think there's a way to avoid casting to a dense matrix here?

TomKellyGenetics:
It should be possible to pass an SNN graph directly to the python package (which takes this as an input). I'm open to implementing this but will take some time to look into how to go about it. At the moment, the "leiden" package in R avoids using the igraph library deliberately to reduce the size of the install.

See issue on the standalone R package: https://github.com/TomKellyGenetics/leiden/issues/1

Currently, the adjacency matrix is passed to python, then the SNN graph is computed in Python which is compatible with the Python leidenalg. For raw adjacency matrices, this was faster than computing SNN graphs in R (but could be reconsidered for Seurat objects). If it is worthwhile for performance and handling large datasets, I'm okay to add support for graph methods and igraph dependancies back in (Seurat has this dependency already). There is good support for basic data structures such as a matrix to be passed between R and Python but I'm not sure it's as simple for the separate igraph implementations. A sparse "dgCmatrix" can also be passed to python (compatible with the current leiden version) but this doesn't seem to improve performance.

On the other hand, if performance is a concern, these changes would be incremental compared to a large overhaul using RCpp to cut out Python dependency altogether. This is also being considered and has been discussed with @vtraag and others on social media. I had someone offer to help with this but we've been short on time to implement it: https://github.com/TomKellyGenetics/leiden/issues/4

In short, it's _possible_ but I'm not sure how long it will take to make major changes. Pull Requests are welcome on the standalone package and we can migrate them onto the Seurat version once it is working.

TomKellyGenetics:
Just revisiting this issue to see if there are any more thoughts about it. Is there an urgent need to optimise the R implementation of Leiden for larger matrices? This would be quite involved but I'm open to suggestions or Pull Requests.

andrewhill157:
Sorry for the very late reply, I definitely don't have an urgent need for
it, was just looking to try out comparisons of Leiden to louvain for a few
[...]

PR #1858, opened 2019-07-22
Update leiden clustering to call package

TomKellyGenetics:
As discussed in #1645, the Leiden algorithm performs poorly on large matrices. This calls the "leiden" R package and allows for expanding arguments passed to include weights and igraph object.

Calling leiden on a small igraph object is slower than passing the adjacency matrix to a numpy array. Therefore casting to dense will remain the default. I recommend the "igraph" method to pass large sparse datasets to leiden to reduce memory needed. This method will remain optional (either are supported by the leiden package).

As the leiden package is now on CRAN, I propose to call the CRAN version so that it remains up to date with the R version without further changes to Seurat. I will make efforts to ensure that the package remains backwards compatible and that future versions do not change results. At this point it is fairly stable and will only change to improve performance or maintain compatibility with Python and C libraries.

The igraph method is available on CRAN (leiden 0.3.0). However, please note that this PR requires the *development version* of leiden. I will submit a minor update of leiden to CRAN for compatibility. This will enable passing of `random.seed` to leiden (as done for other clustering methods). For now, please see the github repo for the package.

https://github.com/TomKellyGenetics/leiden/tree/dev


TomKellyGenetics:
Leiden version 0.3.1 has been submitted to CRAN. It can be installed from source and is currently building on Travis. Windows builds will be available shortly. #1645

R/clustering.R
```diff
@@ -676,79 +687,35 @@ RunLeiden <- function(
...
-    ),
-    'SurpriseVertexPartition' = leidenalg$find_partition(
-      snn_graph,
-      leidenalg$SurpriseVertexPartition,
-      initial_membership = initial.membership,
-      weights = weights,
-      node_sizes = node.sizes
-    ),
-    stop("please specify a partition type as a string out of those documented")
+  if (method == "matrix") {
```
mojaveazure: Consider using `switch` instead of if/else statements. Swtiches are optimized for comparisons and are (mildly) faster

```R
> library(microbenchmark)
> cols <- c('red', 'blue', 'green')
> r1 <- microbenchmark({x <- sample(x = cols, size = 1); switch(EXPR = x, 'red' = 'yay', 'blue' = 'meh', 'nope'); })
> r2 <- microbenchmark({x <- sample(x = cols, size = 1); if (x == 'red') 'yay' else if (x == 'blue') 'meh' else 'nope'; })
> r1$expr <- 'switch'
> r2$expr <- 'if/else'
> rbind(r1, r2)
Unit: microseconds
    expr   min     lq    mean median     uq     max neval
 if/else 6.068 6.3380 7.29976 6.5735 6.8275  51.545   100
  switch 5.914 6.1365 9.27236 6.2505 6.5160 249.691   100
```

R/clustering.R
```diff
@@ -676,79 +687,35 @@ RunLeiden <- function(
...
+    #cast to dense (supported by reticulate for numpy.array)
```
mojaveazure, line 699: `input` is never used, should this be `object`?

R/clustering.R
```diff
@@ -676,79 +687,35 @@ RunLeiden <- function(
...
+    input <- as(object, "matrix")
```
mojaveazure: Please remember to import `as` from `methods`

R/clustering.R
```diff
@@ -676,79 +687,35 @@ RunLeiden <- function(
...
+  } else if (method == "igraph") {
+    #Graph is a sparse dgCMatrix
+    is(object, "dgCMatrix")
```
mojaveazure: This line has no affect on the outcome of the function, was this used for testing? If you are using it, please remember to import `is` from `methods`

R/clustering.R
```diff
@@ -676,79 +687,35 @@ RunLeiden <- function(
...
+    #run as igraph object (passes to reticulate)
+    object <- graph_from_adjacency_matrix(adjmatrix = object)
+  } else {
+    warning("method for leiden must be matrix or igraph")
```
mojaveazure: Should this be a `warning` or a `stop`?

TomKellyGenetics:
Thanks for your feedback, the windows build is on CRAN now so I’ll make sure this version passes checks and then refactor as suggested.

R/clustering.R
TomKellyGenetics: Changed to `stop`

R/clustering.R
TomKellyGenetics: This was used to check that a Seurat `Graph` object can be handled as a sparse matrix (from which an igraph object can be generated). We could remove this but I was considering to use it to check that object is compatible with igraph. I've updated this to a `switch` statement as well (with should handle future changes to the input object).

R/clustering.R
TomKellyGenetics: I've added `is` and `an` to import from `methods` in case this is needed. I'm not sure if this is required by CRAN for base R packages.

R/clustering.R
TomKellyGenetics: I've updated the `input` to be the object passed to `leiden::leiden`. This is purely to avoid ambiguity which `object` is being passed to python.

R/clustering.R
TomKellyGenetics: Updated for performance (as has another if/switch statement added).

TomKellyGenetics:
Thanks for your feedback. I've updated the PR branch as suggested. This version runs all checks locally without issues. The windows build seems to be working with the new leiden version now too!Sorry it's taken a while to address #1645 but this should take care of it. Calling leiden as a dependency should make it easier to maintain rather than having multiple versions.

R/clustering.R
mojaveazure: Thanks! We've been rejected for not importing things from `methods`, `utils`, or `graphics` in the past. The only functions we don't need to import, as least as told to us by CRAN when we first went there two years ago, was all functions not from `base` or provided by the package being submitted itself needed to be imported. I don't know if they've changed that, but better to be safe than sorry.

mojaveazure:
Thanks for the PR!

R/clustering.R
TomKellyGenetics: Looking back into my own packages that have been checked through CRAN, it seems that I have had to do this as well. I'll keep in mind for future releases. 😄

PR #1935, opened 2019-08-04
ALRA update: switching to quantiles, adding support for intel MKL implementation of randomized SVD, bug fix

linqiaozhi:
We have made several changes to ALRA, and we were hoping to include them in this new version. These changes were pulled into the main ALRA repo [here](https://github.com/KlugerLab/ALRA/pull/6). Below, we summarize the changes:

-Added a flag `use.mkl` for users who have installed [rpca-mkl](https://github.com/KlugerLab/rpca-mkl), which allows for dramatic speedups over the default [rpca](https://www.rdocumentation.org/packages/rsvd/versions/1.0.1/topics/rpca)-based version. Note that rpca-mkl is still under development and is not on CRAN, so it is not a required package, but if users have already installed it then they can use it by setting this flag to True.

-When choosing `k`, we no longer use the the normal distribution to approximate the distribution of spacings and compute p-values. Instead, we set `k` to correspond to the largest spacing that is more than 6 standard deviations from the "noise" singular value spacings. This leads to effectively the same choice of `k` as the original version, but does not make the normality assumption (that is not true).

-Instead of thresholding based on the most negative value of each gene, we threshold based on the 0.001 quantile, which is more robust to outliers. By setting the `quantile.prob` to 0, you will restore the original behavior of thresholding by the most negative value.

-Fixed bug which introduced NAs during the scaling process for some datasets


R/preprocessing.R
```diff
@@ -1838,22 +1838,62 @@ NormalizeData.Seurat <- function(
...
+  ...
+) {
   CheckDots(...)
   A.norm <- t(x = as.matrix(x = object))
   message("Identifying non-zero values")
   originally.nonzero <- A.norm > 0
   message("Computing Randomized SVD")
-  fastDecomp.noc <- rsvd(A = A.norm, k = k, q = q)
+  if (use.mkl) {
+    fastDecomp.noc <- setNames(
```
mojaveazure: Please remember to declare imports from packages that are neither base nor Seurat

R/preprocessing.R
```diff
@@ -1838,22 +1838,62 @@ NormalizeData.Seurat <- function(
...
+      object = vector(mode = "list", length = 3),
+      nm = c("u", "d", "v")
+    )
+    fastPCAOut <- fastRPCA::fastPCA(
```
mojaveazure: Where does fastRPCA come from? If it's not on CRAN, Bioconductor, or another CRAN-like repository like we use for loomR, we unfortunately can't include it.

Also, please ensure all non-imported but still used packages are listed in the `Suggests` field in `DESCRIPTION`

R/preprocessing.R
```diff
@@ -1838,22 +1838,62 @@ NormalizeData.Seurat <- function(
...
     t(x = fastDecomp.noc$v[,1:k])
-  message("Find most negative values of each gene")
-  A.norm.rank.k.mins <- abs(x = apply(X = A.norm.rank.k, MARGIN = 2, FUN = min))
+  message(sprintf("Find the %f quantile of each gene", quantile.prob))
+  # A.norm.rank.k.mins <- abs(x = apply(X = A.norm.rank.k, MARGIN = 2, FUN = min))
+  A.norm.rank.k.mins <- abs(x = apply(
+    X = A.norm.rank.k,
+    MARGIN = 2,
+    FUN = function(x) {
+      return(quantile(x = x, probs = quantile.prob))
```
mojaveazure, line 1919: Please remember to declare imports from packages that are neither base nor Seurat

R/preprocessing.R
linqiaozhi: [fastRPCA](https://github.com/KlugerLab/rpca-mkl) is not on CRAN, and it is still under active development. This is mainly because of its dependence on Intel MKL. In order to not require users to install Intel MKL on their computers, we have to distribute a binary...and that makes it difficult to submit to CRAN. 

ALRA does not depend on fastRPCA; it is only used if the user sets the `use.mkl` argument indicating they installed this package from the Github repository. Given that it will not interfere with normal usage (i.e. no one will be required to install this package), is it still not allowed? From your comment I understand that to be the case, but I want to confirm before we remove it.

linqiaozhi:
> Hi George,
> 
> This looks great! I've made some minor style changes and would like you to check some imports before we merge this in.

Thanks for your help, @mojaveazure!

> Also, in an attempt to slim down the Seurat codebase, we've started moving some functionality to [SeuratWrappers](https://github.com/satijalab/seurat-wrappers/). We do this to relax CRAN-compliance on these methods (including allowing packages that are hosted only on GitHub), allow easier and faster development from the community, and serve as a place to highlight these community-provided methods. `RunALRA` and `ALRAChooseKPlot` are both destined to be migrated (you may have noticed the deprecation warnings), with [vignettes already hosted at SeuratWrappers](https://github.com/satijalab/seurat-wrappers/blob/master/docs/alra.md) showcasing ALRA on Seurat objects. We're happy to merge this in here for our next release and will take care of moving these changes to SeuratWrappers, but would like to point you there for future updates to `RunALRA`

This makes a lot of sense. After you merge it, we will make any further updates to the SeuratWrappers repository.

R/preprocessing.R
mojaveazure: Unfortunately, that is correct. If we use a package in Seurat, CRAN requires that it be hosted on CRAN, Bioconductor, or another CRAN-like repository. Using a package hosted only on GitHub is not allowed for CRAN-hosted packages, and if fastRPCA isn't going to CRAN anytime soon, we do ask that it be removed from this PR.

However, we can include it in SeuratWrappers; that package will not be going to CRAN as it was designed to allow `Seurat` object-based methods using packages not hosted on CRAN, Bioconductor, or another CRAN-like repository.

linqiaozhi: > Unfortunately, that is correct. If we use a package in Seurat, CRAN requires that it be hosted on CRAN, Bioconductor, or another CRAN-like repository. Using a package hosted only on GitHub is not allowed for CRAN-hosted packages, and if fastRPCA isn't going to CRAN anytime soon, we do ask that it be removed from this PR.

That makes sense. 

> However, we can include it in SeuratWrappers; that package will not be going to CRAN as it was designed to allow Seurat object-based methods using packages not hosted on CRAN, Bioconductor, or another CRAN-like repository.

Given this, before I go ahead and remove the fastRPCA compatibility, I was wondering about your timeframe for moving ALRA to SeuratWrappers. Since we can include this functionality after it is moved to SeuratWrappers---and you are planning on moving it anyway---perhaps we should just wait until then to update ALRA in Seurat?

mojaveazure: We already have an implementation (basically copy-pasted from Seurat) [available in SeuratWrappers](https://github.com/satijalab/seurat-wrappers/blob/master/R/alra.R). We are currently preparing Seurat v3.1.0, in which `RunALRA` will be marked as deprecated. We will likely mark it as defunct in a later point release of the v3.1.X series (somewhere around v3.1.1, v3.1.2, or v3.1.3) or at the latest v3.2.0, at which point `RunALRA` in Seurat will cease to work and users will be directed to SeuratWrappers.

Since we already have an implementation there, we can take your updates here, including usage of fastRPCA, and incorporate them into the SeuratWrappers implementation of `RunALRA`. In addition, we can add everything except fastRPCA usage here, with a message if `use.mkl = TRUE` directing users to `RunALRA` in SeuratWrappers.

mojaveazure:
Hi George,

I've taken care of the changes we discussed. If you could give everything one last look over, both here and at satijalab/seurat-wrappers#8, that'd be great.

linqiaozhi:
@mojaveazure This looks great, thanks for making the changes!

PR #1969, opened 2019-08-13
Unable to wrap PCA/TSNA/UMAPPlot function in custom function

kriemo:
```r
library(Seurat)
data("pbmc_small")

make_plot <- function(object){
  PCAPlot(object)
}

make_plot(pbmc_small)
##'Error in `[[.Seurat`(object, reduction) : 
##'Cannot find 'make_plot' in this Seurat object
```

`SpecificDimPlot()` checks the first function of the call stack to determine which reduction (pca, tsne, etc) should be called. This behavior breaks when another function (in this case `make_plot`) is higher in the call stack.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 3.0.3.9029
+Version: 3.0.3.9030
 Date: 2019-08-09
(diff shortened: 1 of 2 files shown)


mojaveazure:
Hi Kent,

Thanks for catching this!

PR #1974, opened 2019-08-13
Correcting version dependency - dot to dash

smoe:
Hello, I have just created a Debian package for Seurat and it is these kind of small things that are then spotted because the auto-generated dependency list could not be matched.


mojaveazure:
Hi Steffen,

Thanks for catching this!

PR #2102, opened 2019-09-13
Fix/activitymatrix

timoast:
Modifications to #2100

PR #2131, opened 2019-09-23
Fix activity matrix

drbecavin:
See #2102 for fix explanation


drbecavin:
I do not know why Travis crash, I just changed 1 line

andrewwbutler:
Hi @drbecavin,

Sorry for missing this earlier but I believe we caught the same issue when submitting a patch release (3.1.1) to CRAN recently. If you try reinstalling the latest version, this should be working again.

Issue #2168, opened 2019-10-04
Leiden algorithm with igraph method


orrzor:
Hi Seurat team  and @TomKellyGenetics ,
I am having trouble running the Leiden algorithm with the igraph method #1645. I see this error when I run on both my own data, and the PBMC vignette https://satijalab.org/seurat/v3.1/pbmc3k_tutorial.html.

```
library(dplyr)
library(Seurat)

# Load the PBMC dataset
pbmc.data <- Read10X(data.dir = "/Users/orr/Downloads/seurat/filtered_gene_bc_matrices/hg19/")

# Initialize the Seurat object with the raw (non-normalized data).
pbmc <- CreateSeuratObject(counts = pbmc.data, project = "pbmc3k", min.cells = 3, min.features = 200)

pbmc[["percent.mt"]] <- PercentageFeatureSet(pbmc, pattern = "^MT-")

pbmc <- subset(pbmc, subset = nFeature_RNA > 200 & nFeature_RNA < 2500 & percent.mt < 5)

pbmc <- NormalizeData(pbmc, normalization.method = "LogNormalize", scale.factor = 10000)
pbmc <- FindVariableFeatures(pbmc, selection.method = "vst", nfeatures = 2000)
pbmc <- ScaleData(pbmc, features = rownames(pbmc))
pbmc <- RunPCA(pbmc, features = VariableFeatures(object = pbmc))
pbmc <- FindNeighbors(pbmc, dims = 1:10)
pbmc <- FindClusters(pbmc, resolution = 0.5, algorithm = 4, method = "igraph", random.seed = 100) # fail
pbmc <- FindClusters(pbmc, resolution = 0.5, algorithm = 4, method = "matrix", random.seed = 100) # success

pbmc <- RunUMAP(pbmc, dims = 1:20)
DimPlot(pbmc, reduction = "umap")
```

I get the following error
```
pbmc <- FindClusters(pbmc, resolution = 0.5, algorithm = 4, method = "igraph", random.seed = 100) # fail
Warning: The following arguments are not used: method
Error in switch(EXPR = is(object), Graph = input <- graph_from_adjacency_matrix(adjmatrix = object),  : 
  EXPR must be a length 1 vector
```

Also when I use the matrix method, I get the warning
```
pbmc <- FindClusters(pbmc, resolution = 0.5, algorithm = 4, method = "matrix", random.seed = 100) # success
Warning: The following arguments are not used: method
```

[...]

PR #2169, opened 2019-10-04
Patch to correct method inputs for Leiden

TomKellyGenetics:
Implement method input for FindClusters.Seurat: #1645 #2168

Enables "igraph" input to FindClusters (passed to FindClusters.default) to avoid casting to a dense matrix. An igraph matrix is computed from a sparse dgCMatrix.


R/clustering.R
```diff
@@ -13,6 +13,9 @@ NULL
 #' @importFrom pbapply pblapply
 #' @importFrom future.apply future_lapply
 #' @importFrom future nbrOfWorkers
+#' @importFrom igraph graph_from_adjacency_matrix
+#' @importFrom methods as
```
mojaveazure: Is this being used in `FindClusters.default` anywhere? I can't find any calls to `methods::as`

R/clustering.R
```diff
@@ -13,6 +13,9 @@ NULL
...
+#' @importClassesFrom Matrix dgCMatrix
```
mojaveazure: Why are we importing the dgCMatrix class? This should only be necessary if we're creating a subclass, which doesn't seem to be happening.

R/clustering.R
```diff
@@ -82,6 +85,13 @@ FindClusters.default <- function(
             edge.file.name = edge.file.name
           )
         } else if (algorithm == 4) {
+          if (method == "matrix" || method == "igraph"){
+            if (method == "igraph"){
+            object <- graph_from_adjacency_matrix(object)
+            }
+          } else {
+            stop("method must be 'matrix' or 'igraph'")
+          }
```
mojaveazure: Why are we doing these checks if we're parallelizing the clustering but not otherwise? Also, wouldn't this be better implemented in `RunLeiden`?

R/clustering.R
```diff
@@ -82,6 +85,13 @@ FindClusters.default <- function(
             edge.file.name = edge.file.name
           )
         } else if (algorithm == 4) {
+          if (method == "matrix" || method == "igraph") {
+            if (method == "igraph") {
+              object <- graph_from_adjacency_matrix(adjmatrix = object)
+            }
+          } else {
+            stop("method must be 'matrix' or 'igraph'")
+          }
```
mojaveazure: Why is this check happening if we're parallelizing the clustering but not otherwise? Also, shouldn't this check be in `RunLeiden` to avoid duplicating code?

R/clustering.R
```diff
@@ -711,7 +723,7 @@ RunLeiden <- function(
     "matrix" = input <- as(object, "matrix"),
     #run as igraph object (passes to reticulate)
     "igraph" = switch(
-      EXPR = is(object),
+      EXPR = is(object)[1],
       #generate igraph if needed (will handle updated snn class)
       "Graph" = input <- graph_from_adjacency_matrix(adjmatrix = object),
       "dgCMatrix" = input <- graph_from_adjacency_matrix(adjmatrix = object),
```
mojaveazure: I feel like this whole `switch` construct for the igraph method could be simplified using `inherits`; this way, people can subclass and use their own custom objects and still have it work with `RunLeiden`

```R
input <- if (inherits(x = object, what = 'list')) {
  graph_from_adj_list(adjlist = object)
} else if (inherits(x = object, what = c('dgCMatrix', 'matrix'))) {
  graph_from_adjacency_matrix(adjmatrix = object)
} else if (inherits(x = object, what = 'igraph')) {
  object
} else {
  stop(...)
}
```

R/clustering.R
mojaveazure: Whoops, forgot to refresh the files changed list

R/clustering.R
TomKellyGenetics: Ah, didn't realise there was a separate RunLeiden call in the parallel version. Yes let's move it there so it's called by both.

R/clustering.R
TomKellyGenetics: Sorry, yes these have been removed.

R/clustering.R
TomKellyGenetics: Ah, sorry I misunderstood if importing class/methods was needed.

R/clustering.R
TomKellyGenetics: Complete agree, this is a better approach. I started with this as simple if statement (matrix or igraph) but it's gotten more complex for sparse matrices, lists, etc. I got a bit stuck on how to handle a Seurat "Graph" object (sparse adjmat). It seems to inherit "Matrix" (not "matrix" or "dgCMatrix").

TomKellyGenetics: I think in either case `graph_from_adjacency` does not require casting to dense to make a `igraph` object. #1645

TomKellyGenetics:
Thanks, I've made the suggested changes so this should take care of it.

R/clustering.R
mojaveazure: Interesting, the [class definition for `Graph` objects](https://github.com/satijalab/seurat/blob/master/R/objects.R#L153-L156) is simply inheriting from `dgCMatrices` so I wonder why it needs to check inheritance from `Matrix` objects. Either way, doesn't hurt.

mojaveazure:
Thanks for the patch!

PR #2182, opened 2019-10-07
bug fix for RunTSNE when supplying features directly

Yue-Jiang:
I think you need to transpose this matrix so that each row is an observation and each column is a feature. To see this, try 

```
DefaultAssay(cbmc) <- "ADT"
features <- c("adt_CD3", "adt_CD11c", "adt_CD8", "adt_CD16", "CD3E", "ITGAX", "CD8A", 
    "FCGR3A")
cbmc <- RunTSNE(cbmc, features = features, reduction.name = "tsne_adt", reduction.key = "adtTSNE_", dims = NULL, reduction = NULL)
```
and it gives an error
```
Error in .check_tsne_params(nrow(X), dims = dims, perplexity = perplexity,  : 
  perplexity is too large for the number of samples
```
it thinks the number of samples is too small because it's currently the number of ADT features.

cbmc as in https://satijalab.org/seurat/v3.1/multimodal_vignette.html


andrewwbutler:
Hi Yue Jiang,

Thanks for the fix! (and sorry for the delayed response).

Yue-Jiang:
No worries! Thanks for fixing the style:)

PR #2268, opened 2019-10-30
Axis label

Alanocallaghan:
This is a minor edit; it just corrects the y axis label for `ExIPlot`s


R/visualization.R
```diff
@@ -3604,7 +3604,7 @@ ExIPlot <- function(
     FUN = function(x) {
       return(SingleExIPlot(
         type = type,
-        data = data[, x, drop = FALSE],
+        data = data[, features, drop = FALSE],
```
andrewwbutler: Why change this to `features`? I believe `SingleExIPlot` expects a single column data.frame.

alanocallaghan: That's unintended, I meant to stash and discard that. I'll remove

alanocallaghan: Done, should be up to date with devel also

andrewwbutler:
Thanks for the fix!

Alanocallaghan:
Thanks for the quick response!

PR #2284, opened 2019-11-04
Use palettes from pals package

### LINKED ISSUE #2264: DimPlot could use better colour defaults
It's really common to see clustering plots made with Seurat with the default ggplot2 discrete colour scale. This scales (haha) very badly with increased K, with clusters eventually becoming indistinguishable.

It would be great if you would consider using a palette which is more useful for the typically large K in modern single cell experiments. Obviously above ~25-30 this becomes difficult, but under that is fairly doable ime. If you agree I'm more than happy to submit a PR.

example of a bad plot (several clusters blend together, unclear whether some are distinct):
bad

less bad (alphabet palette from pals package):
lessbad

Alanocallaghan:
Currently this adds the option to choose palettes from pals, and also automatically chooses one if there are few enough unique values. If it's something that would be useful elsewhere in the package I could refactor this out into a common discrete scale before merging.

Resolve #2264


Alanocallaghan:
Hey, I suspected that'd be an issue - the list of dependencies is already quite long. Honestly, since the dependency is just for colour palettes, my plan would just be to copy the colours with attribution into the package.

Alanocallaghan:
fyi pals is also GPL v3 so I am pretty sure that's absolutely permitted under licence

Alanocallaghan:
Alright @mojaveazure - I've removed the dependency, and things appear to work okay on my end. I also added a token test for the palette function. Let me know if there's anything else I ought to do :)

Alanocallaghan:
@mojaveazure - I removed the dependency and fixed the build errors, so this should be good to look at again

Alanocallaghan:
@mojaveazure - any chance you can have another look when you have time? I think I've made all requested changes

mojaveazure:
Hi Alan,

Sorry for the delay, we've been a bit busy with the spatial and patch releases. Thanks for the fixes and thanks again for the PR!

Alanocallaghan:
No worries, I see the volume of issues on here at the best of times, never mind when you're also adding new functionality/data types! Cheers :+1:

PR #2389, opened 2019-12-04
Don't materialize dense matrix in AverageExpression and FindMarkers

troels:
Hi!

I've been encountering some problems while analyzing scRNA-experiments with Seurat:

Calling apply on the expression matrix in AverageExpression and FindMarkers, makes the expression matrix dense.

This densification can accidentally use up a lot of memory and sometimes spawn too many cells for R's matrix package to handle.

By using rowMeans the dense middle result matrix is avoided and much less memory is consumed, while increasing performance slightly.

Additionally I've added a test for AverageExpression.


andrewwbutler:
Thanks for the PR! That's definitely a more efficient way of handling the averaging for those functions.

PR #2469, opened 2020-01-02
Fix getEdges() Method of the Network Class

evolvedmicrobe:
The `getEdges` method was a public method of the `Network` class that allows users to obtain a copy of the Edges of a Network object.  This method/interface was ported over from Java to C++ with a bug in it, though because the `getEdges` method is never called by either the Java or C++ version of the code, this bug had no effect on the current code.  However, to improve the code and maintain it's similarity with the original Java version, it's probably worth fixing this unused method.

The specific bug was that instead of filling an array from `start + value_1` to `start + value_2` with a constant value, it was filled from `start + value_1` to `end + value_2` with a constant value, which leads to unexpected behavior due to an array overflow.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Seurat
-Version: 3.1.2.9000
-Date: 2019-12-16
+Version: 3.1.2.9002
+Date: 2020-01-03
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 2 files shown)


evolvedmicrobe:
Hey guys, happy 2020! Just back to work and catching up on emails when I saw this, looking forward to checking out the new stuff later this week.  I hope everyone had a good holiday break.

andrewwbutler:
Thanks for the fix Nigel and happy new year to you as well!

PR #2569, opened 2020-02-03
Link to sctransform methods that populate slots of SCTransform output

amcdavid:
This will hopefully make it clearer to users what they are getting, and why in the returned object


mojaveazure:
Hi Andrew,

Thanks for catching this!

Issue #1260, opened 2019-03-21
FeaturePlot issues and suggestions in Seurat3


jeremycfd:
I've noticed a few problems with the Seurat3 FeaturePlot function that I wanted to provide feedback on:
1) When using the split.by option, FeaturePlot correctly separates according to the factor of interest; however, it seems that each sub-plot scales the color (corresponding to feature expression) separately. This only became obvious to me when I plotted a feature that is not expressed in any cells in one of the sub-plots. The default behavior with facet_wrap() is to maintain the color scale across all subplots, and I think that should probably be the default here as well (and also provide the option to display that scale if possible); otherwise, I fear a lot of people will be inadvertently mislead by these plots.

2) The ncol option in FeaturePlot doesn't seem to work as far as I can tell, or perhaps just not how I envision it working. Setting it to any value doesn't seem to change the resulting plot. I think most people would appreciate being able to control this in a way similar to nrow and ncol options in ggplot2's facet_wrap(), at least if there is only a single feature being plotted, which currently can be done by tacking on the facet_wrap() to the FeaturePlot object.

3) I understand if you don't want to implement this as a feature, but do you have any suggestions on how to edit the font size for different aspects of the FeaturePlots through ggplot() options? I've dug through the FP object but haven't been able to find where those sorts of things are defined.

Thanks!

andrewwbutler:
Hi,

Thanks for the feedback!

1. I agree. The `split.by` option still needs a bit of work regarding scaling/legends. Will post back here when we have any updates.
2.  I'm not sure what issue you're having with `ncol`. Using `pbmc_small`, this example shows changing from 3 to 1 column. Could you provide an example where it breaks?
```R
FeaturePlot(object = pbmc_small, features = c("MS4A1", "LYZ", "CD3D"), ncol = 3)
FeaturePlot(object = pbmc_small, features = c("MS4A1", "LYZ", "CD3D"), ncol = 1)
```
3. When possible, all plots that Seurat v3 produces are made using the ggplot2. Things like font size are controlled through the theme framework. 
https://ggplot2.tidyverse.org/reference/theme.html
So for example, to change the axis text size:
```R
FeaturePlot(object = pbmc_small, features = "MS4A1") + theme(axis.text = element_text(size = 20))
```

bumproo:
Hi,
[...]

PR #2620, opened 2020-02-14
Consistent `FeaturePlot` axis limits

mfansler:
This partially addresses what is documented in Issue #1260. In particular, it fixes a bug in `FeaturePlot` that currently causes the shared y-axis limits to be discarded during subsequent changes to the axis.

Ideally, a proper fix would rework the code to avoid overwriting the axis scaling in the first place. Instead, this PR minimally patches the current code to make sure the `limits` argument of `scale_y_continuous` is consistently set throughout.


mfansler:
Looking at this again, I realized that the issue I cited does not actually document the bug this fixes. Below is an example, where splitting by samples (`condition` in the code below) results in subsets of cells that occupy different parts of the tSNE plot. The current `FeaturePlot` code fails to preserve the y-axis across the plots, making it less than ideal for visual comparison. This PR would instead yield consistent axes applied throughout.

---

### Example Code

``` lang: r
library(magrittr)
library(Seurat)

## data from Dobie et al., 2019
seu.chronic %>%
  subset(subset=(enrichment == "HSC")) %>%
  FeaturePlot(features=c("nGene", "nUMI", "percent.mito", "effectiveGenes"), 
              split.by="condition", label=TRUE, coord.fixed=TRUE,
              cols=viridis::plasma(20))
```

### Result
image

mojaveazure:
Thanks for catching this, Mervin!

sameetmehta:
This problem is not fully solved.  I used another dimensionality reduction, but the plot is with x, and y limits c(-1, 1).  If I trying to do `p + xlim(-0.05, 0.05)`, and I have used `split.by` in the `FeaturePlot`, only the last plot is plotted at correct limits, not the others.

mfansler:
> This problem is not fully solved. I used another dimensionality reduction, but the plot is with x, and y limits c(-1, 1). If I trying to do `p + xlim(-0.05, 0.05)`, and I have used `split.by` in the `FeaturePlot`, only the last plot is plotted at correct limits, not the others.

Hi @sameetmehta 

This patch was not intended to provide post-processing adjustments such as that shown by your example. AFAIK, currently there is not an interface in the API to manually specify axis limits. In order to get your example to work effectively as a post-processing adjustment, one would need to iterate through all the subplots in the ggplot object and add the `xlim(...)` directive to each.

If this is a feature you expect to frequently require and suspect others might want it, then it may be worth adding it to the API directly (i.e., avoid post-processing). In that case, consider creating a new issue for such a feature request and see if there is any feedback for how best to add it. At a glance, I think one could add `xlims` and `ylims` arguments to the `FeaturePlot` function, then check to see if those are non-NULL ([probably here in the code](https://github.com/satijalab/seurat/blob/0bbb2797e0a717aa7dc09d95781be53987f1907f/R/visualization.R#L1107)), using those values instead of the extremes of the data.

Issue #2629, opened 2020-02-17
MULTIseqDemux: Error in model(x) : could not find function "model"


AbbiEdwards:
I am attempting to demultiplex some ECCITE-Seq data, HTODemux doesn't seem to work for any of my replicates whatever parameter I use but it was suggested that MULTIseqDemux would be more suitable anyway.

I have one rep which runs perfectly through and gives expected clustering results but the other rep gives this error: Error in model(x) : could not find function "model"

It has been treated exactly the same, following the vignette using the output from cite seq count. In previous issues I have seen people suggest removing tag that haven't worked/ tags with zeros in the cite seq count output but there are no zeros that I can see. The only difference between the two is the first has 9000 cells and the second 8000.

htoRep2 <- Read10X("../cite-seq/rep2/umi_count", gene.column = 1)
rownames(htoRep2) <- str_remove(rownames(htoRep2), "-[^-]+$")
hashtags2 <- setdiff(rownames(htoRep2), "unmapped")
htoRep2 <- htoRep2[hashtags2,]
htoRep2 <- CreateSeuratObject(counts = htoRep2, assay = "HTO")
htoRep2 <- NormalizeData(htoRep2, assay = "HTO", normalization.method = "CLR")
htoRep2 <- MULTIseqDemux(htoRep2, assay = "HTO", verbose = TRUE, autoThresh = TRUE, qrange = seq(from = 0.1, to = 0.9, by = 0.05))

Am I doing something very stupid and obvious?

bbimber:
I just started hitting this too, FWIW

bbimber:
I think this is related to error handling in classifyCells.  this is the code from MULTIseq's implementation:

model <- tryCatch( { approxfun(bkde(barTable.n[,i], kernel="normal")) },
                   error=function(e) { print(paste0("No threshold found for ", colnames(barTable.n)[i],"...")) } )
if (class(model) == "character") { next }

and here is Seurat's port of it:

model <- tryCatch(
      expr = approxfun(x = bkde(x = data[, i], kernel = "normal")),
      error = function(e) {
        message("No threshold found for ", colnames(x = data)[i], "...")
      }
    )
    if (is.character(x = model)) {
      next
    }

In the first case, when an error occurs, model becomes the result of print(), so a string.  in the second case, the return is message(), which sets model to NULL.  Therefore the check "class(model) == "character")" doesnt work as intended.

AbbiEdwards:
This bugfix worked for me, thank you!

PR #2651, opened 2020-02-23
Bugfix Multiseq's classifyCells

bbimber:
# Note

See #2629.

I believe the differences in the tryCatch between Multiseq's implementation and the seurat port need different error handling.  Since you're using message(), the result from an error is NULL, not a string (like multiseq's code would do).  this small change should fix this.


andrewwbutler:
Makes sense to me. Thanks for the fix!

Issue #3117, opened 2020-06-07
Error in FindClusters


william0701:
When I run FindClusters, I get an error as:
```
Error in names(ids) <- colnames(x = object) : 
  'names' attribute [11740] must be the same length as the vector [11739]
```
The data I use is here [file](https://www.ncbi.nlm.nih.gov/geo/download/?acc=GSM4557338&format=file&file=GSM4557338%5FHIP044%5Fcell%2Ecounts%2Ematrices%2Erds%2Egz). And my process code is

```
 library(dplyr)
  library(Seurat)
  library(patchwork)
  library(ggplot2)
  library(SingleR)
  library(SingleCellExperiment)
  library(scater)
 
  pbmc_data <- readRDS(file = rds.file)$exon
  pbmc <- CreateSeuratObject(counts = pbmc_data,project = 'test')
  pbmc <- PercentageFeatureSet(pbmc, pattern = "^MT-", col.name = "percent.mt")
  pbmc <- SCTransform(pbmc, vars.to.regress = "percent.mt", verbose = FALSE)
  pbmc <- RunPCA(pbmc, verbose = FALSE)
  pbmc <- RunUMAP(pbmc, dims = 1:50, verbose = FALSE)
  pbmc <- FindNeighbors(pbmc, dims = 1:50, verbose = FALSE)
  pbmc <- FindClusters(pbmc, verbose = FALSE)
```

andrewwbutler:
Thanks for reporting! This should now be fixed on the `develop` branch.

PR #3203, opened 2020-06-29
fix edge case when last node has no edges

andrewwbutler:
This PR addresses https://github.com/satijalab/seurat/issues/3117

The issue arises when the last column in the SNN matrix has 0 edges that pass the edge weight filtering parameters. In the `matrixToNetwork` function, total number of nodes `nNodes` was set as the max element from the `node1` and `node2` vectors. However, if the only nonzero element in the SNN matrix for the last column is the diagonal 1 (self), the `nNodes` gets set to 1 fewer than the actual total number of nodes. This was why the function was returning one fewer cluster assignment and causing issues in the Seurat function that tries to store the cluster labels. 

Here I modified the `matrixToNetwork` function to directly take `nNodes` as a parameter that can be obtained from the matrix itself to avoid this issue. The other place that `matrixToNetwork` is used is in `readInputFile`. It was less clear what to do here and since it's not really used, I just defaulted back to the original behavior.

@evolvedmicrobe , if you have any thoughts/feedback, would be happy to incorporate.


src/RModularityOptimizer.cpp
```diff
@@ -82,8 +82,8 @@ IntegerVector RunModularityClusteringCpp(Eigen::SparseMatrix<double> SNN,
     if(node1.size() == 0) {
       stop("Matrix contained no network data.  Check format.");
     }
-
-    network = matrixToNetwork(node1, node2, edgeweights, modularityFunction);
+    int nNodes = std::max(SNN.cols(), SNN.rows());
```
cdarby, line 85: As I understand it, SNN is the weighted adjacency matrix, so it ought to be square. Maybe you should require that it is square to prevent further issues down the line?

andrewwbutler: It's a good point but I think checking for that should probably happen upstream of this function. Originally back in the v1 days, for this clustering functionality we were calling out to a java implementation provided by the authors of the clustering papers (http://www.ludowaltman.nl/slm/). It was then basically directly translated to this c++ implementation which was nice for both speed and it made distribution/installation easier. While we almost always tend to use it with the SNN matrix, the clustering algorithms themselves don't require the input network to have that structure and some of the tests that we have for it (in https://github.com/satijalab/seurat/blob/master/tests/testthat/test_modularity_optimizer.R)  don't actually use a square matrix as input.

Issue #3168, opened 2020-06-22
Leidenalg: FindClusters with method='igraph' with default weights=NULL is not good: suggestions to fix it


therealgenna:
Currently (Seurat v3.1.5, `Seurat:::RunLeiden`), using `method='igraph'` with the default of `weights=NULL` when calling `FindClusters(..., algorithm=4, ...)` is not working well, as it calls `igraph::graph_from_adjacency_matrix()` also with the default of `weighted=NULL`, which seems to drop all edge weights below 1 (I suppose it might use floor()); since all SNN weights are <=1, most of the edges are dropped yielding most of the vertices as singletons (which are later assigned to a nearby cluster with the default group.singletons=T of FindClusters.) 

The straightforward workaround is to pass `weights=T` in conjunction with `method='igraph'` - see Seurat:::RunLeiden. This produces clustering identical to the default `method='matrix'` (with the default `weights=NULL`).

Probably, the correct fix of the code would be to use `graph_from_adjacency_matrix(adjmatrix = object, weighted = TRUE)` inside `Seurat:::RunLeiden()` irrespective of passed argument `weights` being NULL or not. Presently, this is not the case.

However, the graph constructed by `igraph::graph_from_adjacency_matrix()` is by default directed (`mode='directed'`), and there is no way to change that via Seurat function arguments (need to change the code for that). I wonder why not use the `undirected` graph? This makes more sense to me, and has twice fewer edges (excl. diagonal); perhaps the method='matrix' code might need to be modified as well to be equivalent to method='igraph', weighted=T, mode='undirected' (or maybe not - see leiden:::leiden.matrix and leiden:::leiden.igraph).

`igraph::graph_from_adjacency_matrix(..., mode='undirected')` complains that matrix is not symmetric (which is not true - see below) and does not work, if passed Seurat Graph, but passing with explicitly set dimensions works:
```
tmp=dim(obj@graphs$RNA_snn)
snn_graph <- igraph::graph_from_adjacency_matrix(obj@graphs$RNA_snn[1:tmp[1],1:tmp[2]], mode='undirected', weighted=T)
```
Alternatively, can use mode='max' or or several other alternatives, if sure that the matrix IS symmetric.

```
Matrix::isSymmetric(obj@graphs$RNA_snn) # FALSE
# BUT:
all(Matrix::t(obj@graphs$RNA_snn)==obj@graphs$RNA_snn) # TRUE
tmp=dim(obj@graphs$RNA_snn)
[...]

Issue #2574, opened 2020-02-05
Control handling of weighted graphs for FindClusters


TomKellyGenetics:
As discussed (see: https://github.com/satijalab/seurat/pull/2440) for the Leiden algorithm, control of weighted graph inputs for `FindClusters` would be beneficial. Weights can be specified explicitly and passes to the clustering algorithm, however it would be more convenient to control whether or not weights are derived from the values of the adjacency matrix. This leads to very different results between when weights are used or not.

Therefore I propose to add a `weighted` argument (defaults as `weighted = TRUE`) to `FindClusters`. If no `weights` are given (`weights = NULL`) then the values of the adjacency matrix are used. As discussed with @andrewwbutler default support for weighted graphs is preferable.

https://github.com/satijalab/seurat/blob/49a1be0427f2f26a531eb468ba93eeb18d8a2edb/R/clustering.R#L39
https://github.com/satijalab/seurat/blob/49a1be0427f2f26a531eb468ba93eeb18d8a2edb/R/clustering.R#L153

Specific to `RunLeiden` I propose to change the `method = "igraph"` to create an undirected weighted matrix in this case. An undirected graph is preferable as the "shared nearest neighbours" adjacency matrix should be symmetrical. Note that **this will change the default results** but it is preferable. It is not clear to me whether other methods supported by `RunModularityClustering` take weights into account.

https://github.com/satijalab/seurat/blob/49a1be0427f2f26a531eb468ba93eeb18d8a2edb/R/clustering.R#L691

Please note that Leiden has been added to the core `igraph` library and an `r-graph` implementation in development. This implementation will be preferable for performance once it is available: https://github.com/igraph/igraph/pull/1259 https://github.com/igraph/rigraph/issues/346

satijalab:
![](https://github.trello.services/images/mini-trello-icon.png) [Move to igraph implementation of Leiden](https://trello.com/c/trV0m7fI/1-move-to-igraph-implementation-of-leiden)

TomKellyGenetics:
As discussed in ##3233, it is preferable to have handling weighted graphs be the default for `method = "igraph"`. It should also be the default for `method = "matrix"` provided that a non-binary sparse matrix is passes to `leiden` with integer weights for each non-zero edge.

rsatija:
[...]

PR #3233, opened 2020-07-06
New default handling of `weights` parameter for `RunLeiden`

andrewwbutler:
This PR addresses https://github.com/satijalab/seurat/issues/3168

When `FindClusters` is called with `algorithm = 4` (leiden algorithm) and `method = 'igraph`, the graph built with `igraph::graph_from_adjacency_matrix()` now always uses `weighted = TRUE`. This changes previous default behavior as the `weights` parameter in `FindClusters` previously determined whether this was set to `TRUE`. The `weights` parameter in `FindClusters` has correspondingly been removed from `FindClusters`. 

To maintain consistency between `method = "igraph"` and `method = "matrix"`, the `mode` parameter in `graph_from_adjacency_matrix` was left as the default (as opposed to setting to `undirected`). For some reason using `mode = "undirected"` here would sometimes give slightly different results. 

@TomKellyGenetics @therealgenna

### DIFF
--- a/R/clustering.R
+++ b/R/clustering.R
@@ -15,7 +15,7 @@ NULL
 #' @importFrom future nbrOfWorkers
 #'
 #' @param modularity.fxn Modularity function (1 = standard; 2 = alternative).
-#' @param initial.membership,weights,node.sizes Parameters to pass to the Python leidenalg function.
+#' @param initial.membership,node.sizes Parameters to pass to the Python leidenalg function.
 #' @param resolution Value of the resolution parameter, use a value above
 #' (below) 1.0 if you want to obtain a larger (smaller) number of communities.
 #' @param algorithm Algorithm for modularity optimization (1 = original Louvain
@@ -40,7 +40,6 @@ FindClusters.default <- function(
   object,
   modularity.fxn = 1,
   initial.membership = NULL,
-  weights = NULL,
   node.sizes = NULL,
   resolution = 0.8,
   method = "matrix",
@@ -87,7 +86,6 @@ FindClusters.default <- function(
             method = method,
             partition.type = "RBConfigurationVertexPartition",
             initial.membership = initial.membership,
-            weights = weights,
             node.sizes = node.sizes,
             resolution.parameter = r,
             random.seed = random.seed,
@@ -125,7 +123,6 @@ FindClusters.default <- function(
           method = method,
           partition.type = "RBConfigurationVertexPartition",
           initial.membership = initial.membership,
-          weights = weights,
           node.sizes = node.sizes,
           resolution.parameter = r,
           random.seed = random.seed,
@@ -155,7 +152,6 @@ FindClusters.Seurat <- function(
   graph.name = NULL,
   modularity.fxn = 1,
   initial.membership = NULL,
-  weights = NULL,
   node.sizes = NULL,
   resolution = 0.8,
   method = "matrix",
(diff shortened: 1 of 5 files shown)


therealgenna:
@andrewwbutler That looks good to me - thank you!

TomKellyGenetics:
Looks great @andrewwbutler, only minor point is that scripts which already use `FindClusters` with a `weights` parameter may no longer work.

I recommend deprecating the `weights` argument instead for the next Seurat release.

``` r
#define a function
my_function <- function(x, y = NULL){
  if(is.null(y)) y <- 0
  return(x+y)
}
my_function(2)
#> [1] 2
my_function(2, 3)
#> [1] 5

#remove argument y
my_function <- function(x){
  y <- 0
  return(x+y)
}
my_function(2)
#> [1] 2
my_function(2, 3)
#> Error in my_function(2, 3): unused argument (3)

#deprecate argument y
my_function <- function(x, y = NULL){
  if(!is.null(y)) {
    char <- as.character(deparse(sys.call()))
    .Deprecated(new = gsub(paste(",", y), "", char),
               package="Suerat", 
               old = char)
  }
     y <- 0
  return(x+y)
}
my_function(2)
#> [1] 2
my_function(2, 3)
#> Warning: 'my_function(2, 3)' is deprecated.
#> Use 'my_function(2)' instead.
#> See help("Deprecated") and help("Suerat-deprecated").
#> [1] 2
```

<sup>Created on 2020-07-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

R/clustering.R
```diff
@@ -747,7 +736,7 @@ RunLeiden <- function(
     object = input,
     partition_type = partition.type,
     initial_membership = initial.membership,
-    weights = weights,
+    weights = NULL,
```
TomKellyGenetics, line 739:
```suggestion
    weights = NULL,
```

Remove in next release (deprecate for now).

R/clustering.R
```diff
@@ -155,7 +152,6 @@ FindClusters.Seurat <- function(
   graph.name = NULL,
   modularity.fxn = 1,
   initial.membership = NULL,
-  weights = NULL,
```
TomKellyGenetics, line 158:
```suggestion
  weights = NULL,
```

Keep for next release and remove in following one.

R/clustering.R
```diff
@@ -720,17 +714,12 @@ RunLeiden <- function(
       },
     "igraph" = {
       input <- if (inherits(x = object, what = 'list')) {
-        if (is.null(x = weights)) {
```
TomKellyGenetics, line 723:
```suggestion 
     if(!is.null(weights)) {
          char <- as.character(deparse(sys.call()))
          .Deprecated(new = gsub(paste(",", y), "", char),
                     package="Suerat", 
                     old = char)
        }
```

Deprecate `weights` input (give warning that will be removed in future and should not be called in scripts).

andrewwbutler:
Thanks for taking the time to look this over @TomKellyGenetics!  I agree that we should introduce changes that could potentially break peoples scripts in a stepwise (deprecated -> defunct) manner. However, in this case `FindClusters` has `...` as a parameter in the function definition, meaning that even if they included `weights` in script, it should still run without an error. 

We also have an internal function `CheckDots()` that tries to verify and warn when unused parameters are getting swallowed by the `...` parameter (this was a particular concern that we tried to guard against in the transition from v2 to v3). In this scenario, the user would get the following warning:
`Warning: The following arguments are not used: weights`

We could keep the `weights` parameter in the `RunLeiden` function itself but since it is not an exported function, I thought it would be less of a concern to drop it there.

TomKellyGenetics:
PS also addresses: #2574

Issue #2949, opened 2020-05-03
Error while compiling FIt-SNE v1.1.0 or 1.0.0 on Fedora 31 5.6.8-200 solved while using with Seurat 3.1


rafalwoycicki:
Hi,
I do not know if you had a problem similar to mine, but if yes so there is a solution which I have found. I have not found a similar subject here in issues so I am posting it, if it was and I have not seen it so I apologize.

First of all, I have tried to use FIt-SNE v1.2.1 with Seurat without success, but as learned from this forum (as the wrapper is not updated to v1.2.1), I tried to install v.1.1.0 than.
My system is:
(base) [ravwoy@work-ravwoy FIt-SNE-1.0.0]$ uname -a
Linux work-ravwoy 5.6.8-200.fc31.x86_64 #1 SMP Wed Apr 29 19:10:01 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
with 96 GB RAM and 12 cores (2 x Intel Xeon each with 6 cores).

I am not a developer, nor informatician, I am biologist who learned some programming and data analysis, co Bioinformatician with moreon the Bio site, so forgiove me not being professional.

While compiling v.1.1.0 (or v.1.0.0) I got this error:
[ravwoy@work-ravwoy FIt-SNE-1.0.0]$ g++ -std=c++11 -O3  src/sptree.cpp src/tsne.cpp src/nbodyfft.cpp  -o bin/fast_tsne -pthread -lfftw3 -lm

In file included from src/tsne.cpp:42:
src/annoylib.h: In instantiation of ‘void AnnoyIndex<S, T, Distance, Random>::get_nns_by_item(S, size_t, size_t, std::vector<_Tp>*, std::vector<T>*) [with S = int; T = double; Distance = Euclidean; Random = Kiss32Random; size_t = long unsigned int]’:
src/tsne.cpp:1213:76:   required from here
src/annoylib.h:674:25: warning: taking address of packed member of ‘Minkowski::Node<int, double>’ may result in an unaligned pointer value [-Waddress-of-packed-member]
  674 |         _get_all_nns(m->v, n, search_k, result, distances);
      |                      ~~~^
src/annoylib.h: In instantiation of ‘S AnnoyIndex<S, T, Distance, Random>::_make_tree(const std::vector<_Tp>&) [with S = int; T = double; Distance = Euclidean; Random = Kiss32Random]’:
[...]

Issue #2935, opened 2020-04-30
dimensional_reduction.R fix needed for FIt-SNE 1.2.1 version 


rafalwoycicki:
Hello,

I wanted to run mca <- RunTSNE(mca, dims = 1:75, tsne.method = "FIt-SNE", nthreads = 4, max_iter = 2000) today but I am getting the info that version 1.2.1 is not compatibile with Seurat.
I have checked and dimensional_reduction.R holds check for version 1.0 and 1.1.0 and gives error with different version.
As I do not see it possible to download earlier version of FIt-SNE than 1.2.1, do you think that Seurat will work with 1.2.1? Can I somehow by myself change the code in dimensional_reduction.R so that it will accept 1.2.1 version of FIt-SNE?

Thank you .

Best regards,
Rafał

liuxl18-hku:
#2792

mojaveazure:
This has been brought to our attention. We have plans to implement support for FIt-SNE 1.2 but have other things on our plate at the moment.

rafalwoycicki:
Thanks @mojaveazure @liuxl18-hku

PR #3281, opened 2020-07-17
Support for FIt-SNE 1.2+

### LINKED ISSUE #2792: FIt-SNE wrapper requires upgrading to 1.2.0
We released Version 1.2.0 of FIt-SNE, which means that Seurat's wrapper will also need updating to use the latest executable. In the meantime, users should download the Version 1.1.0, as mentioned in the Seurat documentation. 

I should be able to update the [wrapper](https://github.com/satijalab/seurat/blob/87e2454817ed1d5d5aa2e9c949b9231f2231802f/R/dimensional_reduction.R#L1682-L1872) soon, but I wanted to submit this issue in case 1) someone encounters this error in the meantime or 2) someone else ends up updating it before I do.

saketkc:
Closes #2792
Also see #2935, #2949

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Seurat
-Version: 3.2.0.9901
-Date: 2020-07-17
+Version: 3.2.0.9002
+Date: 2020-07-27
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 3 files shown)


linqiaozhi:
This is a huge help, thanks @saketkc!

PR #3408, opened 2020-08-21
Add "do.shuffle" feature to DimPlot

samuel-marsh:
Hi Seurat Team,

This PR adds an additional logical parameter to `DimPlot` called `do.shuffle`.  This feature is based on the code/parameter of the same name from LIGER's [`plotByDatasetAndCluster` function](https://github.com/MacoskoLab/liger/blob/2532a37e0af6b5a13f821c7e8d2cdc9945fcdf9e/R/liger.R#L3040-L3044).  It makes for simple way to create a random ordering of the points in `DimPlot`.  This is especially useful when visualizing dim reduction by sample/dataset of origin and there are large number of cells and datasets. Below is example of default `DimPlot` on the left and `DimPlot` with `do.shuffle = TRUE` on the right.
do_shuffle_example

Since this is new feature I have set `do.shuffle = FALSE` as the default.

Believe everything should be in keeping with wiki guidelines for PRs but can easily change if needed.

Best,
Sam

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Seurat
-Version: 3.2.0.9014
-Date: 2020-08-19
+Version: 3.2.0.9015
+Date: 2020-08-24
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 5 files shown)


samuel-marsh:
Hi @andrewwbutler 

Looks great!!  Definitely agree moving it up makes more sense and adding seed parameter is nice.  Original liger function had ability for user set seed too but I had simplified it so as to not add to much in the PR but agree it's good to have included.

Thanks again!
Sam

Issue #3609, opened 2020-10-15
CellCycleScoring() unable to provide custom number of control genes to the call of AddModuleScore().


enblacar:
Dear Seurat developers,

I am facing an issue regarding `Seurat::CellCycleScoring` function. In the source code, this function accepts any extra parameters for `Seurat::AddModuleScore()` (since it uses the `...` notation). However, when trying to set a custom number of control genes (`ctrl` parameter), the following error is returned: 

```
Error in AddModuleScore(object = object, features = features, name = name,  : 
  formal argument "ctrl" matched by multiple actual arguments
```

This is mainly due to the fact that, in the call to `Seurat::AddModuleScore()` within `Seurat::CellCycleScoring`, the `ctrl` parameter is internally set as the minimum length between the lists of S.phase genes and G2M.phase genes:

```r
features <- list(S.Score = s.features, G2M.Score = g2m.features)
object.cc <- AddModuleScore(object = object,
                            features = features,
                            name = name,
                            ctrl = min(c(vapply(X = features, FUN = length, FUN.VALUE = numeric(length = 1)))), 
                            ...)
```

Therefore, this prevents the user from modifying this parameter. However, it might become essential to do so in the case of analyzing a low quality sample which does not have enough genes to produce 24 bins. Although the number of bins can also be modified, I find more interesting to modify the number of control genes.

As a result, unless this incapability of modifying the parameter is a desired behaviour, I would consider this a bug and the function should be modified accordingly.

The steps to reproduce the error are quite straightforward:

```r
# Given a Seurat object called "obj".
s.genes <- Seurat::cc.genes.updated.2019$s.genes
g2m.genes <- Seurat::cc.genes.updated.2019$g2m.genes

# Apply CellCycleScoring function, but providing ctrl parameter.
obj <- Seurat::CellCycleScoring(object = obj,
                                s.features = s.genes,
                                g2m.features = g2m.genes,
                                set.ident = FALSE,
                                ctrl = 75) # Or any other number, specially a low
                                           # one if we have a bad quality sample.
```

To solve this issue, I modified the function to:
- Have an extra parameter that accepts the desired number of control genes, and defaults to 100 if not provided.
[...]

PR #3613, opened 2020-10-16
update CellCycleScoring Function

samuel-marsh:
Hi Seurat Team,

This PR is in reference to issue #3609.  Checked the function and appears to be working.  Wiki doesn't specify whether you prefer users to roxygenize changes prior to submitting PR or do it on your end so I just edited the code and description.  Can easily roxygenize on my end if you prefer.

Best,
Sam

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -418,6 +418,10 @@ CaseMatch <- function(search, match) {
 #' @param object A Seurat object
 #' @param s.features A vector of features associated with S phase
 #' @param g2m.features A vector of features associated with G2M phase
+#' @param ctrl Number of control features selected from the same bin per
+#' analyzed feature supplied to \code{\link{AddModuleScore}}.
+#' Defaults to value equivalent to minimum number of features
+#' present in 's.features' and 'g2m.features'.
 #' @param set.ident If true, sets identity to phase assignments
 #' @param ... Arguments to be passed to \code{\link{AddModuleScore}}
 #' Stashes old identities in 'old.ident'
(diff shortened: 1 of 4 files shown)


samuel-marsh:
Not sure why the checks failed but appears to be unrelated to the actual code in the PR?  Thoughts?

Best,
Sam

R/utilities.R
```diff
@@ -445,16 +449,22 @@ CellCycleScoring <- function(
...
   set.ident = FALSE,
   ...
 ) {
   name <- 'Cell.Cycle'
   features <- list('S.Score' = s.features, 'G2M.Score' = g2m.features)
+  if(is.null(ctrl)) {
+    ctrl <- min(vapply(X = features, FUN = length, FUN.VALUE = numeric(length = 1)))
+  } else {
+    ctrl <- ctrl
+  }
```
mojaveazure: The `else` part of this if/else is not necessary. Also, adjust the style for proper spacing and naming of arguments.
```suggestion
  if (is.null(x = ctrl)) {
    ctrl <- min(vapply(X = features, FUN = length, FUN.VALUE = numeric(length = 1)))
  }
```

mojaveazure (changes requested):
Hi Sam,

This looks great, thank you! I've made a suggestion below regarding style, but other than that I'm happy with the fix.

samuel-marsh:
Sounds good! Just committed the suggestion.  Also do you prefer to run `roxygenize` on your end for the update to manual entry or should I run and add commit?

Thanks!
Sam

mojaveazure:
I've taken care of updating the help pages. We import documentation from packages not automatically installed when installing Seurat. Rather than having you cherry pick which aspects of the documentation to commit, I've handled it just to ensure that no bits of the documentation get lost. This will change in the future as we move away from this fragile design.

mojaveazure (approved):

mojaveazure:
Thanks for taking care of this, Sam!

samuel-marsh:
No problem, happy to help!!

PR #3658, opened 2020-10-28
fix RenameCells bug

helenmiller16:
Hello! I came across a bug in RenameCells, for datasets that have gone through batch integration via the SCTransform route, per https://satijalab.org/seurat/v3.0/integration.html. 

It uses `cells_step1` to pull the relevant cells, but this will fail when SCTransform was run with ncells is less than the number of cells in that batch. 

I started writing up an issue but it seemed easier to just submit a PR since it is just one line of code. I'm happy to submit an issue or provide more info if needed. Thanks!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Seurat
-Version: 3.2.2.9008
-Date: 2020-10-27
+Version: 3.2.2.9009
+Date: 2020-11-03
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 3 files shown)


andrewwbutler:
Hi,

Thanks very much for the fix! I modified it slightly to preserve the list of cells in `cells_step1`. Will get it merged in if the modifications look good to you.

helenmiller16:
Looks good to me, thanks!

PR #3682, opened 2020-11-04
Fix highlight order depending on alpha order

JZL:
Hi,

I can also submit this separately as an issue but I think I traced down the issue.

When I followed the spatial vignette for my cell type classifications, I noticed that some of the visualizations were inverted (red where the cell type was present and grey where it was not). I attached a minimally working example notebook \[3\] but, to summarize, as far as I can tell, it depends on the alphabetical order of the class name compared to the internal `Unselected` class name. So if I create two identical identity classes with names `V_true` and `UA_True`, then `V_true`  returns the inverted version [1] but `UA_True` returns the correct version [2].

Tracing it through, the code goes out of its way to avoid this but when it `factor`izes the `highlight` variable in `SetHighlight`, it doesn't set the level order and this order is used by `ggplot` later

Happy to work with you on debugging further!

\[1\]: 

image

\[2\]: 

image

\[3\]:
[Seurat_Dimplot_alpha_inversion.zip](https://github.com/satijalab/seurat/files/5490399/Seurat_Dimplot_alpha_inversion.zip)

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -4811,7 +4811,7 @@ SetHighlight <- function(
   plot.order <- sort(x = unique(x = highlight), na.last = TRUE)
   plot.order[is.na(x = plot.order)] <- 'Unselected'
   highlight[is.na(x = highlight)] <- 'Unselected'
-  highlight <- as.factor(x = highlight)
+  highlight <- factor(x = highlight, levels = plot.order)
   return(list(
     plot.order = plot.order,
     highlight = highlight,


JZL:
I noticed the CI checks failed and with a quick check, seems like it failed on `SCTransform` related tests which is a little strange because my change should (?) only affect plotting. Maybe it's because I pushed to the `spatial` branch and there is some difference in CI tests passing between branches.

Any pointers to where I should look to fix this would be great (I'm also happy to create a new PR on a different branch)

andrewwbutler:
Hi @JZL ,

Thanks for the fix and sorry for the branch confusion. The `spatial` branch is old at this point and should probably be removed. I made a new PR with the same change (https://github.com/satijalab/seurat/pull/3694) but just based of the `develop` branch instead. This should resolve the CI issues and we can get the fix merged in.

PR #3734, opened 2020-11-16
Revised CellRanger input versioning logic

### LINKED ISSUE #3653: Small change in `Read10X_h5` related to CellRanger versioning logic
I would love to change the line 
https://github.com/satijalab/seurat/blob/fe93b05745e55ec2f66e3f0b4c4196aad9f4d5a7/R/preprocessing.R#L1155

to
```R
    if (hdf5r::existsGroup(infile, 'matrix')) {
```

**Why?**

I think this logic for determining CellRanger h5 version is more robust.  I also develop a tool that produces CellRanger-format h5 files as outputs (https://github.com/broadinstitute/CellBender), which are to be loaded in Seurat.  However, since my tool is written in python and saves files using PyTables, the files have a `PYTABLES_FORMAT_VERSION` attribute regardless of whether the format is v2 or v3.  This change would allow Seurat to load these files seamlessly, and the change should not have any drawbacks.

Please let me know if I can / should make a pull request, and how you like to handle code contributions.

sjfleming:
Resolves #3653

This small change improves the robustness of the logic for determining input CellRanger h5 version.  Outside software tools that are written in python and save files in CellRanger format using PyTables have a `PYTABLES_FORMAT_VERSION` attribute regardless of whether the format is v2 or v3.  This change makes the v2/v3 decision based on the presence of `'matrix'` in the h5 file, which is a major organizational change between CellRanger v2 and v3 files.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,4 +1,4 @@
-oPackage: Seurat
+Package: Seurat
 Version: 3.2.2.9012
 Date: 2020-11-17
 Title: Tools for Single Cell Genomics
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
-Package: Seurat
-Version: 3.2.2.9011
-Date: 2020-11-09
+oPackage: Seurat
+Version: 3.2.2.9012
+Date: 2020-11-17
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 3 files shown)


andrewwbutler:
Thanks @sjfleming!

sjfleming:
Thank you!

PR #3748, opened 2020-11-18
Add keep.scale param to FeaturePlot when creating split plots

samuel-marsh:
Hi Seurat Team,

This is PR to add new param to `FeaturePlot`.  The current behavior of `FeaturePlot` when plotting a split feature plot is to create a gradient scale for each of the plots independently.  However, this can be deceiving to user if the legend is not plotted as the same color on the different plots can represent incredibly different values.
Example of current behavior with following code:
```
FeaturePlot(object = obj, features = "Gene", split.by = "Meta_Name", order = T) & theme(legend.position = "right")
```
OLD_split

This PR adds an additional parameter `keep.scale`.  This rescales the gradient to be constant across all plots.
Example:
```
FeaturePlot(object = obj, features = "Gene", split.by = "Meta_Name", keep_scale = T, order = T) & theme(legend.position = "right")
```
keep-scale_split

Couple of notes about implementation.
1.  There may be better way to implement this behavior than I have done so open to changes if it fits better with intended function behavior.

2.  While I would normally try and preserve the current implementation of figure when adding new parameters, in this case I think a default value of `keep.scale = TRUE` is optimal even though is contrary to current function behavior.  My reasoning was two-fold. One, as the function is currently implemented the lack of legends when plotting a `split.by` plot may lead user to believe that both plots are on the same scale.  Two, the potential for confusion is potentially confounded by the patchwork  vs. ggplot2 modification syntax.  If user was to simply add `FeaturePlot(...) + theme(legend.position = "right")` instead of `FeaturePlot(...) & theme(legend.position = "right")` they would end up with legend they may think applies to all plots but in fact only applies to final plot in the series.  Of course, final choice up to you but just my thoughts.

_\*Edit\* Gene name and split.by titles were removed post from these plots.  They will appear when code is run._

Thanks very much as ever for all the work you do and excellent package!!
Best,
Sam

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Seurat
-Version: 3.2.2.9012
-Date: 2020-11-17
+Version: 3.2.2.9013
+Date: 2020-12-04
 Title: Tools for Single Cell Genomics
 Description: A toolkit for quality control, analysis, and exploration of single cell RNA sequencing data. 'Seurat' aims to enable users to identify and interpret sources of heterogeneity from single cell transcriptomic measurements, and to integrate diverse types of single cell data. See Satija R, Farrell J, Gennert D, et al (2015) <doi:10.1038/nbt.3192>, Macosko E, Basu A, Satija R, et al (2015) <doi:10.1016/j.cell.2015.05.002>, and Stuart T, Butler A, et al (2019) <doi:10.1016/j.cell.2019.05.031> for more details.
 Authors@R: c(
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 ### Added
 - Titles added to `DimPlot` when specifying `group.by` parameter
+- `keep.scale` parameter added to `FeaturePlot` to control scaling across multiple features and/or splits.
 
 ### Changes
 - `Same` deprecated in favor of `base::identity`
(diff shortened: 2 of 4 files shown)


andrewwbutler:
Hi Sam,

Thanks for the PR! I definitely agree that this is a useful parameter to add. The situation gets a little more complicated though when you're plotting multiple features and also using `split.by`. In this case, would we want to preserve the scale across all plots or just across plots for each feature (row)? In order to have this added granularity, one alternative implementation could be to change `keep.scale` from a boolean to accepting one of:
* `all` - same scale for every plot returned
* `feature`- same scale for each plot of the same feature
* `NULL` - each plot scaled individually

In this case, I think `feature` might be a sensible default. We could also extend the functionality here when passing multiple features without split.by, to more easily facilitate plotting multiple feature plots with the same scale.

samuel-marsh:
Hi Andrew,

I like that idea, I'll work on changing PR to incorporate that.  I think you're right that setting it to feature makes the most sense.     

In my mind we don't necessary need to apply this functionality to the plotting of multiple features without `split.by`.  When you plot multiple features without `split.by` because by default legends are plotted with all of those features so user can interpret the difference in scale and often on per feature basis it is more informative to visualize the min/max per feature and could make plotting of multiple features with extremely disparate levels of expression confusing.  That said if you and team think it's worthwhile I can incorporate that too.

Thanks!
Sam

samuel-marsh:
Hi Andrew,

Please see new commit with new parameter behavior as we discussed.  The implementation seems to work fine in my environment but give it go and make sure it behaves correctly for you as well in regards to "feature" vs "all" vs NULL.  I also updated the documentation with choices and description for each option.

Let me know what you think and if there is anything else that I should incorporate.

Thanks!
Sam

andrewwbutler:
Thanks for implementing the changes! I agree that most of the time plotting multiple features without `split.by` should be on the different scales but we've had the feature requested occasionally and it was easy enough to implement here so I went ahead and enabled that as well (not default but will work now if you specify `"all"`).

PR #3876, opened 2020-12-31
Added flexibility of choice for cell column. Preparing compatibility …

Hoohm:
…with CITE-seq-Count 1.5.0

Hello!

I'm about to release a new version of CITE-seq-Count (1.5.0) with a lot of new changes and performance increases.

One feature is the ability to use the "mapping" barcode file from 10X between the mRNA and the Protein cell barcode in 10xV3 chemistries.

I wanted to keep the original barcode in the barcodes.tsv file for traceability.
The first column is the mapped cell barcode to the mRNA cell barcode, the second is the cell barcode from the Protein data.

On top of that, the unmapped feature is removed form the UMI count and the features.tsv now hold two columns. First being the sequence of the protein TAG, the second being the feature_name. These changes reduce overhead for Seurat users when reading CSC data using the Read10X function.

All those changes should make reading CSC outputs easier and cleaner.

Hope all of you are well!

Wish you a happy new year!


andrewwbutler:
Hey Patrick! 

Hope all is well with you too. Thanks for the updates and the PR!

Issue #3897, opened 2021-01-08
DimPlot() changes point shape for more than 50 000 points. 


krejciadam:
Hi!
  I noticed weird behavior of DimPlot() today: 

When plotting more than 50 000 cells, the default point shape changes from circle to cross. To me, this looks quite ugly. What is more important, it is not in line with the function's documentation and I think it can be quite confusing. Moreover, `shape.by` stops working for more than 50 000 cells and everything is always plotted as cross.

`DimPlot(cells[, c(1:50000)])`
![point](https://user-images.githubusercontent.com/10741857/104006500-dc74df80-51a6-11eb-85da-05a3ee2bcc2d.png)
`DimPlot(cells[, c(1:50001)])`
![cross1](https://user-images.githubusercontent.com/10741857/104006518-e26ac080-51a6-11eb-8b2d-839da9126dc5.png)
`DimPlot(cells[, c(1:50001)], pt.size = 0.5)`
![cross2](https://user-images.githubusercontent.com/10741857/104006542-e991ce80-51a6-11eb-963b-7133921e1e9d.png)
`DimPlot(cells[, c(1:50000)], pt.size = 0.5, shape.by = "Phase", group.by = "Phase")`
![shape1](https://user-images.githubusercontent.com/10741857/104007700-a59fc900-51a8-11eb-980e-7092240e16a7.png)
`DimPlot(cells[, c(1:50001)], pt.size =  0.5, shape.by = "Phase", group.by = "Phase")`
![shape2](https://user-images.githubusercontent.com/10741857/104007723-acc6d700-51a8-11eb-8602-1591e5e16753.png)

I'm not sure whether this started happening after I updated all Bioconductor packages to most current versions (Bioconductor 3.12.) or it is connected to Seurat 3.2.3. This is my `SessionInfo()`

Thanks! 
Adam

```
R version 4.0.3 (2020-10-10)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Linux Mint 19

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/libopenblasp-r0.2.20.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
[...]

PR #3898, opened 2021-01-08
Update DimPlot Manual Entry

samuel-marsh:
Hi Seurat Team,

Just quick PR based on #3897 to update the manual entry for `shape.by` in `DimPlot` dependence on value of `raster`.

Best,
Sam


timoast:
Thanks for the PR Sam!

samuel-marsh:
No problem!

PR #3902, opened 2021-01-08
Add raster warning message

samuel-marsh:
Hi Seurat Team,

Was just thinking more on PR #3898 that I pushed earlier based on #3897.  I thought it might be helpful to have one time warning about the plots being rasterized if greater than 50,000 points present.  I've never written warnings like this based on option settings but I think based on looking at rest of Seurat codebase I've got it right, but please feel free to correct!

I added separate warnings to both `SingleDimPlot` and `SingleCorPlot` as those are the functions effected.

Tagging @timoast here because he merged my earlier PR today.

Thanks again!
Sam


saketkc:
Thanks @samuel-marsh, I will close this and make one based on some of your changes to `release/4.0.0`. We agree, that a message would be good. We have also decided to up the threshold to 100k cells. 

Thanks again!

Issue #4215, opened 2021-03-12
Argument specification for Read10X is not optimal


Alanocallaghan:
Having `data.dir` specified as `NULL` implies that it's an optional argument. 
https://github.com/satijalab/seurat/blob/29e7eb7ba08519e34406ed03d323dec8787ff225/R/preprocessing.R#L773-L781

You also get this useless error message because it's trying to execute `length(NULL[[1]])` here:
https://github.com/satijalab/seurat/blob/29e7eb7ba08519e34406ed03d323dec8787ff225/R/preprocessing.R#L883

``` r
library("Seurat")
#> Attaching SeuratObject
Read10X()
#> Error in full.data[[1]]: subscript out of bounds
```

If you just don't specify the argument you get something approaching a sensible error without most of the function body being used at all; it just errors when trying to evaluate an argument you've not supplied.

https://github.com/Alanocallaghan/seurat/blob/b4d935584357f64c628ece2ac3ff1c9d748456a0/R/preprocessing.R#L773-L781

``` r
library("Seurat")
#> Attaching SeuratObject
Read10X()
#> Error in Read10X(): argument "data.dir" is missing, with no default
```

Generally good practice is to not specify mandatory arguments so you get this sensible error message. Optional arguments can be `NULL` if you can later figure out a sensible value for them based on the other argument values.

andrewwbutler:
Closing as the proposed PR has now been merged

PR #4216, opened 2021-03-12
Dont specify mandatory argument as NULL in Read10X

Alanocallaghan:
See  #4215


andrewwbutler:
Thanks for the suggestion! Seems like a good change to me. Could you make the change on a branch based off of `develop` instead of `master`? Just makes housekeeping a little easier in terms of versioning/updating the NEWS file.

Alanocallaghan:
I merged your dev branch into my master

Issue #3622, opened 2020-10-20
Warning message due to "future" package update


ktrns:
I am running my `Seurat` workflow in combination with the `future` package. 

I recently updated to the latest package versions: 
```
> packageVersion("Seurat")
[1] ‘3.9.9.9002’
> packageVersion("future")
[1] ‘1.19.1’ 
```

I am getting the following warning multiple times, e.g. when calling `Seurat::NormalizeData`: 
```
UNRELIABLE VALUE: Future (‘future_lapply-1’) unexpectedly generated random numbers without specifying argument '[future.]seed'. There is a risk that those random numbers are not statistically sound and the overall results might be invalid. To fix this, specify argument '[future.]seed', e.g. 'seed=TRUE'. This ensures that proper, parallel-safe random numbers are produced via the L'Ecuyer-CMRG method. To disable this check, use [future].seed=NULL, or set option 'future.rng.onMisuse' to "ignore".
```

I found [this](https://www.r-bloggers.com/2020/09/future-1-19-1-making-sure-proper-random-numbers-are-produced-in-parallel-processing/) link to be useful. I think what you have to do is to set `future.seed = TRUE` when using `future_lapply` in your code. 

I am now ignoring the messages by `options(future.rng.onMisuse="ignore")`. 

Hope this helps
Katrin

BenvanderVeer:
I am having the same issue - tried adding seed = TRUE to the options function in future, but didn't work. Waiting for an update of seurat.

best, Ben

andrewwbutler:
Hi,

Thanks for reporting. We've pushed a fix for these warnings so updating to the latest `release/4.0.0` (3.9.9.9008) should remove them. Either way though, they can be safely ignored in this context as no random seed is being used in that function.

PR #4260, opened 2021-03-20
future.seed=TRUE set to avoid warning and create reliable random #s

kana4:
This is just a one liner that fixes [this warning](https://github.com/satijalab/seurat/issues/3622)


andrewwbutler:
Hi,

Can you provide a reproducible example that exhibits those warnings? In general, for functions that don't rely on a random seed (e.g. `NormalizeData()`), this warning can be avoided by setting `rng=false` in the `exports()` tag of the relevant Rcpp function.

andrewwbutler:
Closing as we believe the issue (https://github.com/satijalab/seurat/issues/3622) has been resolved.

PR #4268, opened 2021-03-22
Fix score.thresh parameter in ScoreJackStraw.Seurat 

colganwi:
Currently the score.thresh parameter in ScoreJackStraw.Seurat doesn't do anything because it is hardcoded to 1e-5 in ScoreJackStraw.DimReduc. This patch fixes this.


andrewwbutler:
Thanks for the fix!

Issue #4531, opened 2021-05-25
as.SingleCellExperiment Fails in Seurat_4.0.2


bbimber:
Hello,

I'm seeing one warning and two errors in R 4.1, Seurat_4.0.2, and SeuratObject_4.0.1. The most severe is w/ as.SingleCellExperiment. Here is a repro

```

# Create dummy object:
mat <- matrix(1:100, ncol = 10)
colnames(mat) <- LETTERS[1:10]
rownames(mat) <- LETTERS[1:10]
  
# this works, but prints the warning "Warning: The following arguments are not used: row.names"
# I pointed this issue out with code pointers here: https://github.com/mojaveazure/seurat-object/issues/13
seuratObj <- Seurat::CreateSeuratObject(mat)
  
```

and then try this:

```
  sce <- as.SingleCellExperiment(seuratObj)

```
which gives:

Error in value[[3L]](cond) : 
  invalid subscript 'e' in 'altExp(<SingleCellExperiment>, type="character", ...)':
  'RNA' not in 'altExpNames(<SingleCellExperiment>)'

I assume this has to do w/ trying to make as.SingleCellExperiment work with multiple assays in the object? The problem is here:

```
for (assayn in assay){
	# This fails if you try to swap the active experiment. 
	sce <- SingleCellExperiment::swapAltExp(x = sce, name = assayn, saved = orig.exp.name)
	SummarizedExperiment::rowData(x = sce) <- S4Vectors::DataFrame(x[[assayn]][[]])
	sce <- SingleCellExperiment::swapAltExp(x = sce, name = orig.exp.name, saved = assayn)
}

```

PR #4532, opened 2021-05-25
Bugfix to as.SingleCellExperiment

bbimber:
This is related to the issue I opened here:  https://github.com/satijalab/seurat/issues/4531, which includes a simple repro case.

Whenever I run as.SingleCellExperiment(), it gives errors like this:

Error in value[3L] :
invalid subscript 'e' in 'altExp(, type="character", ...)':
'RNA' not in 'altExpNames()'

I believe the issue is that the 4.0.2 version of as.SingleCellExperiment tries to call SingleCellExperiment::swapAltExp() when passing the name of the currently active experiment. 

I think that worked in as.Seurat.SingleCellExperiment(), since you inject "originalexp" into the object, so iterating the true assay names works.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Seurat
-Version: 4.0.2
-Date: 2021-05-20
+Version: 4.0.2.9001
+Date: 2021-05-28
 Title: Tools for Single Cell Genomics
 Description: A toolkit for quality control, analysis, and exploration of single cell RNA sequencing data. 'Seurat' aims to enable users to identify and interpret sources of heterogeneity from single cell transcriptomic measurements, and to integrate diverse types of single cell data. See Satija R, Farrell J, Gennert D, et al (2015) <doi:10.1038/nbt.3192>, Macosko E, Basu A, Satija R, et al (2015) <doi:10.1016/j.cell.2015.05.002>, Stuart T, Butler A, et al (2019) <doi:10.1016/j.cell.2019.05.031>, and Hao, Hao, et al (2020) <doi:10.1101/2020.10.12.335331> for more details.
 Authors@R: c(
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,10 @@
+## Seurat develop
+
+## Added
+
+## Changes
+- Fix issues with `as.SingleCellExperiment.Seurat` for the latest verion of SingleCellExperiment ([#4532](https://github.com/satijalab/seurat/pull/4532))
+
 # Seurat 4.0.2 (2020-03-20)
 ## Added
 - New `AddAzimuthScores()` and `AddAzimuthResults()` functions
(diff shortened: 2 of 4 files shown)


bbimber:
Note: this adds a test case to illustrate the failure

mojaveazure:
Thanks for taking care of this! The issue is not entirely with Seurat v4.0.2, but also with SingleCellExperiment 1.14.1 (Bioc 3.13) changing how they handle the primary altExp. Prior versions of SingleCellExperiment had all alternate experiments stored together as altExps, but the latest version moves the primary altExp to mainExp

I've made some modifications to your tests:

1. I've set the test to skip on CRAN since CRAN doesn't always have Bioc packages installed or have weird versions installed (eg. Bioc devel) on their testing servers and we don't want to be bogged down by missing/incorrect Bioc packages.
2. I've also put a check in the tests to only run the test when SingleCellExperiment is installed: SingleCellExperiment is a suggested package, which means it's not always installed when Seurat is; we don't want our tests to fail on a missing suggested package.
3. Finally, I've commented out the tests that are specific to SingleCellExperiment 1.14.1; the latest version of SingleCellExperiment is specific to R 4.1 (Bioc 3.13 is specific to R 4.1), and since we still support R 4.0 we don't want our tests to fail because of SingleCellExperiment's requirements. We'll change our own requirements to stay in line with our dependencies (any package listed in `Depends` or `Imports` in `DESCRIPTION`), but not suggested packages.

bbimber:
@mojaveazure OK, thanks. what's your release schedule for bugfixes like this?

mojaveazure:
Once the CI passes, we'll get this merged into the `develop` branch. However, we cannot submit to CRAN right away as their policies require a minimum of two weeks between submissions, and they prefer a month (limited server capacity, especially for large packages such as Seurat). Given that Seurat v4.0.2 was just released last week, it will likely be another 2-3 weeks before we start considering another release.

bbimber:
@mojaveazure OK. While I have you, have you noticed this new warning (which seems like a simple bug): https://github.com/mojaveazure/seurat-object/issues/13

mojaveazure:
The warning `Warning: The following arguments are not used: row.names` is expected. When we moved from Seurat v2 to v3, a lot of functions had very slight parameter changes. The `...` were introduced to many functions since we moved from a pure functions generics/methods, allowing us to introduce new specialized routines for assays vs dimensional reductions, as well as specialized assays (eg. `SCTAssay`, `Signac::ChromatinAssay`).

Feedback we got from this switch was that when users passed a v2 parameter to a v3 function, that parameter was silently ignored and the v3 default was silently used instead. In order to make it clear when a v2 parameter was ignored, we introduced `CheckDots` to warn when a parameter was passed to a function via `...`, but not used in a function. This helped users transition from the v2 parameters to the v3 parameters.

The reason this specific warning is happening is that we introduced the ability to use `data.frame`-derivatives that don't have row names (eg. `tibble`, `data.table`) when creating Seurat objects. As Seurat requires all assay slots to be a `matrix` or `dgCMatrix`, we convert any non `matrix`/`dgCMatrix` objects to a `dgCMatrix`. This doesn't work for `tibble`/`data.table` as they lose feature names since they don't have rownames. As such, `as.sparse.data.frame` gained a `row.names` parameter to allow users to pass feature names to `CreateSeuratObject` (which eventually calls `as.sparse`) to ensure feature names are preserved. Since this argument is only used for the `data.frame` method, all other `as.sparse` methods throw this warning. We are looking into finding a way to ignore this specific warning, haven't been able to do so yet.

Issue #4378, opened 2021-04-16
Implement DensMAP


lldelisle:
Hi,
I was wondering if you were planing to implement DensMAP as a parameter for UMAP from sklearn:
https://umap-learn.readthedocs.io/en/latest/densmap_demo.html
I can also write a PR if nobody is already doing it on your side.

andrewwbutler:
Hi @lldelisle ,

We'd be happy to incorporate a PR with this functionality if you have the bandwidth to put one together.

erosix:
I would really like to have densMAP for Seurat. *.* Any way this can be implemented soon?

lldelisle:
Hi,
I just wrote #4630 ...

andrewwbutler:
This functionality is now available on the `develop` branch. Please see installation instructions [here](https://satijalab.org/seurat/articles/install.html#install-the-development-version-of-seurat-1)

PR #4630, opened 2021-06-15
support densmap

lldelisle:
Following #4378 . Support densmap for the umap with umap-learn.
I tested it on the PBMC dataset.
I am not sure with the warning/stop decision.


andrewwbutler:
Thanks @lldelisle ! Looks good to me.

PR #4641, opened 2021-06-16
Allow user to specify images to Load10X_Spatial

royfrancis:
- `Load10X_Spatial()` add a new parameter `image` to allow user to pass in an image.
- `Read10X_Image()` adds a new parameter `image.name` to allow user to specify a custom image other than _tissue_lowres_image.png_.
- Practically the idea is to allow user to read in high resolution slide image if needed.


andrewwbutler:
Thanks for the PR! I agree that the added flexibility can be useful here.

PR #4662, opened 2021-06-20
Performance improvement of WilcoxDETest

mogushi:
Hi Seurat Team,

This is a small patch to improve the performance of WilcoxDETest.
When a sparse matrix is converted into a plain matrix prior to Wilcoxon rank-sum test, the execution time can be reduced.
For larger dataset (e.g. 10,000 cells), the execution time is typically reduced by 20-25%.
It also works good with parallel execution by future library using plan(strategy = multiprocess(workers = n)) .

I performed a simple validation using the pbmc3k dataset, and obtained top 6 p-values were the exactly the same as the following URL:
  https://satijalab.org/seurat/articles/de_vignette.html

Similar conversion from a sparse matrix to a plain matrix prior to my.sapply() may be effective in other DE tests such as t-test.

Mogushi

### DIFF
--- a/R/differential_expression.R
+++ b/R/differential_expression.R
@@ -1987,25 +1987,7 @@ WilcoxDETest <- function(
   my.sapply <- ifelse(
     test = verbose && nbrOfWorkers() == 1,
     yes = pbsapply,
-    no = function(X, FUN) {
-      gc(reset = TRUE)
-
-      # heuristic parameter for automatic estimation of batch size;
-      # for example, 100 genes for each iteration with 10,000-cell data
-      batch_size <- 1e6 / ncol(data.use)
-      index_list <- split(X, ceiling(X / (batch_size * nbrOfWorkers())))
-
-      if (length(index_list) == 1) {
-        p_val <- future_sapply(X = index_list[[1]], FUN = FUN)
-      } else {
-        p_val <- sapply(index_list,
-          function(z) {
-            future_sapply(X = z, FUN = FUN)
-          }
-        )
-      }
-      unlist(p_val)
-    }
+    no = future_sapply
   )
   overflow.check <- ifelse(
     test = is.na(x = suppressWarnings(length(x = data.use[1, ]) * length(x = data.use[1, ]))),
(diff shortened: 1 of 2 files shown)


mogushi:
Line 1985: 
This modification improves performance. I performed [some benchmark tests and verification](https://github.com/mogushi/seurat) using the pbmc3k.final dataset.
```
-  data.use <- data.use[, c(cells.1, cells.2), drop = FALSE]
+  data.use <- as.matrix(x = data.use[, c(cells.1, cells.2), drop = FALSE])
```

Line 1990:  (experimental) reduction of memory usage of future_saaply; not for performance improvement

saketkc:
Thanks @mogushi! I am going to take a look at this.

mogushi:
Thanks @saketkc for your attention.
As discussed in issue #4689, the modification in Line 1985 is a solution for the second question.

mogushi:
Lines 2038-2047: 
Under the circumstances when limma cannot be used, replacing wilcox.test(formula) with wilcox.test(x, y) will improve performance.
wilcox.test(x, y) is about 10% slower than imma::rankSumTestWithCorrelation, but it is much faster than wilcox.test(formula). 

Line 1990:
 To focus on discussion about performance improvement of Wilcoxon rank-sum test, an experimental patch for memory reduction was removed.

saketkc:
Hi @mogushi, thanks for your contribution and for putting in so much effort with the benchmarking. Your observations are correct: making a dense matrix can lead to speedups in almost all cases, however an important caveat is that this operation will fill the memory too quickly for large datasets.

We decided to support densification of matrix before doing DE by proving an extra argument `densify=TRUE` instead of 'densifying' by default. We hope you will find this useful. Please do let me know if you have any follow up suggestions. You can try this version from the develop branch with [these](https://satijalab.org/seurat/articles/install.html) instructions.  

Thanks once again for the pointers!

mogushi:
Thanks @saketkc for your careful consideration.
`densify = TRUE` option would be helpful for speeding up analysis of relatively small dataset.

PR #4719, opened 2021-07-02
fix typos in two preprocessing error messages

AustinHartman:


andrewwbutler:
Thanks for the typo fix!

PR #4780, opened 2021-07-15
Add pct.diff parameter to DE results dataframe

samuel-marsh:
Hi Seurat Team,

Just quick PR and not 100% sure changes are in right place so feel free to edit.  Just PR to add a percent difference column in addition to pct.1 and pct.2 to the results from `FindMarkers`/`FindAllMarkers`.  While it's easy to calculate and add post-hoc thought it might be nice to have included by default as it provides nice quick idea of the basic magnitude and direction of change in percent expressing.

Thanks as always!
Sam

### DIFF
--- a/R/differential_expression.R
+++ b/R/differential_expression.R
@@ -892,12 +892,13 @@ FoldChange.default <- function(
       length(x = cells.2),
     digits = 3
   )
+  pct.diff <- pct.1 - pct.2
   # Calculate fold change
   data.1 <- mean.fxn(object[features, cells.1, drop = FALSE])
   data.2 <- mean.fxn(object[features, cells.2, drop = FALSE])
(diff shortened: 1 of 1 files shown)


samuel-marsh:
Looks like the check failure might be due to the addition of extra column?

timoast:
Hi Sam, thanks for the PR. After some internal discussion, we're not convinced it's really necessary to add this given how easy it is for users to add it themselves. We're trying to be conscious of users who expect a certain output format/dimension from `FindMarkers` and we'd like to minimize unnecessary changes that could break pipelines for people, especially outside of major Seurat version changes.

samuel-marsh:
No worries!  I kinda figured that might be the case given that it was already possible to filter results on a pct.diff but it wasn't explicitly returned.  

Best,
Sam

PR #5076, opened 2021-09-13
add violin raster

samuel-marsh:
Hi Seurat Team,

Just a quick PR if you guys like it, that adds raster plotting to VlnPlots.  Given individual data points can still be helpful but like other plots rendering that many points in vector form becomes inefficient in multiple ways.

I used ggrastr and added that to package imports page as well.

Thanks,
Sam


saketkc:
Thanks Sam! Do you have an example I could look at? Just want to avoid adding another dependency if possible.

saketkc:
Thanks very much! I will take a closer look and also discuss it internally.

samuel-marsh:
Hi @saketkc makes sense and good idea to make optional.  One change is that I believe ggrastr should be moved to the Suggests section instead of Imports in DESCRIPTION file as I believe anything in Imports section will be installed when installing Seurat which doesn't make it optional anymore.

Thanks very much!

saketkc:
Thanks for checking - I did remove it from imports, but did not add it to suggests, which I have fixed.

saketkc:
Thanks a lot for your contribution, Sam!

Issue #5388, opened 2021-12-15
Setting higher dpi/resolution in rasterized plots


rdchow:
I really like the new raster=TRUE setting in DimPlot/FeaturePlot, but many of my colleagues are wondering if we could increase the resolution/dpi of the resulting rasterized points. 

My understanding is that DimPlot uses the geom_scattermore() function, which has a default pixel setting of (512,512). Could you implement another setting into DimPlot that allows us to set the pixel dimensions to something higher, which gets passed to geom_scattermore()? Thanks!

samuel-marsh:
@rdchow I think this makes lot of sense!!  I just submitted PR to the devs that would add this parameter where needed #5391.

Best,
Sam

saketkc:
This should be addressed by #5392  and is now in the develop branch. You can see the installation instructions [here](https://satijalab.org/seurat/articles/install.html)

PR #5391, opened 2021-12-15
add raster.dpi parameter scattermore-based plots

samuel-marsh:
Hi Seurat Team,

Just quick PR supporting what I think seems like good feature request in #5388.  This adds relevant new parameter `raster.dpi` to relevant plots that use `geom_scattermore`.  I have also added appropriate documentation and a warning check to make sure that supplied values for raster.dpi are valid.  Pretty sure I caught all the spots (`SingleCorPlot` and `SingleDimPlot` and their downstream functions).

Similar to `raster` I set parameter to NULL in the base plotting functions and set default value in downstream functions that is equivalent to current scattermore defaults.  Alternatively could change it so that parameter is NULL in downstream functions and if left NULL it will default to `c(512, 512)` if you think that's better on user end.

Thanks again!
Sam

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -752,6 +752,8 @@ ColorDimSplit <- function(
 #' ggplot object. If \code{FALSE}, return a list of ggplot objects
 #' @param raster Convert points to raster format, default is \code{NULL} which
 #' automatically rasterizes if plotting more than 100,000 cells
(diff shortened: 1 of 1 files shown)


samuel-marsh:
Apologies looks like I create branch where there are conflicts with more recent develop branch changes.  I'll leave for now in case you prefer to resolve conflicts yourselves but also happy to go through and resolve (leave all changes present) if that's easier for you.

Best,
Sam

PR #5581, opened 2022-02-04
Add newlines to end of object definitions

george-hall-ucl:
This is a very minor commit to add newline characters to the end of some object definitions.

Previously, if we inspected an AnchorSet object in the console then we got:

An AnchorSet object containing <some number> anchors between the reference and query Seurat objects. 
 This can be used as input to TransferData.> 

i.e. the prompt is on the same line as the last line of the message. I presume that this is not deliberate? If it was deliberate then sorry for the pointless change! I made this change for objects with the TransferAnchorSet, IntegrationAnchorSet, ModalityWeights, and SCTModel signatures.

Thanks,
George

# Note


AustinHartman:
thanks!

PR #5615, opened 2022-02-10
Add misc removal to DietSeurat

samuel-marsh:
Hi Seurat Team,

So I realize this is edge case but I think it's edge case with little downside to add.  This is PR to add parameter to remove the `@misc` slot when using `DietSeurat`  since the misc slot can store anything it can become quite large in size and users may want to remove easily when using `DietSeurat` to reduce object size.  I set the default to TRUE because like I said this is probably edge case where users would want to remove it so under normal circumstances it wouldn't be removed.

Thanks,
Sam

### DIFF
--- a/R/objects.R
+++ b/R/objects.R
@@ -454,6 +454,7 @@ CreateSCTAssayObject <- function(
 #' remove all DimReducs)
 #' @param graphs Only keep a subset of Graphs specified here (if NULL, remove
 #' all Graphs)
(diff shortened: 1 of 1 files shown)


lianov:
Thanks for this PR Sam. As a fellow user of Seurat I can certainly see the benefit of this to some use-cases I have come across as well. So just adding my +1 here.

AustinHartman:
thanks!

PR #5746, opened 2022-03-16
Update preprocessing.R

duocang:
After rename seurat cells' name, colnames(obj $RAN@counts) returens a named vector.

> head(colnames(obj@assays$RNA@counts), 2)
AAACCCACAACTAGAA_1 AAACCCACAAGGGTCA_1
"AAACCCACAACTAGAA-col0" "AAACCCACAAGGGTCA-col0"

SCTransform will return an error because of the following code.

```r
# Make sure rownames of cell attributes match cell names in count matrix
if (!identical(rownames(cell_attr), colnames(umi))) {
  stop('cell attribute row names must match column names of count matrix')
}
```

### DIFF
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -1558,7 +1558,7 @@ SCTransform <- function(
   assay <- assay %||% DefaultAssay(object = object)
   assay.obj <- GetAssay(object = object, assay = assay)
   umi <- GetAssayData(object = assay.obj, slot = 'counts')
-  cell.attr <- slot(object = object, name = 'meta.data')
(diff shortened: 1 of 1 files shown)


moi-taiga:
I have used the below as a personal workarround, would it be suitable as a fix for this?
```r
# Make sure rownames of cell attributes match cell names in count matrix
if (!setequal(rownames(cell_attr), colnames(umi))) {
    stop('cell attribute row names must match column names of count matrix')
  }
```

dcollins15:
Thanks for using Seurat!

PR #5846, opened 2022-04-12
Fix #5845 raster always on VlnPlot

### LINKED ISSUE #5845: raster always enabled in VlnPlot
Hi,

Just a quick note that there's a bug in `SingleExIPlot` which always turns on rasterization even when `raster=FALSE`. 

```r
  if (PackageCheck('ggrastr', error = FALSE)) {
    # Set rasterization to true if ggrastr is installed and
    # number of points exceeds 100,000
    if ((nrow(x = data) > 1e5) & !isFALSE(raster)){
      message("Rasterizing points since number of points exceeds 100,000.",
              "\nTo disable this behavior set `raster=FALSE`")
    }
    raster <- TRUE
  }
```

image

samuel-marsh:
Hi Seurat Team,

Quick fix for #5845 which turned raster on all the time as long as the package check passed instead of if the raster conditions were met.  Just moved `raster <- TRUE` up one set of brackets.

Best,
Sam

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Seurat
-Version: 4.1.0.9003
-Date: 2022-03-29
+Version: 4.1.0.9004
+Date: 2022-04-15
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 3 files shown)


saketkc:
Thanks @samuel-marsh!

samuel-marsh:
Happy to help! :)

PR #6070, opened 2022-06-13
Add ReadDragen to support Illumina Dragen output

parvsachdeva:
Added a function `ReadDragen` that is capable of ingesting output from Dragen scrna pipeline.

An optional barcode summary file (`*barcodeSummary.tsv`) is searched for in the input directory and used to filter `*barcodes.tsv`.

Other functionalities are similar to `Read10x`

# Note

### DIFF
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -3479,3 +3479,185 @@ RegressOutMatrix <- function(
   dimnames(x = data.resid) <- dimnames(x = data.expr)
   return(data.resid)
 }
+
+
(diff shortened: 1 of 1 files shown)


AustinHartman:
Hi, thanks for the PR. It looks like there is a bit of duplication with Read10x, so is it possible to use Read10x to load your data in (possibly after renaming files and specifying the barcodes you wish to filter after loading)? If not, this function is probably better suited to be in [seurat-wrappers](https://github.com/satijalab/seurat-wrappers)

PR #6348, opened 2022-08-26
Allow AddModuleScore to return the scores data frame

scottgigante-immunai:
In order to run `AddModuleScore` in parallel, it would need to not return `object`, but instead just the metadata. This PR allows the user to set this behavior if desired

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -125,9 +125,12 @@ AddAzimuthScores <- function(object, filename) {
 #' @param search Search for symbol synonyms for features in \code{features} that
 #' don't match features in \code{object}? Searches the HGNC's gene names
 #' database; see \code{\link{UpdateSymbolList}} for more details
+#' @param return_object Return \code{object} with the module scores stored as
+#' metadata. If \code{FALSE}, returns the module scores as a data frame.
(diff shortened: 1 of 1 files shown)


samuel-marsh:
Wonder if it might be better to modify to enable parallel internally with future package functions?  Looking at internals feels like it probably should work.
Then could still return object which function kind of defines with having “Add” in name.

scottgigante-immunai:
I would be happy with that solution, but I'm probably not the right person to write it :)

samuel-marsh:
@scottgigante-immunai I got you you covered ;).

I've got code working, just running quick timing check.  If you want to close this PR I'll push new one shortly and tag you on it so you know when incorporated into develop branch.

Best,
Sam

scottgigante-immunai:
You're a champion @samuel-marsh . Thanks for the quick feature addition!

samuel-marsh:
@scottgigante-immunai happy to help.  Found little bug though so troubleshooting that.  Hope to push PR tomorrow though.

samuel-marsh:
@scottgigante-immunai PR is live and tagged you.

Best,
Sam

PR #6440, opened 2022-09-19
Xenium support

pmarks:
Preliminary Xenium support for Seurat. Opening this as a draft to get the discussion going. We are very open to suggestions about how to approach this.  I'm also curious what general work in required in the `feat/imaging` branch before it can merge to `develop`. We should have some bandwidth to help with that.

Open questions:
- [x] `transcripts.csv.gz` loading speed - this seems to be the bottleneck for most datasets.  We currently use `read.csv`, but could we try `readr::read_csv` or `data.table::fread`? 
- [x] `transcripts.csv.gz` - probably want to make optional? Can get very large.
- [x]  How to handle "Negative Control Probe" and "Negative Control Padlock" feature types. These are encoded as a separate `feature_type` in the matrix, but they are really technical controls of the in-situ RNA assay. They aren't useful for biological analysis, but we emit them to help users assess the specificity of the assay. Right now we're just dropping them in `LoadXenium` - should we provide an optional way to include them?
- [ ] Morphology image - do we want to have the DAPI image available as a background, similar to how we have the H&E image for Visium.
- [x] refine version requirements: I don't think SeuratObject 4.1.1 is actually required. It's possible the sp 1.5 is required, but that version is specified in SeuratObject.


mojaveazure:
Hi Pat, thanks for starting this! I'll try to respond to your questions as best as I can

- loading speed: `data.table::fread` is the fastest out of the three and I would suggest using that over `read.csv`; we currently don't depend on `data.table`, but leave it as a suggested package. I would recommend the same `Read/LoadXenium`. One thing to note is that `data.table::fread` doesn't natively support gzipped files, so you'll also need `R.utils` to handle that (most people use a shell command to handle gzipped files in `data.table::fread`, but that is not portable to Windows)
- `transcripts.csv.gz` being optional: all of our imaging `Read/Load*` functions support partial loads. For example, `ReadVitessce` can load the counts, coordinates, molecules, or any combination of the three. I would recommend a similar approach for `Read/LoadXenium`
- negative probes: pinging @AustinHartman
- morphology images: the `FOV` structure does not currently support backing images. This is on our list of things to support, but we do not have a timeline for this
- version requirements: it's our practice that every new release of {Seurat} is pinned to the latest version of {SeuratObject}; when this goes to CRAN, the minimum version of {SeuratObject} will be updated to the latest CRAN version (currently 4.1.2)

AustinHartman:
Regarding the negative probes, I think an optional way to include them would be beneficial for QC. Negative control probe counts and negative control padlock counts can be stored as additional assays in the Seurat object

pmarks:
@mojaveazure @AustinHartman OK, I think this is ready for a closer review now.  @jsicherman from 10x may also help get this in shape.

RajneeshSrivastava:
Team,
Now that 10x has dropped transcripts.csv.gz from the xenium outputs in the new release ([release notes](https://www.10xgenomics.com/support/software/xenium-onboard-analysis/latest/release-notes/release-notes-for-xoa))
what should be the strategy for data analysis given that I have multiple Xenium runs from the previous version (v2.0.1.0) and a couple of runs from recent v3.0.0.0. Is there a possibility to integrate still using Seurat pipeline?
https://github.com/satijalab/seurat/issues/9060

jsicherman:
@RajneeshSrivastava #8605 adds support for reading the `.parquet` files through the `arrow` library (relevant lines are [here](https://github.com/satijalab/seurat/pull/8605/files#diff-71e85752f8998371335d4cee5d51ccb8d7956746bc65e45f63d4f5e5acfe06d3R2544-R2554))

Once that lands, loading either `.csv.gz` or `.parquet` for Xenium outputs should be fully supported 😄 in the meantime, you could also install from the branch

PR #6527, opened 2022-10-10
Support for integrating large datasets

zhanghao-njmu:
When working with large datasets, change the integration matrix format to ["spam"](https://cran.r-project.org/web/packages/spam/index.html) to avoid the "long vectors not supported yet" error.


zhanghao-njmu:
PRs should be made with develop as the base branch.

Issue #6900, opened 2023-01-27
SpatialFeaturePlot plotting multiple images with same scale 


psecekartal:
Hello, 
I have several visium slides and I would like to plot them by using same scale to be able to compare, however min.cutoff/max.cutoff parameters doesnt work. Can you add a function like keep.scale as in FeaturePlot? Or do you have any suggestion?
Thank you, ece

alikhuseynov:
I'm not from the Seurat development team, however, as an alternative you can use the `slot = "scale.data"` which are the z-scores (zero mean, unit variance) for plotting the genes.

Another way is to scale the `slot = "data"` to range of `[0,1]`
Here is an exampe:
```
# eg on PBMC small dataset
library(magrittr)
scaled.mat <- 
  sapply(rownames(pbmc_small) %>% seq, 
         function (i) { 
           scales::rescale(GetAssayData(pbmc_small, assay = "RNA", slot = "data")[rownames(pbmc_small)[i],], to = c(0,1)) }
  ) %>% t

# check expression ranges
quantile(scaled.mat, c(0.01, 1))
# should output:
#1% 100% 
#0    1 

# add genes names
dimnames(scaled.mat)[[1]] <- rownames(pbmc_small)
# set new scaled matrix to the slot "scale.data"
pbmc_small <- SetAssayData(pbmc_small, slot = "scale.data", scaled.mat)

# now plot genes as needed.. using slot = "scale.data"
```

samuel-marsh:
Hi @TeodoraTockovska

The PR was recently merged and is part of the dev branch.  Hopefully can be of use to you.

Best,
Sam

PR #6905, opened 2023-01-30
keep.scale for  SpatialPlot/SpatialFeaturePlot

samuel-marsh:
Hi Seurat Team,

This is PR to try and address #6900.  The code and implementation is based on `FeaturePlot` implementation but obviously feel free to adjust if you have different preferences for the spatial plotting version.  It defaults to `"feature"` to scale all images to the feature max for the individual feature but also accepts `"all"` to scale all images to the global feature max of all features being plotted, or `NULL` to scale all images individually.

Best,
Sam


AustinHartman:
Thanks for the PR!

PR #6986, opened 2023-02-27
fix(Xenium): load renamed feature

jsicherman:
The feature type got changed from `Blank Codeword` to `Unassigned Codeword`, so this just loads from the new feature if the the old feature isn't found.


Ceallaig33:
Hi @AustinHartman and @jsicherman - thanks for the fix on the develop branch. Looks like not updated on the seurat5 branch yet maybe? We ran into issue when trying LoadXenium() with xenium-1.3.0.6 output in seurat5 install.

jsicherman:
Hi @Ceallaig33 - that's right, it looks like this has not yet merged to master or been in a CRAN release. Apologies for the inconvenience! `devtools::install_github("satijalab/seurat@develop")` should get you the fix from this branch in order to restore functionality on 1.3+ outputs until the official release.

PR #7444, opened 2023-06-13
proportional sketching

inofechm:
I have added a parameter to the SketchData function which allows one to sketch a proportion of total cells per layer/dataset rather than a fixed ncells. 
This is particularly useful if integrating multiple datasets that different vastly in size and where we want to sample these datasets proportionally. ie. if one dataset has 20 000 cells and another has 200 000 cells and prop.ncells=0.1 it sketches 2000 cells from first dataset and 20000 cells from the next.

### DIFF
--- a/R/sketching.R
+++ b/R/sketching.R
@@ -29,6 +29,7 @@ SketchData <- function(
   over.write = FALSE,
   seed = 123L,
   cast = 'dgCMatrix',
+  prop.ncells = NULL,
   verbose = TRUE,
   ...
 ) {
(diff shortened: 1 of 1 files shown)


winner0809:
has this parameter been successfully implemented? I tried it, didn't seem to progress, just showing "Calcuating Leverage Score" and no error message either

winner0809:
has this parameter been tested successfully? I run the following script but the end result just showed default setting of 5K from each RNA count layer was sketched rather than proportionally based on the cell number in each layer.

`obj <- SketchData(
  object = obj,
  prop.ncells = 0.04,
  method = "LeverageScore",
  sketched.assay = "sketch"
)`

PR #4101, opened 2021-02-18
Modify Read10X for GEO compatibility

samuel-marsh:
Hi Seurat Team,

This PR is in response to #4096 which should simplify things for importing scRNA-seq data downloaded directly from GEO.  Should be fully compatible with both the import of a single dataset as well as when supplying a named vector of directories to the `data.dir` parameter.

Thanks again for all your hard work!!
Best,
Sam

PR #7531, opened 2023-07-05
just a tiny suggestion

opitzl:
Dear Seurat-Dev-Team,
it would be nice if the Read10X function 
would support an optional prefix in the CellRanger output data or to read 10X data from GEO without renaming.

### DIFF
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -789,7 +789,8 @@ Read10X <- function(
   gene.column = 2,
   cell.column = 1,
   unique.features = TRUE,
-  strip.suffix = FALSE
+  strip.suffix = FALSE,
+  prefix = ''
 ) {
   full.data <- list()
   has_dt <- requireNamespace("data.table", quietly = TRUE) && requireNamespace("R.utils", quietly = TRUE)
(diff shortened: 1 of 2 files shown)


samuel-marsh:
Agree this would be great if implemented.

In meantime as fyi I have written functions to handle such situations (GEO files etc) in my package scCustomize.  `Read10X_GEO` can read in one or more files from same directory with prefixes and if you are reading all files doesn’t require you to specify prefix or files to share prefix (will return list with entries named by prefix).
https://samuel-marsh.github.io/scCustomize/articles/Read_and_Write_Functions.html

Best,
Sam

samuel-marsh:
Will also add if you want Seurat native solution you can simply use `ReadMtx` function.

Best,
Sam

samuel-marsh:
Also actually just remembered `ReadMtx` was solution added when I sent similar PR while back #4101 .  So think that will likely remain supported solution to file prefix/non-standard names.

Best,
Sam

saketkc:
Hi @opitzl and @samuel-marsh, thanks for the suggestion and discussion. We think `ReadMtx` is a versatile approach to handle these cases. Thanks for your PR @opitzl - we really appreciate it, but also want to keep the technology specific methods (such as 10X) to do exactly one task. Please use `ReadMtx` for other use cases.

samuel-marsh:
Thought as much based on our previous conversations!  Thanks again @saketkc!

PR #7566, opened 2023-07-12
Correct lost gene expression column when loading NanoString CosMx data

mtgregory:
The LoadNanostring() function drops the first gene from the assay expression matrix.  This patch corrects this and no longer drops this gene from the expression matrix.

Specifically, the gene is dropped only from the "assay" slot but not from the "images" slot as show here:
image

Example of this issue not present after the patch in this PR:
image


AustinHartman:
Thanks for the PR!

Issue #7745, opened 2023-08-31
Disable warning of rasterization after 100k points in DimPlot, FeaturePlot etc.


Peter-bbh:
When number of points to plot exceeds 100,000, the plot is defaulted to rasterization and a warning is issued, "Rasterizing points since number of points exceeds 100,000. To disable this behavior set `raster=FALSE`".

First time it is great, but always is anoying. It can be removed by suppressWarnings(), see #6802, but it clutters the code.

An easy fix seems to be to change the preceeding condition in "visualization.R" (three places) from
"if ((nrow(x = data) > 1e5) & !isFALSE(raster)){" to "if ((nrow(x = data) > 1e5) & **is.null**(raster)){"

In this way, only if the raster option is not already specified, the warning will appear. So setting "raster = TRUE" will produce the same plot, but without the warning.

samuel-marsh:
@Peter-bbh this is now to be implemented in Seurat v4.4 (release soon).  See #7842.

Best,
Sam

PR #7842, opened 2023-09-27
Change raster warning only if not explicitly supplied

samuel-marsh:
Hi Seurat Team,

This is small change to clean up console output.  Currently plots based on `SingleExIPlot` or `SingleDimPlot` issue a raster warning whenever the raster parameter is not `FALSE` and the number of points is over 100K.  However, because of this it also continuously prints the message even when user explicitly sets the parameter to be TRUE.  This can clutter output especially if return lots of plots in loop for something the user has explicitly set.  This was mentioned previously in #7745.  While I know that you can suppress such warnings many users may not.

This PR simply changes the `!isFALSE(raster)` to `is.null(raster)` since NULL is the default setting.  Therefore it will only print warning when setting is not explicitly provided by the user.

This may have been specific design decision to warn every time plot is rasterized but seems unnecessary if user is supplying parameter.  If it is  specific design decision then feel free to ignore and close this PR.

Thanks as always,
Sam

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -8123,7 +8123,7 @@ SingleDimPlot <- function(
   raster = NULL,
   raster.dpi = NULL
 ) {
-  if ((nrow(x = data) > 1e5) & !isFALSE(raster)){
+  if ((nrow(x = data) > 1e5) & is.null(x = raster)){
     message("Rasterizing points since number of points exceeds 100,000.",
             "\nTo disable this behavior set `raster=FALSE`")
   }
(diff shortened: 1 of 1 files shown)


saketkc:
Thank you @samuel-marsh! I have implemented the change in https://github.com/satijalab/seurat/commit/643a57d2e10eda962b7b8ae4d0e35b5f8a590b21. Since we were about to release 4.4.0, we decided to do this internally rather then merging your PR. Thank you for your PR!

samuel-marsh:
Hi @saketkc, Sounds good, thanks!!

samuel-marsh:
Hi @saketkc,

I just wanted to let you know for when you propagate changes to the V5 branch (or potentially hotfix the 4.4.0 branch) there is one error in the raster changes in the 4.4.0 branch.  It was actually plot I hadn't caught (`SingleCorPlot`).

The exclamation point before the `is.null` needs to be removed.
https://github.com/satijalab/seurat/blob/57564e282a1ba54c655192d907d83df21164dc78/R/visualization.R#L7845

Right now it will warn when explicitly TRUE or FALSE.

Best,
Sam

saketkc:
Thanks Sam! I missed it too. We might release a hotfix and propogate to V5 (the latest V5 is synced up with V4.4.0)

samuel-marsh:
Sounds good!  I'll close this.  Thanks again!

Issue #7891, opened 2023-10-11
Seurat V5: Bug & Fix for DimPlot cell highlighting with rasterization


samuel-marsh:
Hi Seurat Team,

Tagging @Gesmira because this is related to potential V5 release issues in dependent packages posted in scCustomize package.

Also posting this here as opposed to PR because not exactly sure how you prefer to handle it since it involves both `DimPlot` but also likely a change to `SetHighlight` and probably a warning message.  This bug/error was posted previously here #7789 and error that propagates through to scCustomize was posted this week [scCustomize #132](https://github.com/samuel-marsh/scCustomize/issues/132).

In short the issue is that `geom_scattermore` requires the `pointsize` parameter during rasterization to be a single value where as DimPlot gives each cell a value (which under non-raster conditions allows for different sizes if desired when using `cells.highlight`).

These were my thoughts for fix and if it sounds good I'm happy to take stab at PR or if you'd rather just take care of internally that's all good too.

**Fix to allow successful plot**
1. Add additional parameter and a check to `SetHighlight` to check raster parameter and if `raster = TRUE` change the return value of `highlight.info$size` to be single length vector.
```
# current SetHighlight code
size <- rep_len(x = pt.size, length.out = length(x = cells.all))
highlight <- rep_len(x = NA_character_, length.out = length(x = cells.all))
if (length(x = cells.highlight) > 0) {
  for (i in 1:length(x = cells.highlight)) {
    cells.check <- cells.highlight[[i]]
    index.check <- match(x = cells.check, cells.all)
    highlight[index.check] <- names.highlight[i]
    size[index.check] <- sizes.highlight[i]
  }
}

# new check for raster parameter
if (isTRUE(x = raster)) {
  size <- size[1]
}
```
- Alternative would be to leave `SetHighlight` alone and simply modify within `SingleDimPlot` right before creating plot (though I think `SetHighlight` change makes more sense to me).  

**Add check if user sets values for `pt.size` and leaves default `sizes.highlight` and raster is set TRUE (by default or manually)**
1. This let's user know that when setting `cells.highlight` when `raster = TRUE` then the user supplied value of `pt.size` will be disregarded in favor of the `sizes.highlight` value and that when plot is rastered the background and highlighted cells will be the same size. 

something like:
```
if (!is.null(pt.size) && sizes.highlight != pt.size && raster = TRUE) {
  warn(...)
  }
```

[...]

PR #7914, opened 2023-10-17
Highlight fix

samuel-marsh:
Hi @Gesmira & Seurat Team,

This is PR addresses two issues:
First, addressing #7891.  I have added raster parameter to SetHighlight, and a check to change the pt.size returned if TRUE.

I have also made the fix mentioned in comments for PR #7842 which causes raster warning to accidentally be always displayed regardless of whether TRUE/FALSE is supplied.

Thanks again!
Best,
Sam


samuel-marsh:
fyi: Appears the CI check error is due to BPCells install issue.

Gesmira:
Thanks Sam! Looks great!

PR #7938, opened 2023-10-26
Hotfix highlight

samuel-marsh:
Hi @Gesmira @saketkc,

Tagging you both here in hopes this gets seen before CRAN submission as I know that is imminent.  I found potential error in the PR (#7914) I pushed that was merged last week.

The prior code would have still allowed pt.size to override sizes.highlight in raster scenario depending on the ordering of the cells/identity being highlighted within the overall object.  (Fix `SetHighlight`).  The second commit addresses when the warning is presented to the user so that it is informatively presented.

Best and thanks as always!
Sam

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -8200,7 +8200,7 @@ SingleDimPlot <- function(
   raster <- raster %||% (nrow(x = data) > 1e5)
   pt.size <- pt.size %||% AutoPointSize(data = data, raster = raster)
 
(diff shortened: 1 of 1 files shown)


saketkc:
thanks for the PR @samuel-marsh! We were very swamped with the CRAN submission, so we'll incorporate it in the next release. Still waiting to hear from CRAN, but happy to merge it to develop once it makes it to CRAN.

samuel-marsh:
Hi @saketkc,

I figured that was probably be the case and no worries at all.  Sorry for not recognizing the error in original PR.

Best of luck with full CRAN submission!

Thanks,
Sam

Issue #4545, opened 2021-05-27
UpdateSymbolList generates wrong symbols


apredeus:
Dear Seurat team, 

It seems like `UpdateSymbolList` uses a logic that results in erroneous changes to some gene symbols. 

There are quite a few genes for which official HGNC symbol co-insides with a symbol that was used for *another* gene in the past. The function automatically changes these. For example, HBM is a current name for "hemoglobin subunit mu", and an old name for "LDL receptor related protein 5", which are two totally different genes (https://www.genenames.org/tools/search/#!/?query=HBM). The function, however, converts HBM to LRP5. Similar thing happens to HBG1 and quite a few other genes.

I think there needs to be a check if gene symbol exists in the newest version - and if it does, the symbol should not be changed.

andrewwbutler:
Hi,

Thanks for reporting! We've made some fixes/updates to this function which hopefully resolve this problem. Please try installing the `develop` branch (instructions [here](https://satijalab.org/seurat/articles/install.html#install-the-development-version-of-seurat-1)) and see if that fixes the issue for you.

PR #8179, opened 2023-12-12
Fix: UpdateSymbolList incorrectly renames genes

samuel-marsh:
Hi Seurat Team,

This is PR that builds on previous fix described in #4545.  This is by no means perfect fix (explanation below) so I leave it to you to decide path forward.  If you decide a different solution is warranted I'm happy to help with PR if desired.

The issues is with potential for `UpdateSymbolList` to inappropriately rename genes.  In original case the search for alias symbols was removed from internals of `UpdateSymbolList` by manually setting the parameter to previous symbols only.  However, that unfortunately still causes issues as there are a number of previous symbols which are now symbols of different genes.  For instance the genes MCM2, MCM7, and CCNL1 which are all currently approved genes.  However, in current form `UpdateSymbolList` reverts changes:

```
> UpdateSymbolList(symbols = c("MCM2", "MCM7", "CCNL1"))
  |==============================================================================================================| 100%
Found updated symbols for 2 symbols
MCM2 -> MCM7
CCNL1 -> MCM2
[1] "MCM7" "MCM7" "MCM2"
```

Using the most recent 10X human reference genome (which is filtered so this is not full extent of potential issues), I have found >100 genes which would be inappropriately swapped.

The "simple" solution which is in this PR to avoid potential issues I added parameter to require that an object be specified and synonyms only be used if they are genes not already found in the object.  This limits the function to use with Seurat object but protects against inappropriate renames (though not completely).

The reason it's not complete solution is because most Seurat objects are filtered versions of the count matrix and this often results in objects with half the genes present in the annotation file.  Therefore the function does still leave the possibility to inappropriately rename a gene if it was gene that was filtered out during object creation.  In order to avoid completely, it different fix and for Seurat to store full feature list from the `counts` input somewhere in the object to check against vs checking against the current features with `Features`.

Again if current PR solution is not desired that is totally fine but wanted you to be alert to issue.

Best,
Sam

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -221,7 +221,7 @@ AddModuleScore <- function(
           if (search) {
             tryCatch(
               expr = {
-                updated.features <- UpdateSymbolList(symbols = missing.features, ...)
+                updated.features <- UpdateSymbolList(object = object, symbols = missing.features, ...)
                 names(x = updated.features) <- missing.features
                 for (miss in names(x = updated.features)) {
                   index <- which(x == miss)
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -1726,6 +1726,7 @@ SetQuantile <- function(cutoff, data) {
 #' }
 #'
 UpdateSymbolList <- function(
+  object,
   symbols,
   timeout = 10,
   several.ok = FALSE,
@@ -1740,6 +1741,10 @@ UpdateSymbolList <- function(
     verbose = verbose,
     ...
   ))
+
+  # avoid incorrect renaming
+  new.symbols <- new.symbols[!new.symbols %in% Features(object)]
+
   if (length(x = new.symbols) < 1) {
     warning("No updated symbols found", call. = FALSE, immediate. = TRUE)
   } else {


samuel-marsh:
Hi Seurat Team,

Again still don't have perfect answer but just thought I would provide update here as alternative (though more conservative) method for updating genes.

I have been testing function in dev branch (branch: file_cache_dev; https://github.com/samuel-marsh/scCustomize/blob/c14063f8cd34f3f9f94903c7e82980c68cbd3a84/R/Utilities.R#L1868) of scCustomize to handle things slightly differently.

The function I wrote pulls the entire HGNC data csv (stores as cache using BiocFileCache to avoid need to download every time but allows for updating via cache update).  It then filters input symbols to only symbols which are not currently approved symbols and only checks those symbols to see if they are listed in previous symbols and if they are it provides updated approved symbol.

I say it's more conservative because while it prevents mis-naming of genes it does allow for potential of genes whose names are not updated.  There are examples of genes who have swapped symbols with each other.  Therefore they could potentially be filtered out and not updated.

The level of conservative-ness probably depends on age of input gene set you are inputting as more recent gene sets are less likely to have genes which go un-renamed compared to older.

Again I don't really know what perfect solution to the issue is when only input is gene symbols and not entrez/ensembl ids.  But thought I would make you aware of this potential solution in addition to problems described in first post here.

Best,
Sam

mschilli87:
Is there any update on this front? Is this still maintained/recommended or has this been solved in a different way meanwhile?

samuel-marsh:
Hi @mschilli87,

I’ve created functioning my package scCustomize which can handle this now.  It also works offline after first use with internet.

https://samuel-marsh.github.io/scCustomize/articles/Update_Gene_Symbols.html

Best,
Sam

mschilli87:
So maybe this PR should be replace by one removing Seurat's own implementation and importing that one instead then? If this got merged, the code would be duplicated and need to be maintained in two places.

samuel-marsh:
So this PR doesn’t implement that function from scCustomize as I was going for minimizing dependencies and minimal difference in output format of existing function.  I leave it to Seurat team to decide how they want to implement or change.

Best,
Sam

Issue #7087, opened 2023-03-30
Option to adjust geom_point stroke for FeaturePlot, DimPlot, etc to allow for smaller points


liz-is:
I often find that when I'm plotting relatively small plots (e.g. 4x4 inches pdf) with many points, reducing the point size with the`pt.size` argument to `FeaturePlot`, `DimPlot`, etc, seems to have no impact on the actual point size in the resulting plot. It's basically the same issue as described [here](https://stackoverflow.com/questions/34638902/point-size-in-ggplot-2-0-0). Based on the answers there, I made a custom version of `SingleDimPlot` where I set `stroke = 0` in the call to `geom_point` (i.e., [here](https://github.com/satijalab/seurat/blob/master/R/visualization.R#L8123)). This gave me the desired output - smaller points with less overplotting at the necessary output size. 

I was wondering if this change could be made in the code, or if you could add an option, similar to `pt.size`, that 
would allow users to customise stroke width as needed?

bschilder:
I'd find this incredibly helpful.
To the Seurat team, please do consider merging the PR by @samuel-marsh which appears to resolve the issue.
- https://github.com/satijalab/seurat/pull/8180

vdsukhov:
I encountered the same issue where I couldn't control the stroke size using the `FeaturePlot` function. I'm currently using Seurat version **5.2.1**. As a workaround, I manually adjusted the stroke size directly within the resulting ggplot objects. Below is a minimal example that demonstrates how to modify the stroke size after the `FeaturePlot` has been generated:

```R
fps <- FeaturePlot(pbmc, features = c("MS4A1", "GNLY", "CD3E", "CD14", "FCER1A", "FCGR3A", "LYZ", "PPBP", "CD8A"))
for (i in seq_along(fps)) {
    fps[[i]]$layers[[1]]$aes_params$stroke <- 0
}
fps
```

After that, I can freely adjust the point size as needed:

```R
for (i in seq_along(fps)) {
    fps[[i]]$layers[[1]]$aes_params$size <- 1.25
}
fps
```

---
Interestingly, the `DimPlot` function includes a `stroke.size` argument that works as expected, allowing direct control over the stroke size. It's a bit surprising that `FeaturePlot` doesn't have it, even though it in some way similar to `DimPlot`.

longmanz:
Hi, thanks to @samuel-marsh and @vdsukhov for submitting PRs that fix this bug. It will be merged into the main soon so I am closing this issue.

PR #8180, opened 2023-12-12
Add stroke.size parameter DimPlot

samuel-marsh:
Hi Seurat Team,

This is PR to address #7087.  This adds the `stroke.size` parameter to `DimPlot` and consequently `SingleDimPlot`.  The parameter is set to `NULL` by default.  When this is the case the `stroke.size` is set to static value which appears to be default ggplot2 value.  I tested the plotting with this value of stroke.size vs. current DimPlot and the plots appeared identical to me.

Best,
Sam

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.1.0.9014
+Version: 5.1.0.9015
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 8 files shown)


samuel-marsh:
Hi @dcollins15,

Here is quick PR I thought I would ping.  Adds single parameter to `DimPlot`.  I just rebased branch but let me know if there is anything else you need me to do!

Best,
Sam

dcollins15 (approved):
Another solid one—thanks @samuel-marsh 👍

PR #8203, opened 2023-12-19
Update umi.assay in GetResidual

aladyeva-wustl:
The current implementation of GetResidual in PrepSCTIntegration does not allow applying it to Spatial datasets as they don't contain "RNA" assay which is set as default value of parameter umi.assay.
This fix change default umi.assay to NULL and in the further check picks RNA or Spatial values, depending on which is presented in the object. If both are presented it defaults to the first value.
Also allows easy to extend of any future default assay values.

# Note


saketkc:
Hi @aladyeva-wustl, thanks for the PR! Our recommendation would be to use `umi.assay="Spatial"` when invoking `GetResiduals`. The default of `RNA` is to cater to most frequent use case (RNA -> SCTransform). While your solution is good, it would not be helpful if the default layer is anything except RNA or Spatial (and would anyway require the user to specify the parameter explicitly)

aladyeva-wustl:
Hi @saketkc we can't explicitly specify umi.assay, because we are running PrepSCTIntegration function and there are no such parameters that are passed to GetResiduals.
My decision to add both default values to GetResiduals instead of PrepSCTIntegration is motived by idea that if umi.assay is missing in other functions they will also not work for Spatial transcriptomics data.

saketkc:
I see, that is a valid point.

Issue #8267, opened 2024-01-05
Error with jackstraw with matrix stored on disk (bpcells)


audrey-gibert:
Hello,
I'm working with a dataset stored on disk and I can not run the `jackstraw` command. The returned error message : 
`Erreur dans order(c(col(x = x2)), runif(n = length(x = x2))) :   arguments de tailles différentes`

### The reproducible example
<details>

<summary>Library loading</summary>

```{r}
library(dplyr)
library(Seurat)
library(ggplot2)
library(hdf5r)
library(hrbrthemes)
library(future)
library(tidyverse)
library(magrittr)
library(BPCells)
library(Azimuth)
library(glmGamPoi)
library(presto)
options(future.globals.maxSize = 55000 * 1024^2, # 1 * 1024^2 -> 0.001Gb RAM
        future.seed = TRUE)
plan("multisession", workers = 24)
library(conflicted)
conflicts_prefer(dplyr::select)
conflicts_prefer(dplyr::filter)
```
</details>

```{r}
# Loading the pbmc3k dataset
library(SeuratData)
AvailableData()
InstallData("pbmc3k")
data("pbmc3k")
UpdateSeuratObject(pbmc3k)
attributes(pbmc3k)$images <- list()
pbmc3k[["RNA"]] <- as(object = pbmc3k[["RNA"]], Class = "Assay5")
pbmc3k <- subset(pbmc3k, downsample = 500)
pbmc3k_BPcells <- pbmc3k
```
```{r}
# It works fine with the classical seurat object
pbmc3k <- NormalizeData(pbmc3k)
pbmc3k <- FindVariableFeatures(pbmc3k)
pbmc3k <- ScaleData(pbmc3k)
pbmc3k <- RunPCA(pbmc3k, reduction.name = "pca")
pbmc3k <- JackStraw(pbmc3k, reduction = "pca")
```
```{r}
# First, loading with BPCells package
my_dir <- '/home/augibert/Documents/Projects/test/brain_counts'
write_matrix_dir(mat = pbmc3k_BPcells[["RNA"]]$counts, dir = my_dir)
counts.mat <- open_matrix_dir(dir = my_dir)
pbmc3k_BPcells[["RNA"]]$counts <- counts.mat
# Then, same piece of code as before
pbmc3k_BPcells <- NormalizeData(pbmc3k_BPcells)
pbmc3k_BPcells <- FindVariableFeatures(pbmc3k_BPcells)
[...]

PR #8271, opened 2024-01-06
Improve BPCells compatibility (JackStraw + RunPCA)

bnprks:
These are some changes brought about from [this BPCells issue](https://github.com/bnprks/BPCells/issues/62), then referenced in Seurat issue #8267

1. Make the `JackStraw` function compatible with BPCells objects. This approach shuffles the matrix subset as a dense matrix, but leaves the remainder of the data on-disk. For best performance the BPCells object should be stored in row-major order so that the matrix subsets can be performed efficiently, but it will work either way.
2. Improve performance of `RunPCA` by using the BPCells svds function. This takes advantage of the same solver RSpectra uses, but at the C++ level to avoid overheads of repeatedly creating BPCells C++ objects on each solver iteration. It also avoids the pitfall that irlba can only use its fast C version with in-memory matrices and otherwise falls back to a slower pure-R implemenation.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.1.0.9011
+Version: 5.1.0.9012
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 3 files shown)


dcollins15:
Apologies @bnprks! My intent was to merge this in on the heels of #8966 but got pulled off in another direction. Circling back now as we prepare for the next CRAN release 🙃 The good news is that with CI checks now properly configured, we should be much more responsive to pull requests going forward.

These changes LGTM 🚀 

This will be the next PR to merge, but before that, we'll need to:
- Rebase `bnprks:develop` onto `satijalab:develop`.
- Update the changelog.
- Bump the package version (5.1.0.9011 -> 5.1.0.9012).

If you're keen to tackle these updates, feel free to push them up; otherwise, I'll take care of it shortly. 

Thanks, as always, for the new features and for your patience as we integrate them 🙏

bnprks:
Hi @dcollins15, glad to hear you're putting in a new CRAN release! I made the changes you requested, though feel free re-word the changelog message or edit anything else as needed -- I don't mind my branch getting edited even with a force push since I'm just using the branch for this PR

dcollins15 (approved):

Issue #5614, opened 2022-02-10
Shifted high resolution image with  Load10X_Spatial()?


Mnew1288:
Hi,

I am not sure it's a bug or not. Anyway, no problem to use the default low resolution PNG image. But when I used the high resolution PNG image, I got something like this:
![test](https://user-images.githubusercontent.com/99414904/153438143-9435d3b2-d6f0-44e5-a07b-c99d5a3878bb.png)

code: plot2 <- SpatialFeaturePlot(MG.spatila_data, features = "nCount_Spatial") + theme(legend.position = "right")

Seems the dots are not overlapped with the image. By the way. I noticed my section is pretty small and have only 375 dots on it. 

Thanks

Lily-life:
Hi,
I have the same issue. The plot does not overlay correctly when high resolution PNG image is used.
![ACTA2_small](https://user-images.githubusercontent.com/6047337/158625388-34f8550a-a8bc-4a67-aa5d-ae5ad3703c42.png)
Could you fix that? Thank you!

biosyssun:
When I loaded high resolution image, SpatialFeaturePlot cannot works well.

img <- Read10X_Image(paste(dat.dir,'spatial',sep='/'),image.name = 'tissue_hires_image.png')
seurat.obj <- Load10X_Spatial(dat.dir,image=img)

Please fix it.

I used Seurat 4.1.1

jincanke:
Hi, @biosyssun I encountered the same problem. Have you solved it? Could you help me?

ggruenhagen3:
Hi,

I encountered the same problem as @Mnew1288 and I found a solution that works for me. I noticed that the points seemed scaled down small in the upper left corner and I figured that we must be using the wrong scalefactor. Seurat uses something they call `scalefactors` to adjust the size of the dots/image. This is stored in `object@images$slice1@scale.factors` and there are 4 of these scale factors: spot, fiducial, hires, and lowres. Seurat seems to expect in general that low-res images are being used, so it uses the lowres scalefactor during plotting. Since we're trying to use the high-res image, we want to use the hires scalefactor instead. I don't there's a direct way to tell Seurat to do this, but instead we can just set the lowres scalefactor to the hires scalefactor. Like this:

`object@images@slice1@scale.factors$lowres = object@images@slice1@scale.factors$hires`

Hope this helps!

lsudupe:
@ggruenhagen3 thank you so much for this solution!

[...]

PR #8383, opened 2024-01-26
Try to restore high res image import.

### LINKED ISSUE #8381: `Read10X_Image` high resolution option with version > 4
Hello,

It seems to me that in version < 5, it was possible to import a high resolution image with `Read10X_Image()`: 
https://github.com/satijalab/seurat/blob/ff03fdf21f1b8fea9ee247d0fd83df5811507027/R/preprocessing.R#L1033C1-L1034C62

Then, the scale factor could be modified as per #5614. 

In the current version, it looks like 'readPNG` called by `Read10X_Image()` always looks for the low res png: 
https://github.com/satijalab/seurat/blob/41d19a8a55350bff444340d6ae7d7e03417d4173/R/preprocessing.R#L1151C1-L1152C77

I might be naive, but it would be nice to have this option back. Thank you.

SJK

kimsjune:
This aims to fix #8381 by restoring `image.name` option for `readPNG()` and consequently `Read10X_Image()`. The master branch version has "tissue_lowres_image.png" hard-coded. This was not the case in an earlier [version](https://github.com/satijalab/seurat/blob/ff03fdf21f1b8fea9ee247d0fd83df5811507027/R/preprocessing.R#L1033C1-L1034C62), so I just copied it back. There is a scale factor [error](#5614) if the high res image is used, which can be mitigated by: 
```
hires <- Read10X_Image(paste0(data.dir,'spatial',sep=''),image.name='tissue_hires_image.png')
hires@scale.factors$lowres = hires@scale.factors$hires
```
Also mentioned [here](https://github.com/satijalab/seurat/issues/5614#issuecomment-1194555472).

Hope this can help.

### DIFF
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -1148,8 +1148,8 @@ Read10X_h5 <- function(filename, use.names = TRUE, unique.features = TRUE) {
 #' @export
 #' @concept preprocessing
 #'
-Read10X_Image <- function(image.dir, filter.matrix = TRUE, ...) {
-  image <- readPNG(source = file.path(image.dir, 'tissue_lowres_image.png'))
+Read10X_Image <- function(image.dir, image.name = "tissue_lowres_image.png", filter.matrix = TRUE, ...) {
+  image <- readPNG(source = file.path(image.dir, image.name))
   scale.factors <- fromJSON(txt = file.path(image.dir, 'scalefactors_json.json'))
   tissue.positions.path <- Sys.glob(paths = file.path(image.dir, 'tissue_positions*'))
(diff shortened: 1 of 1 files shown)


dcollins15:
Thanks for the PR!

We restored the `image.name` parameter for `Read10X_Spatial` in v5.1.0 per your suggestion 👍 

We've also introduced the `image.scale` parameter to `SpatialDimPlot` and `SpatialFeaturePlot` so that you can select the appropriate scale factor to use for your image. Closing this PR since the feature has been implemented 

We're always appreciative of contributions from external collaborators - thanks again!

PR #8659, opened 2024-03-20
Fix PercentageFeatureSet

### LINKED ISSUE #8658: PercentageFeatureSet doesn't always work when specifying assay.
I've been trying to run `PercentageFeatureSet` on an object that already has multiple assays, with default one being `integrated`.

I do the following, but it doesn't work:
```r
> PercentageFeatureSet(object = seurat, pattern = "^RP[LS]", assay = "RNA")
NULL
```
Pretty sure it's because `PercentageFeatureSet` searches for `counts` layers, but doesn't specify which `assay` to look for layers in.
https://github.com/satijalab/seurat/blob/c54e57d3423b3f711ccd463e14965cc8de86c31b/R/utilities.R#L1173-L1174

So, since `integrated` assay doesn't have a `counts` layer nothing works.

My guess is that just pushing the `assay` parameter down to `Layers` call should fix this.

maxim-h:
Fixes #8658

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -1170,7 +1170,7 @@ PercentageFeatureSet <- function(
     warn(message = "Both pattern and features provided. Pattern is being ignored.")
   }
   percent.featureset <- list()
-  layers <- Layers(object = object, search = "counts")
+  layers <- Layers(object = object, search = "counts", assay = assay)
   for (i in seq_along(along.with = layers)) {
     layer <- layers[i]
     features.layer <- features %||% grep(


samuel-marsh:
Hi,

Not member of dev team but just a thought here.  A check should also probably be added to prevent integrated assay from being selected here because as you mentioned it doesn't have counts and shouldn't be used, would just provide more informative error message to end user.

Best,
Sam

maxim-h:
Hi Sam.

It is a good point. As of now it runs without errors, but just returns `NULL`. So it was really hard to understand what the problem was without looking at the code. 

I wouldn't be confident in throwing errors from functions that currently return. Might mess up something elsewhere.
But adding a warning message might be useful. Only it probably should be coming from the `Layers` function and not `PercentageFeatureSet`. So it needs to be a pull request to [SeuratObject](https://github.com/satijalab/seurat-object).

samuel-marsh:
I'll leave up to dev team to decide but personally I think abort in percentagefeatureset makes more sense to me.  End users may still end up confused (for those that don't understand integration doesn't have corrected counts) as to why function fails.  And a warning about integrated assay not being appropriate for this particular function makes sense.  But either way works.

Best,
Sam

Issue #8895, opened 2024-05-15
Load Nanostring meta.data doesn't transfer well 


roanvanscheppingen:
When building a Seurat from Cosmx flatfiles, the import options for meta.data is limited and don't work fully as described. Now the meta.data import is very limited compared to the native Seurat returned by the Cosmx machine. But actually, nanostring recommends building the Seurat using flatfiles. Many meta.data columns are inaccessible using LoadNanostring.

Seurat documentation:

> metadata | Type of available metadata to read; choose zero or more of:
> 
> - “Area”: number of pixels in cell segmentation
> - “fov”: cell's fov
> - “Mean.MembraneStain”: mean membrane stain intensity
> - “Mean.DAPI”: mean DAPI stain intensity
> - “Mean.G”: mean green channel stain intensity
> - “Mean.Y”: mean yellow channel stain intensity
> - “Mean.R”: mean red channel stain intensity
> - “Max.MembraneStain”: max membrane stain intensity
> - “Max.DAPI”: max DAPI stain intensity
> - “Max.G”: max green channel stain intensity
> - “Max.Y”: max yellow stain intensity
> - “Max.R”: max red stain intensity

The meta.data file contains more, but cannot be fully imported.
Bigger bug, defining "Mean.G" (or Y, R) gives **Error in `[.data.frame`(md, , metadata) : undefined columns selected**

>  [1] "nn_59ee731b.4b6e.431c.b051.5ba71eb87378_1_cluster_cluster_64fc893f.482c.409a.a5b3.e29eb0073160_1"
 [2] "RNA_nbclust_f2d90735.f357.4075.a6a6.1b713a91355e_1_clusters"                                     
 [3] "RNA_nbclust_f2d90735.f357.4075.a6a6.1b713a91355e_1_posterior_probability"                        
 [4] "cell"                                                                                            
 [5] "nCount_RNA"                                                                                      
 [6] "nFeature_RNA"                                                                                    
 [7] "nCount_negprobes"                                                                                
 [8] "nFeature_negprobes"                                                                              
 [9] "fov"                                                                                             
[10] "Area"                                                                                            
[11] "AspectRatio"                                                                                     
[12] "CenterX_local_px"                                                                                
[13] "CenterY_local_px"                                                                                
[14] "Width"                                                                                           
[15] "Height"                                                                                          
[16] "Mean.Histone"                                                                                    
[17] "Max.Histone"                                                                                     
[18] "Mean.CD68"                                                                                       
[...]

PR #8900, opened 2024-05-16
Update preprocessing.R to create more metadata import options 

roanvanscheppingen:
Reads all available colnames of the metadata.file instead of hardcoding a few choices to be given in hardcoded choices.

The options to load nanostring data had a hardcoded choice option where things like "Mean.Y" and "Mean.R" are coded. However, the standard Flatfiles from Cosmx data have different colnames. Therefore this code changes the hardcoding into reading the colnames of the metadata.file. See also issue #8895

### DIFF
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -1997,14 +1997,11 @@ ReadNanostring <- function(
     several.ok = TRUE
   )
   if (!is.null(metadata)) {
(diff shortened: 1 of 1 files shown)


swbioinf:
Is that error just an internet connection error? Is it possible for anyone to trigger a rerun the test? This change would be very useful :)

PR #8910, opened 2024-05-17
Fixed min.cells.group issue in `FindConservedMarkers()`

hamelyz:
# Problem

`FindConservedMarkers` throws an error on too few cells even when parameter `min.cells.group` is lower than number of cells. 

Distinct but related PR: https://github.com/satijalab/seurat/pull/5079

The current behavior of FindConservedMarkers is the following:
If a group has greater than number of ` min.cells.group` but less than 3 (the default for `min.cells.group` in function `FindMarkers()`) you get an error from ValidateCellGroups within FindMarkers because it is less than the threshold. 

This is because the parameter `min.cells.group`, when used in `FindConservedMarkers`, was not passed down to FindMarkers

The error is as follows:

```

Error in `map()`:
ℹ In index: 4.
Caused by error in `ValidateCellGroups()`:
! Cell group 1 has fewer than 3 cells
Backtrace:
  1. ... 
 11. Seurat::FindConservedMarkers(...)
 13. Seurat:::FindMarkers.Seurat(...)
 15. Seurat:::FindMarkers.StdAssay(...)
 17. Seurat:::FindMarkers.default(...)
 18. Seurat:::ValidateCellGroups(...)
 19. base::stop("Cell group 1 has fewer than ", min.cells.group, " cells")
 
```

# Implementation

I added the parameter to min.cells.group the FindMarkers() function that is run within FindConservedMarkers() so that that setting is passed down

### DIFF
--- a/R/differential_expression.R
+++ b/R/differential_expression.R
@@ -369,6 +369,7 @@ FindConservedMarkers <- function(
       ident.1 = ident.use.1,
       ident.2 = ident.use.2,
       verbose = verbose,
+      min.cells.group =   min.cells.group,
       ...
     )
     names(x = marker.test)[i] <- levels.split[i]


anashen (approved):
Thank you @hamelyz for the PR!

Issue #8391, opened 2024-01-29
address 0x2b1b5e393b48, cause 'memory not mapped' error


basetwin:
My data is scRNA-seq data of 2 million cells. As data is so large, I merged representative scRNA-seq sample data in python (scanpy), and then I converted it into R seurat object using BPCells. After that, I processed my seurat object (Normalize, FindVariableFeatures, ScaleData). These steps (normalize ~scale) do not raise errors but in RunPCA step, the error arises.

Warning: Converting to a dense matrix may use excessive memory
This message is displayed once every 8 hours.

 *** caught segfault ***
address 0x2b780a393b48, cause 'memory not mapped'

Traceback:
 1: build_csparse_matrix_double_cpp(iter)
 2: asMethod(object)
 3: as(from, "dgCMatrix")
 4: as.matrix(.)
 5: as(from, "dgCMatrix") %>% as.matrix()
 6: asMethod(object)
 7: as(x, "matrix")
 8: as.matrix.IterableMatrix(X)
 9: as.matrix(X)
10: apply(X = data.use, MARGIN = 1L, FUN = var)
11: PrepDR5(object = object, features = features, layer = layer,     verbose = verbose)
12: RunPCA.StdAssay(object = object[[assay]], assay = assay, features = features,     npcs = npcs, rev.pca = rev.pca, weight.by.var = weight.by.var,     verbose = verbose, ndims.print = ndims.print, nfeatures.print = nfeatures.print,     reduction.key = reduction.key, seed.use = seed.use, ...)
13: RunPCA(object = object[[assay]], assay = assay, features = features,     npcs = npcs, rev.pca = rev.pca, weight.by.var = weight.by.var,     verbose = verbose, ndims.print = ndims.print, nfeatures.print = nfeatures.print,     reduction.key = reduction.key, seed.use = seed.use, ...)
14: RunPCA.Seurat(obj, npcs = 30)
15: RunPCA(obj, npcs = 30)

how can I deal with this error? or there are any other ways to input extracted reduction information into seurat object?
I have also converted preprocessed data (from scanpy, from normalization to runumap) using seuratdisk, but this step also raises errors.
Loading required package: reticulate
Warning: Unknown file type: h5ad
Warning: 'assay' not set, setting to 'RNA'
Creating h5Seurat file for version 3.1.5.9900
Adding X as data
Adding X as counts
Adding meta.features from var
Validating h5Seurat file
[...]

PR #8966, opened 2024-06-01
PrepDR5 with BPCells remove dense matrix conversion

bnprks:
I think this is the likely culprit for issue #8391.

The problem here is that `PrepDR5` has a check to remove features which have the same value for every cell (i.e. have zero variance). Unfortunately, the way it does this causes BPCells matrices to be converted to dense matrices in memory. This change instead uses the variance function provided by BPCells which can operate fully disk-backed.

This means that RunPCA effectively does not get any memory-usage benefit from using BPCells, since PrepDR5 will unnecessarily convert the whole scale.data matrix in-memory. 

With this fix, the RunPCA function should be able to use consistently low memory for BPCells-enabled Seurat objects.

My guess is that the crashes from issue #8391 are people with large datasets running out of memory due to this small oversight in PrepDR5.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Seurat
-Version: 5.1.0.9001
-Date: 2024-06-10
+Version: 5.1.0.9002
+Date: 2024-07-02
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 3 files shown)


dcollins15:
Thanks for the fix @bnprks everything LGTM! 

As we discussed it's _conceivable_ that this change could affect some of the PCA values produced for `IterableMatrix` instances but, having compared the output of:

```
apply(counts, MARGIN = 1L, FUN = var)
```

and

```
BPCells::matrix_stats(
  as(counts, Class = "IterableMatrix"), 
  row_stats="variance"
)$row_stats["variance",]
```

I'm satisfied that any variation is within acceptable parameters and shouldn't cause any noticeable changes to downstream analysis 👌 

If there are no objections, I'm going to:

1. Rebase onto `develop`
2. Update the `NEWS.md` file
3. Update the `DESCRIPTION` file

and then we can merge away 🚀 

Thanks again!

bnprks:
Of course, feel free to make whatever edits you need to prep your merge -- looking forward to having this fix make it into a Seurat release!

dcollins15 (approved):

PR #8298, opened 2024-01-10
small fix for `ReadXenium`

alikhuseynov:
Hi, this is just a quick fix for `ReadXenium` to always return dataframes when `sapply` is used.
..see issue #8265 
thanks

Issue #8265, opened 2024-01-04
Error with LoadXenium


SarahE97:
I ran 2 different xenium runs and when trying to use the LoadXenium function to create a seurat object, one of them works great, the other responds with this error: 
> xenium.obj <- LoadXenium(path, fov = "fov")
10X data contains more than one type and is being returned as a list containing matrices of each type.
|--------------------------------------------------|
|==================================================|
Error in if (ncol(x = coords) >= 3) { : argument is of length zero

Any ideas where I can start troubleshooting it? Is it possibly a data issue? 
Thanks

zskylarli:
Hi! I'm unable to reproduce your error without the object - could you copy/paste your `traceback()` here? Additionally, have you checked that the required files are nested in the folder that your path points to, especially `cells.csv.gz` and `cell_boundaries.csv.gz`?
?

SarahE97:
When I run traceback this is the result: > xenium.obj <- LoadXenium(path, fov = "fov")
10X data contains more than one type and is being returned as a list containing matrices of each type.
|--------------------------------------------------|
|==================================================|
Error in if (ncol(x = coords) >= 3) { : argument is of length zero
> traceback()
3: CreateCentroids.default(data$centroids)
2: CreateCentroids(data$centroids)
1: LoadXenium(path, fov = "fov")

Here's the list of present files: 
> list.files(path=path, pattern=NULL, all.files=FALSE, full.names=FALSE)
 [1] "analysis"                          "analysis.zarr.zip"                
 [3] "analysis_summary.html"             "aux_outputs"                      
 [5] "cell_boundaries.csv.gz"            "cell_boundaries.parquet"          
 [7] "cell_feature_matrix"               "cell_feature_matrix.h5"           
 [9] "cell_feature_matrix.zarr.zip"      "cells.csv.gz"                     
[11] "cells.parquet"                     "cells.zarr.zip"                   
[13] "experiment.xenium"                 "gene_panel.json"                  
[15] "Men_MSC1_Meninges_cells_stats.csv" "Men_MSC1_Meninges_coordinates.csv"
[17] "metrics_summary.csv"               "morphology.ome.tif"               
[19] "morphology_focus.ome.tif"          "morphology_mip.ome.tif"           
[21] "nucleus_boundaries.csv.gz"         "nucleus_boundaries.parquet"       
[23] "transcripts.csv.gz"   "transcripts.parquet"               
[25]             "transcripts.zarr.zip"

[...]

PR #9135, opened 2024-07-22
Fix/load xenium

dcollins15:
Updates `LoadXenium` to accommodate datasets without "Blank Codeword" or "Unassigned Codeword" matrices. Relates to:
- https://github.com/satijalab/seurat/issues/8265

Cherry-picked from:
- https://github.com/satijalab/seurat/pull/8298

Big thanks to @jsicherman for implementing the fix 🙏

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.1.0.9003
+Version: 5.1.0.9004
 Date: 2024-07-22
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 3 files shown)


jsicherman:
This should probably also cherry-pick https://github.com/satijalab/seurat/pull/8298/commits/22caa141de2357b64809f791a3c66807ea40e96e from @alikhuseynov whose PR I piggy-backed on before (I think that is to support users on an old version of R where the default for `simplify` was different)

dcollins15:
> This should probably also cherry-pick https://github.com/satijalab/seurat/commit/22caa141de2357b64809f791a3c66807ea40e96e from @alikhuseynov whose PR I piggy-backed on before (I think that is to support users on an old version of R where the default for simplify was different)

That commit has already been merged 👌 see #9132

rsatija (approved):

Issue #9346, opened 2024-09-29
FoldChange results depend on the group size


roi-meir:
Hi

Seurat recently changed the pseudocount used in the log fold change computation. Instead of using a fixed pseudocount, it is divided by the group size. This can cause a a different log fold change that depends on the group's size even if the mean is the same.
The log fold change results will prefer the smaller group because it's pseudocount will be smaller 

The code appears in multiple locations but for example from [differential_expression.R](https://github.com/satijalab/seurat/blob/master/R/differential_expression.R#L1073)
```r
counts.mean.fxn <- function(x) {
    # return(log(x = rowMeans(x = x) + pseudocount.use, base = base))
    return(log(x = (rowSums(x = x) + pseudocount.use)/NCOL(x), base = base))
  }
```

I think that we should add a small constant epsilon as a pseudocount (as done in Scanpy)

To reproduce it I created a small dataset with equal expression in all cells and got log fold change depending on the group size. 

```r
library(dplyr)
library(Seurat)

obj <- CreateSeuratObject(data.frame(x=rep(0.1, 10), y=rep(1, 10), z=rep(10, 10), row.names = 1:10) %>% t)
Idents(obj) <- colnames(obj)
message("5 vs 2")
FoldChange(obj, ident.1 = 1:5, ident.2 = 6:7, slot = "counts")
message("5 vs 4")
FoldChange(obj, ident.1 = 1:5, ident.2 = 6:9, slot = "counts")
message("5 vs 5")
FoldChange(obj, ident.1 = 1:5, ident.2 = 6:10, slot = "counts")
```

Output 
```
5 vs 2
   avg_log2FC pct.1 pct.2
x -1.00000000     1     1
y -0.32192809     1     1
z -0.04182018     1     1
5 vs 4
    avg_log2FC pct.1 pct.2
x -0.222392421     1     1
y -0.058893689     1     1
z -0.007054758     1     1
5 vs 5
  avg_log2FC pct.1 pct.2
x          0     1     1
y          0     1     1
z          0     1     1
```

sessionInfo:
```
R version 4.3.1 (2023-06-16)
Platform: x86_64-conda-linux-gnu (64-bit)
Running under: Gentoo/Linux

Matrix products: default
BLAS/LAPACK: ~/miniconda/envs/test_env/lib/libopenblasp-r0.3.21.so;  LAPACK version 3.9.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C
[...]

PR #9348, opened 2024-09-30
Add a constant pseudocount independent of group size

roi-meir:
# Problem

Fixing issue #9346. Currently, the supplied epsilon in FoldChange and FindMarkers is divided by the group size, resulting in higher lof fold change when one group is larger than the other. 

# Implementation
This fix adds a constant pseudocount and changes the default value to 1e-6 to ensure a small pseudocount.

This fix isn't backward compatible with code that supplied a high pseudocount. I'm not sure what are the guidelines for such cases. If you think the fix should be different I would be happy to change the implementation.


cyrillustan:
I can agree with your change of taking the average first and then taking the logarithm, but why do you want to change the pseudocount to 1e-06? expm1() and log1p() are inverse functions. On a linear scale, having a pseudocount of 1e-06 and 1e-10 does not make a difference, but on a logscale it actually changes the values greatly.

roi-meir:
I didn't change the order of the operations. The current code also takes the mean first and then the logarithm. I only meant to change the value of the pseudocount. 
The current code first take the expm1 to revert the normalization, compute the mean and then add a pseudocount which is 1/NCOL(x). I changed it to add a fixed small pseudocount but I still add it before the log

I can rewrite the current expression to get the exact results 
```r
log( rowSums(expm1(x)) + pseudocount.use)/NCOL(x)
# Equal to
log(rowMeans(expm1(x)) + pseudocount.use/NCOL(x))
```

Because the psuedocount is divided by the number of samples in the group, we would get different log expression to group of different sizes

cyrillustan:
Thank you for this very comprehensive PR! We are impressed by the dedication and the level of work here! As I posted in issue #9346, after some discussion in the group, we decided not to change this for now. The reasoning was posted there. We appreciate you bringing our attention to this issue!

Issue #9436, opened 2024-10-28
`DimPlot` convenience functions infer arguments from the call stack


rharao:
The convenience functions for `DimPlot` through `SpecificDimPlot` (namely, `PCAPlot`, `TSNEPlot`, and `UMAPPlot`) rely on false assumptions.

_Example_: Try running `Seurat::UMAPPlot` (specifying the namespace rather than just getting `UMAPPlot` from the search list) on a Seurat object that has a `"umap"` reduction. One should expect the function to access this with `.[["umap"]]`.

```
Error in object[[reduction]] : 
  'seurat::umap' not found in this Seurat object
```

It's trying to find `.[["seurat::umap"]]` because, instead of `UMAPPlot` just calling `DimPlot` with `reduction = "umap"`, `SpecificDimPlot` is doing **string manipulations on the call stack** to guess which particular convenience function was called.

Existing code for `UMAPPlot`, `TSNEPlot`, and `PCAPlot` are identical down to the bytecode: they pass their `object` and dots arguments, unmodified, to the non-exported `SpecificDimPlot`, and return the result, unmodified. There is **no logic** on either side and they are bare aliases for the unexposed function; their only purpose is to show up on the call stack.

`SpecificDimPlot` as it is (comments mine):
``` r
function (object, ...) 
{
    funs <- sys.calls()
    # Get the name of the preceding function on the call stack as a string
    name <- as.character(x = funs[[length(x = funs) - 1]])[1]
    # Remove the substring "Plot", if it's there, and convert to lowercase
    # This string is the guessed reduction name for now
    name <- tolower(x = gsub(pattern = "Plot", replacement = "", 
        x = name))
    # Put the input args in a list
    args <- list(object = object)
    args <- c(args, list(...))
    # Tentative better guess (this is undocumented):
    # Search in the names of the object for any string that has _both_
    # the guessed reduction name (case insensitive) _and_
    # the object's default assay name (case sensitive) as substrings
    reduc <- grep(pattern = name, x = names(x = object), value = TRUE, 
        ignore.case = TRUE)
    reduc <- grep(pattern = DefaultAssay(object = object), x = reduc, 
        value = TRUE)
    # If zero or multiple of those strings were found, just use the first guess
    # Update the args list with the guess
    args$reduction <- ifelse(test = length(x = reduc) == 1, yes = reduc, 
        no = name)
    # Now pass the whole thing on to `DimPlot` and hope the guess was right
[...]

PR #9493, opened 2024-11-18
Fix DimPlot convenience function logic (#9436)

rharao:
Currently `PCAPlot`, `UMAPPlot`, and `TSNEPlot` all do nothing but call `SpecificDimPlot`, which does **string manipulation on the call stack** to see which one of those was called and then guess the correct reduction to dispatch to `DimPlot`. See #9436 for examples of how this causes problems.

I am open to feedback on this.


mccormickceci:
Thank you for opening this thoughtful PR! I think that the code you have written would be a strong workaround to avoid some of the call stack interpretation. At this time, however, we are no longer going to be making changes to these convenience functions and its helper SpecificDimPlot() so I will be closing this PR.

rharao:
Does this mean those functions will be deprecated?

mccormickceci:
There is not a plan as of currently to deprecate the functions! Just means that we will not be making edits to them in the code base.

PR #9541, opened 2024-12-10
feature: updates for cellranger 9.0

stephenwilliams22:
This PR adds support for cellranger 9.0 which performs automatic cell type annotation. Creates a simple loading function similar to `Load10x_spatial` for 10x single cell data which will also put cell annotations in the `meta.data`.


stephenwilliams22:
> Thanks for this! I'm excited to try the automatic cell type calling in Cellranger 9.0. I'm not part of the developer team, but would you be open to some comments on this addition from a user's perspective?

100%. Comments welcome.

NAMESPACE
```diff
@@ -301,6 +301,7 @@ export(LeverageScore)
 export(LinkedDimPlot)
 export(LinkedFeaturePlot)
 export(Load10X_Spatial)
+export(Load10x)
```
rharao: I understand this might be contrary to official brand guidelines, but other functions in this package have consistently used an uppercase "X" in "10X": `Read10X`, `Read10X_h5`, `Load10X_Spatial`, etc. In the interest of uniformity of what the user sees, I would prefer to keep the uppercase "X" for this function.

R/preprocessing.R
```diff
@@ -646,7 +646,70 @@ Load10X_Spatial <- function (
...
+#' Load a 10x Genomics Single Cell Experiment into a \code{Seurat} object
+#'
+#' @inheritParams Read10X
+#' @inheritParams SeuratObject::CreateSeuratObject
+#' @param data.dir Directory containing the H5 file specified by \code{filename}
+#' and the image data in a subdirectory called \code{spatial}
+#' @param filename Name of H5 file containing the feature barcode matrix
+#' @param filter.matrix Only keep spots that have been determined to be over
+#' tissue
+#' @param to.upper Converts all feature names to upper case. This can provide an
```
rharao, line 700: (Preemptive-- this parameter doesn't seem to be implemented yet)
In my opinion, it would be better not to suggest that blind uppercasing is ever an appropriate way of finding mouse-human orthologs, even in an exploratory analysis. 
If the user has some good reason to convert all the symbols to uppercase, they can simply use `toupper` and replace the rownames of the object.

R/preprocessing.R
```diff
@@ -646,7 +646,70 @@ Load10X_Spatial <- function (
...
+
+  filename <- list.files(data.dir, filename, full.names = FALSE, recursive = FALSE)
+  counts.path <- file.path(data.dir, filename)
+  if (!file.exists(counts.path)) {
+    stop(paste0("File not found: '", counts.path, "'"))
+  }
 
+  counts <- Read10X_h5(counts.path, ...)
+  if (is.list(counts)) {
+    object <- Reduce(function(x, y) merge(x, y, add.cell.ids = names(counts)),
```
rharao: You merge the objects two at a time in the anonymous function here, but the length of `names(counts)` is the length of `counts`, which could be greater than two.

`SeuratObject:::merge.Seurat` does recurse if you give it a single Seurat object in the first argument and a list of Seurat objects in the second. A common construction is to do something like `merge(x[[1]], x[-1], add.cell.ids = names(x)` on a list of Seurat objects `x`.

R/preprocessing.R
```diff
@@ -646,7 +646,70 @@ Load10X_Spatial <- function (
...
+                     lapply(names(counts), function(name) CreateSeuratObject(counts[[name]], assay = name)))
+  } else {
+    object <- CreateSeuratObject(counts, assay = assay)
+  }
+
+  cell_types_path <- file.path(data.dir, "cell_types", "cell_types.csv")
+  if (file.exists(cell_types_path)) {
+    cell.types <- read.csv(cell_types_path)
+    object@meta.data <- dplyr::left_join(tibble::rownames_to_column(object@meta.data, "barcode"),
```
rharao: I often do exactly this in my analysis scripts (barcodes to column, left join metadata, barcodes to rownames), but using the dplyr left join here introduces another dependency. Not a problem for me, but some other people may want to avoid this and use `base::merge`

R/preprocessing.R
```diff
@@ -646,7 +646,70 @@ Load10X_Spatial <- function (
...
+#'
+#'
+#' @export
+#' @concept preprocessing
+#'
+#' @examples
+#' \dontrun{
+#' data_dir <- 'path/to/data/directory'
+#' list.files(data_dir) # Should show filtered_feature_bc_matrix.h5
+#' Load10X(data.dir = data_dir)
```
rharao, line 716: Note uppercase "X" here

R/preprocessing.R
```diff
@@ -646,7 +646,70 @@ Load10X_Spatial <- function (
...
 }
 
+#' Load a 10x Genomics Single Cell Experiment into a \code{Seurat} object
+#'
+#' @inheritParams Read10X
+#' @inheritParams SeuratObject::CreateSeuratObject
+#' @param data.dir Directory containing the H5 file specified by \code{filename}
+#' and the image data in a subdirectory called \code{spatial}
+#' @param filename Name of H5 file containing the feature barcode matrix
+#' @param filter.matrix Only keep spots that have been determined to be over
```
rharao: (Preemptive-- this parameter doesn't seem to be implemented yet)
This sense of "filtering the matrix" is different than the more common one, which is removing barcodes from the counts matrix corresponding to empty and low-read GEMs. A user might have the misconception that they could set this option while providing the raw matrix to get the filtered matrix in the Seurat object.
To avoid confusion, how about calling this parameter something like `tissue.spots.only`?

rharao (commented):
Appreciate it!

NAMESPACE
stephenwilliams22: Yup, good catch

R/preprocessing.R
stephenwilliams22: This was carry over from using `Load10X_Spatial` as a baseline to build this function. I think simply removing it is the best coarse of action

R/preprocessing.R
stephenwilliams22: This option was put in originally by the seurat team. I'll let them address the usefulness here.

R/preprocessing.R
rharao: Oh, that makes sense. Thanks for clearing that up

R/preprocessing.R
rharao: ... or `AddMetaData`: you could do something like `object <- AddMetaData(object, tibble::column_to_rownames(cell.types, "barcode"))` instead of the left join

stephenwilliams22:
@dcollins15 sorry for the little formatting edits. Not really sure where those were introduced. Feel free to fix and roxygenise and push directly to my branch as it seems what I'm going on my side is introducing little diffs.

stephenwilliams22:
@evolvedmicrobe I wanted to give you a heads up on this PR

R/preprocessing.R
```diff
@@ -645,6 +645,133 @@ Load10X_Spatial <- function (
...
+  if (is.list(counts)) {
+    seurat.list <- lapply(names(counts), function(name) {
+      CreateSeuratObject(
+        counts = counts[[name]],
+        assay = name,
+        project = name
+      )
+    })
+
+    for (i in 1:seq_along(seurat.list)) {
```
rharao: > `1:seq_along(seurat.list)`

should be `seq_along(seurat.list)`.
Alternatively, you could replace the for loop with:
```r
seurat.list <- lapply(seurat.list, \(x) if (Assays(x) %in% c("Gene Expression", "RNA")) {Add_10X_CellTypes(data.dir, x)} else {x})
```

rharao: I did also notice that if `Add_10X_CellTypes` doesn't find a cell types csv, it returns the original without warning or error, so if you want to rely on that, you could skip the conditional here.

R/preprocessing.R
```diff
@@ -645,6 +645,133 @@ Load10X_Spatial <- function (
...
+  filename <- list.files(data.dir, filename, full.names = FALSE, recursive = FALSE)
+  counts.path <- file.path(data.dir, filename)
+  if (!file.exists(counts.path)) {
+    stop(paste0("File not found: '", counts.path, "'"))
+  }
+
+  counts <- Read10X_h5(counts.path, ...)
+
+  if (to.upper) {
+    counts <- imap(counts, ~{
```
rharao: Nit: because this formula operates on the counts matrices only and doesn't use the list's names, I think the `imap` call could be replaced by `base::sapply`.
Frankly, I'm not sure why this package imports `purrr::imap` anyway, as I can't find it used anywhere in the codebase, not even in `Load10X_Spatial`. But that's an issue for another time.

R/preprocessing.R
```diff
@@ -645,6 +645,133 @@ Load10X_Spatial <- function (
...
+  }
+  if (!file.exists(data.dir)) {
+    stop(paste0("No such file or directory: '", data.dir, "'"))
+  }
+
+
+  filename <- list.files(data.dir, filename, full.names = FALSE, recursive = FALSE)
+  counts.path <- file.path(data.dir, filename)
+  if (!file.exists(counts.path)) {
+    stop(paste0("File not found: '", counts.path, "'"))
```
rharao: Nit: for readability, the `paste0` can be removed, as `stop` will concatenate all its dots arguments with no separator.

R/preprocessing.R
```diff
@@ -645,6 +645,133 @@ Load10X_Spatial <- function (
...
+    return(object)
+  } else {
+    return(object)
+  }
+}
+
+#' Load a 10x Genomics Single Cell Experiment into a \code{Seurat} object
+#'
+#' @inheritParams Read10X
+#' @inheritParams SeuratObject::CreateSeuratObject
```
rharao, line 696: Might be worth noting that the supplied `assay` argument won't be used if the h5 is multimodal and returns a list.

R/preprocessing.R
```diff
@@ -645,6 +645,133 @@ Load10X_Spatial <- function (
...
+#' @param data.dir Directory containing the H5 file specified by \code{filename}
+#' and the image data in a subdirectory called \code{spatial}
```
rharao: Carryover from `Load10X_Spatial`, doesn't apply here

rharao (commented):
I like it!

R/preprocessing.R
stephenwilliams22: personally, I like the simplicity and readability of tidy functions. I'll leave it up to the seruat developers to decide if they mind the extra import. I haven't heard any complaints since we wrote this function a few years ago.

rharao: It's not really an issue; the choice one way or another doesn't affect me. (Personally, I agree; I always prefer the purrr functions in my own scripts.) I only wanted to mention it for the attention of the dev team.
Thanks for your receptiveness :smile:

R/preprocessing.R
stephenwilliams22: will make a note about the assay. ultimately we don't return a list but a single seurat object so I don't want to add so much detail as to be confusing.

R/preprocessing.R
dcollins15: +1 to replacing the `imap` call with a `sapply` to match the current implementation of `Load10X_Spatial`.

> Frankly, I'm not sure why this package imports purrr::imap anyway, as I can't find it used anywhere in the codebase, not even in Load10X_Spatial. But that's an issue for another time.

This is also a good point, `@importFrom purrr imap` _should_ be dropped from `Load10X_Spatial` and as a suggested dependency. Now that you mention it, `@importFrom grid rasterGrob` could also be dropped from the function. I'll include these tweaks in #9556 👌

R/preprocessing.R
dcollins15: The parameter was introduced well before my time but I tend to agree with @rharao.

R/preprocessing.R
```diff
@@ -645,6 +645,138 @@ Load10X_Spatial <- function (
...
+      add.cell.ids = names(counts),
+      merge.data = FALSE
+    )
+    return(merged.object)
+
+  } else {
+    object <- CreateSeuratObject(counts, assay = assay)
+    if (Assays(object)  %in%  c("Gene Expression", "RNA")) {
+      object <- Add_10X_CellTypes(data.dir, object)
+    }
```
dcollins15, line 774: I'm not sure that `is.list(counts)` conditional is necessary. Instead, you can just force `counts` to always be a list, and then you can do without the `else` block. 

It looks like there's just one `CellTypes` file per `data.dir`? If that's the case, this could be simplified even more—the call to `Add_10X_CellTypes` could be made on the `merged.object`.

dcollins15 (commented):
This is looking great 🎉 Big thanks to @rharao for the code review 🙌 

For these types of loader functions, I think the two main priorities should be:
1. Backward compatibility
2. Consistency

You've done a great job emulating the existing `Load10X_Spatial` but I wonder if the two functions can or should share more logic 🤔

I’d also appreciate a bit more clarity on how Load10X is intended to relate to Read10X. Specifically:
- What versions of Cell Ranger output are compatible with each?
- As far as I can tell, Read10X hasn't been updated since just before the release of Cell Ranger v7.1.0.

I suspect that we could be doing a better job of making `Load10X_Spatial` and `LoadXenium` more mutually coherent but I'm not very familiar with the datatypes or output structure to propose specific suggestions. 

On a related note, what's the easiest way to access 10x datasets in their "canonical" structure? I typically download datasets from the [10x Portal](https://www.10xgenomics.com/datasets) but end up having to manually configure the `data.dir` for `Seurat`. I assume there’s a better method that I’m just ignorant of. 

This leads nicely into my other concern: testing. Creating small, representative datasets for testing was one of the more tedious parts of the Visium HD updates. Any thoughts on streamlining this process would be great.

---

@stephenwilliams22, now that I’ve had a thorough look, I think it would be best to hold off on merging this until after the next release (v5.2.0).

R/preprocessing.R
```diff
@@ -645,6 +645,138 @@ Load10X_Spatial <- function (
...
+#' @return A Seurat object with updated metadata including cell type annotations if the file is found.
+#'
+#' @details
+#' The function searches for a CSV file named "cell_types.csv" in the "cell_types" subdirectory within `data.dir`.
+#' The CSV file should contain at least a "barcode" column that matches the cell barcodes in the Seurat object.
+#' Additional columns in the CSV file will be merged into the Seurat object's metadata.
+#'
+#' @importFrom utils read.csv
+#' @importFrom tibble rownames_to_column column_to_rownames
+#' @importFrom base file.path file.exists merge
```
anashen: (Roxygen) Imports from base need to be removed. A warning is thrown upon loading the package:`In asNamespace(ns, base.OK = FALSE) : operation not allowed on base namespace`.
```suggestion
```

R/preprocessing.R
```diff
@@ -645,6 +645,138 @@ Load10X_Spatial <- function (
...
+  } else {
+    return(object)
+  }
+}
+
+#' Load a 10x Genomics Single Cell Experiment into a \code{Seurat} object
+#'
+#' @inheritParams Read10X
+#' @inheritParams SeuratObject::CreateSeuratObject If multiome 10x data the
+#' assay param will not be used. The names of each assay contained in the matrix are used.
```
anashen: (Roxygen) The detail "If multiome 10x data..." needs to be part of a `@note` tag on its own line, otherwise the function to inherit params from, `SeuratObject::CreateSeuratObject`, will not be recognized by roxygen2.
```suggestion
#' @inheritParams SeuratObject::CreateSeuratObject 
#' @note If multiome 10x data the assay param will not be used. The names of each assay contained in the matrix are used.
```

R/preprocessing.R
anashen: I do like `imap` for its readability, but here I agree with @/rharao that because the names aren't used, it doesn't make much of a difference. Not to mention, `purrr` has since been dropped from package imports.

R/preprocessing.R
stephenwilliams22: I think I'd actually prefer to keep the `is.list(counts)` because it keep the merging logic later in the functions simplified. To me it makes things a bit more clear because counts that are lists must have more than one feature type

anashen: That makes sense, I'm in favor of keeping it as-is. Resolving this thread.

stephenwilliams22:
> This is looking great 🎉 Big thanks to @rharao for the code review 🙌
> 
> For these types of loader functions, I think the two main priorities should be:
> 
> 1. Backward compatibility
> 2. Consistency
> 
> You've done a great job emulating the existing `Load10X_Spatial` but I wonder if the two functions can or should share more logic 🤔
> 
> I’d also appreciate a bit more clarity on how Load10X is intended to relate to Read10X. Specifically:
> 
> * What versions of Cell Ranger output are compatible with each?
> * As far as I can tell, Read10X hasn't been updated since just before the release of Cell Ranger v7.1.0.
> 
> I suspect that we could be doing a better job of making `Load10X_Spatial` and `LoadXenium` more ~mutually coherent~ cohesive but I'm not very familiar with the datatypes or output structure to propose specific suggestions.
> 
> On a related note, what's the easiest way to access 10x datasets in their "canonical" structure? I typically download datasets from the [10x Portal](https://www.10xgenomics.com/datasets) but end up having to manually configure the `data.dir` for `Seurat`. I assume there’s a better method that I’m just ignorant of.
> 
> This leads nicely into my other concern: testing. Creating small, representative datasets for testing was one of the more tedious parts of the Visium HD updates. Any thoughts on streamlining this process would be great.
> 
> @stephenwilliams22, now that I’ve had a thorough look, I think it would be best to hold off on merging this until after the next release (v5.2.0).

I agree with most of this. We can definitely take the time to share code between all the 10x functions. Please let me know if you want to do that for this release.

As far as the specific code here 
- What versions of Cell Ranger output are compatible with each?
  - Both functions should be compatible will all versions of cellranger. The differences are
    - Read10x will not put cell types in the metadata
- `Read10X`
  -   As you mentioned, this is a very old function which doesn't take advantage of the "packaged" h5 matrix. Personally, I think that we might think of an end of life for Read10X. Not that 10x is going to get rid of the `filtered_feature_bc_matrix` directory any times soon but new features will most certainly be focused on the h5 matrix
  
With regard to testing datasets we are in talks internally about this. I'm going to try and make each release of the rangers come with a tiny dataset with the correct structure for testing.

Datasets on our public data portal should be in the structure of any normal ranger run. if they are not I'd like to hear about the issue so we can provide instruction or correct things on our side.

Issue #8283, opened 2024-01-09
Missing pkg_resources attribute from umap-learn (v 0.5.5)


fspecque:
Hi !

I've noticed a bug when computing UMAP using a `Graph` (thus, using `umap-learn`). See the code below to reproduce. *Please note that the version of the `umap-learn` python module I'm using is `0.5.5` and I've installed it using conda ([conda-forge](https://anaconda.org/conda-forge/umap-learn))*
```r
library(Seurat)

data(pbmc_small)

pbmc_small <- FindVariableFeatures(pbmc_small, verbose = F)
pbmc_small <- ScaleData(pbmc_small, verbose = F)
pbmc_small <- RunPCA(pbmc_small, verbose = F)
pbmc_small <- FindNeighbors(pbmc_small, verbose = F)
pbmc_small <- RunUMAP(pbmc_small, graph = "RNA_snn", umap.method = "umap-learn")

# Error in py_get_attr_impl(x, name, silent) : 
#   AttributeError: module 'umap' has no attribute 'pkg_resources'
# Run `reticulate::py_last_error()` for details.
```
Here's the output of `reticulate::py_last_error()`
```r
# ── Python Exception Message ───────────────────────────────────────────────────────────────────────
# AttributeError: module 'umap' has no attribute 'pkg_resources'

# ── R Traceback ────────────────────────────────────────────────────────────────────────────────────
#      ▆
#   1. ├─Seurat::RunUMAP(pbmc_small, graph = "RNA_snn", umap.method = "umap-learn")
#   2. └─Seurat:::RunUMAP.Seurat(pbmc_small, graph = "RNA_snn", umap.method = "umap-learn")
#   3.   ├─Seurat::RunUMAP(...)
#   4.   └─Seurat:::RunUMAP.Graph(...)
#   5.     ├─base::numeric_version(x = umap$pkg_resources$get_distribution("umap-learn")$version)
#   6.     │ └─base::.make_numeric_version(x, strict, .standard_regexps()$valid_numeric_version)
#   7.     ├─umap$pkg_resources
#   8.     └─reticulate:::`$.python.builtin.module`(umap, "pkg_resources")
#   9.       └─reticulate:::`$.python.builtin.object`(x, name)
#  10.         └─reticulate:::py_get_attr_or_item(x, name, TRUE)
#  11.           └─reticulate::py_get_attr(x, name)
#  12.             └─reticulate:::py_get_attr_impl(x, name, silent)
```

[...]

PR #8294, opened 2024-01-10
Fix umap-learn version check

fspecque:
umap-learn >= 0.5.5 misses the `pkg_resources` attribute. This PR fixes the problem when checking `umap-learn`'s version and is compatible with `umap-learn` v 0.5.5 and lower.

fix  #8283

PR #9559, opened 2024-12-18
Fix/umap-learn version check

dcollins15:
Reopening #8294 after accidentally clobbering the source branch 🤦 

Please find @fspecque's original description below:

---

umap-learn >= 0.5.5 misses the pkg_resources attribute. This PR fixes the problem when checking umap-learn's version and is compatible with umap-learn v 0.5.5 and lower.

fix https://github.com/satijalab/seurat/issues/8283

---


mhkowalski (approved):
Looks good!

Issue #9483, opened 2024-11-14
Issue in "Error in object[[assay]][] <- grp.cors: object of type 'S4' is not subsettable" as shown in spatial_vignette.Rmd because new Seurat assay required subset method


joesphbeller:
#8125 and #8809 both touched this topic, but it appears that the problem still persists. It was encouraged in #8125 to open a new issue if it persisted. I made a comment that included a possible solution, but I did not receive a reply. 

Here is the comment I made that is still applicable:
> Hello,
> 
> I am currently encountering the same error on the Vignette with `Seurat` v5.1.0, `SeuratObject` v5.0.2, and `Matrix` v1.7-0.
> 
> `brain <- GroupCorrelation(brain, group.assay="Spatial", assay="Spatial",slot="data", do.plot = FALSE)` yields: `Error in object[[assay]][] <- grp.cors : object of type 'S4' is not subsettable`
> 
> I then found the definition for GroupCorrelation() in utilities.R as well as the **docstring** which states: `@return A Seurat object with the correlation stored in metafeatures`
> 
> This is when I came to the same realization as @shihsama that metafeatures was not an element of the assay (in recent versions). Additionally, it appears that the new Seurat assay can no longer be subsetted this way.
> 
> ### Fixes that work:
> One potential fix is changing this line in GroupCorrelation `object[[assay]][] <- grp.cors` to `try(object[[assay]]@meta.data <- grp.cors, silent = T)` `try(object[[assay]]@meta.features <- grp.cors, silent = T)`
> 
> This will run GroupCorrelation() on the SCT and Spatial Assays successfully and do what the docstring says. However, this change does not fix all issues. When I run the next line: `p1 <- GroupCorrelationPlot(brain, assay = "Spatial", feature.group = "feature.grp", cor = "nCount_Spatial_cor") + ggtitle("Log Norm") + theme(plot.title = element_text(hjust =0.5))` yields: `Error in h(simpleError(msg, call)) : error in evaluating the argument 'i' in selecting a method for function '[': invalid 'type' (S4) of argument`
> 
> This error is because the same subset technique is being used to load into the "data" variable in the GroupCorrelationPlot() function. You could change the following line in GroupCorrelationPlot():
> 
> `data <- object[[assay]][c(feature.group, cor)]` to `try(data <- object[[assay]]@meta.data[, c(feature.group, cor)], silent=T)` `try(data <- object[[assay]]@meta.features[, c(feature.group, cor)], silent=T)`
> 
[...]

PR #9625, opened 2025-01-16
Fix GroupCorrelation and GroupCorrelationPlot for Seurat V5 compatibility

joesphbeller:
Discussed in issue #9483

# Problem
In spatial_vignette.Rmd, GroupCorrelation() and GroupCorrelationPlot() would both generate the following error:

```"Error in object[[assay]][] <- grp.cors: object of type 'S4' is not subsettable"```

This is because the correlation was being stored in metafeatures, which is no longer an element of the assay in V5.

# Solution

Insert 2 ```try()``` statements to make it compatible for previous and most recent versions of Seurat

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -986,7 +986,8 @@ GroupCorrelation <- function(
   grp.cors <- as.data.frame(x = grp.cors[which(x = !is.na(x = grp.cors))])
   grp.cors$gene_grp <- gene.grp[rownames(x = grp.cors)]
   colnames(x = grp.cors) <- c(paste0(var, "_cor"), "feature.grp")
-  object[[assay]][] <- grp.cors
+  try(object[[assay]]@meta.data <- grp.cors, silent = T)
+  try(object[[assay]]@meta.features <- grp.cors, silent = T)
   if (isTRUE(x = do.plot)) {
     print(GroupCorrelationPlot(
       object = object,
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -986,8 +986,7 @@ GroupCorrelation <- function(
   grp.cors <- as.data.frame(x = grp.cors[which(x = !is.na(x = grp.cors))])
   grp.cors$gene_grp <- gene.grp[rownames(x = grp.cors)]
   colnames(x = grp.cors) <- c(paste0(var, "_cor"), "feature.grp")
-  try(object[[assay]]@meta.data <- grp.cors, silent = T)
-  try(object[[assay]]@meta.features <- grp.cors, silent = T)
+  object[[assay]] <- AddMetaData(object[[assay]], grp.cors)
   if (isTRUE(x = do.plot)) {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: R/visualization.R [hunk 1]
```diff
@@ -4700,7 +4700,8 @@ GroupCorrelationPlot <- function(
   cor = "nCount_RNA_cor"
 ) {
   assay <- assay %||% DefaultAssay(object = object)
-  data <- object[[assay]][c(feature.group, cor)]
+  try(data <- object[[assay]]@meta.data[, c(feature.group, cor)], silent=T)
+  try(data <- object[[assay]]@meta.features[, c(feature.group, cor)], silent=T)
```
dcollins15:
```suggestion
  object[[assay]] <- object[[assay]][[c(feature.group, cor)]]
```

R/utilities.R
```diff
@@ -986,7 +986,8 @@ GroupCorrelation <- function(
   grp.cors <- as.data.frame(x = grp.cors[which(x = !is.na(x = grp.cors))])
   grp.cors$gene_grp <- gene.grp[rownames(x = grp.cors)]
   colnames(x = grp.cors) <- c(paste0(var, "_cor"), "feature.grp")
-  object[[assay]][] <- grp.cors
+  try(object[[assay]]@meta.data <- grp.cors, silent = T)
+  try(object[[assay]]@meta.features <- grp.cors, silent = T)
```
dcollins15:
```suggestion
  object[[assay]] <- AddMetaData(object[[assay]], grp.cors)
```

dcollins15 (changes requested):
@joesphbeller. thank you for taking the time and effort to translate your solution into a PR 🙌

> I was not sure if this would cause any other functions that depend on GroupCorrelation or GroupCorrelationPlot to fail

AFAIK the blast radius for these two methods should be pretty small. In any case, your solution works well 🎉 That said, the fix could be refined further, so I’ve suggested a couple of improvements.

Thanks again!

R/visualization.R
joesphbeller: Hi @dcollins15 

Thanks for your suggestion. I did find one small typo:
`object[[assay]] <- object[[assay]][[c(feature.group, cor)]]`
should be
`data <- object[[assay]][[c(feature.group, cor)]]`

I have edited the file to reflect this. Thank you for the suggestions. Let me know if there is anything else I can do to improve this PR.

dcollins15 (approved):
This LGTM 🚀 

I'm going to go ahead and shepherd this one home—I'll be making the following changes:
- Update the base branch for the PR to `main`
- Rebase
- Update the [changelog](https://github.com/satijalab/seurat/blob/1549c8d6ac219508ee451956f0df80ab3507f3b7/NEWS.md)
- Bump the package version (5.2.99.9000 -> 5.2.99.9001)

Thanks again @joesphbeller!

Issue #9656, opened 2025-01-29
Add `raster.dpi`as parameter for `VlnPlot`


lucygarner:
Hi,

It would be great if `raster.dpi` could be added as a parameter for `VlnPlot`. The size of the rasterized points is not modified by the `pt.size` argument.

Best wishes,
Lucy

samuel-marsh:
Hi Lucy,

This is definitely good enhancement to have but it doesn't have an effect on the point size displayed only the dpi for the rastered output layer/image  See first reprex below.  I have created PR to add parameter to `VlnPlot` because this is still good parameter to have when saving rastered images.

In my testing `pt.size` still works for changing the size of rastered points.  See second reprex below.

Best,
Sam

**Changing raster.dpi**  
``` r
library(tidyverse)
#> Warning: package 'lubridate' was built under R version 4.4.1
library(Seurat)
#> Loading required package: SeuratObject
#> Loading required package: sp
#> 'SeuratObject' was built with package 'Matrix' 1.7.0 but the current
#> version is 1.7.1; it is recomended that you reinstall 'SeuratObject' as
#> the ABI for 'Matrix' may have changed
#> 
#> Attaching package: 'SeuratObject'
#> The following objects are masked from 'package:base':
#> 
#>     intersect, t


pbmc <- pbmc3k.SeuratData::pbmc3k.final
pbmc <- UpdateSeuratObject(pbmc)
#> Validating object structure
#> Updating object slots
#> Ensuring keys are in the proper structure
#> Updating matrix keys for DimReduc 'pca'
#> Updating matrix keys for DimReduc 'umap'
#> Warning: Assay RNA changing from Assay to Assay
#> Warning: Graph RNA_nn changing from Graph to Graph
#> Warning: Graph RNA_snn changing from Graph to Graph
#> Warning: DimReduc pca changing from DimReduc to DimReduc
#> Warning: DimReduc umap changing from DimReduc to DimReduc
#> Ensuring keys are in the proper structure
#> Ensuring feature names don't have underscores or pipes
#> Updating slots in RNA
#> Updating slots in RNA_nn
#> Setting default assay of RNA_nn to RNA
#> Updating slots in RNA_snn
#> Setting default assay of RNA_snn to RNA
#> Updating slots in pca
#> Updating slots in umap
#> Setting umap DimReduc to global
#> Setting assay used for NormalizeData.RNA to RNA
#> Setting assay used for FindVariableFeatures.RNA to RNA
[...]

PR #9665, opened 2025-02-04
Raster.dpi vlnplot

samuel-marsh:
Hi Seurat Team,

This is PR addressing #9656 adding the `raster.dpi` parameter to `VlnPlot`.  The issue brought up of point size not changing I cannot replicate (see reprex in issue) but having control of raster dpi is good to have regardless.  

I made this PR to main given deprecation of develop/master branches but let me know if I should change that.

Best,
Sam

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.2.99.9003
+Version: 5.2.99.9004
 Title: Tools for Single Cell Genomics
 Description: A toolkit for quality control, analysis, and exploration of single cell RNA sequencing data. 'Seurat' aims to enable users to identify and interpret sources of heterogeneity from single cell transcriptomic measurements, and to integrate diverse types of single cell data. See Satija R, Farrell J, Gennert D, et al (2015) <doi:10.1038/nbt.3192>, Macosko E, Basu A, Satija R, et al (2015) <doi:10.1016/j.cell.2015.05.002>, Stuart T, Butler A, et al (2019) <doi:10.1016/j.cell.2019.05.031>, and Hao, Hao, et al (2020) <doi:10.1101/2020.10.12.335331> for more details.
 Authors@R: c(
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 ## Changes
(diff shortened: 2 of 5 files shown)


samuel-marsh:
Rebased to updates on main today (2/4)

samuel-marsh:
Hi @dcollins15 

Just wondering if you/team could approve the merging of this PR?

Thanks!
Sam

dcollins15:
Hey @samuel-marsh,

I'm happy to merge this one in—the only blocker is to decide what to do with the `ExIPlot`, `SingleExIPlot`, and `MultiExIPlot` functions. 

`ExIPlot` is not currently being exported so maybe we should drop consider just dropping them. Otherwise, I'm happy to fix up the docstrings to get the `NAMESPACE` fixed up. Assuming that we _do_ want to keep `ExIPlot` around, do we want to extend the raster functionality for `type = "ridge"`?

Any thoughts @rsatija?

---

That last merge was looking a little wonky so I took the liberty of rebasing this PR onto `main` again 👍

samuel-marsh:
Hi @dcollins15,

Thanks for update!  What do you need in terms of fixing the ExIPlot family of functions?  I think should be ok just passing `raster.dpi` through them no?

Don't think tha raster support for RidgePlot is needed because there aren't any points being plotted.  Rasterizing that plot wouldn't gain very much in terms of rendering speed/save file size.

Best,
Sam

dcollins15:
> What do you need in terms of fixing the ExIPlot family of functions?

The [docstring](https://github.com/satijalab/seurat/blob/f8ae979cf2cd6fc51ad70baa47cc8f61f5e6bf94/R/visualization.R#L6667-L6704) for the `ExIPlot` is missing an `@export` tag (and is misformatted), so it is not included in `Seurat's NAMESPACE. Without any context on the intent of having these in addition to `VlnPlot` and `RidgePlot, ` my inclination is to just drop them, but it'd be great if someone could sanity-check my thinking. 

This probably doesn't need to be a blocker for this PR—I'll give folks a little while longer to weigh in. Either way, I'll merge this in sometime next week 🚀

dcollins15 (approved):
Alright, I think I've left this hanging for long enough—we'll punt on the `ExIPlot` question (and the flakiness from `test_read_mtx.R`)...

Thanks @samuel-marsh 🚀

Issue #9709, opened 2025-02-26
Updates to UMAP/uwot function


samuel-marsh:
### Motivation

Hi Seurat Team (@dcollins15 hope you don't mind me tagging you here for a quick look),

Just wanted to post quick issue with couple questions regarding `RunUMAP` and potential enhancements and whether this is something that dev team would be open to PR to add.

### Feature Description

There are 3 components to potential PR that are completely separate from each other so can pick and choose if only some are desired.

1. `uwot::umap` does now seem to support correlation metric so potential remove this warning/error: https://github.com/satijalab/seurat/blob/6d238c4f5394835d52dbc8c60c5b465c63e76c99/R/dimensional_reduction.R#L1459-L1465

2. Add additional explicit parameter to set `n_threads` in uwot implementation.  Could be passed right now with `...` but would be nice to have direct param.

3. Add support for `uwot::umap2` (https://jlmelville.github.io/uwot/reference/umap2.html).  As per the docs for uwot: 
"_This function behaves like umap except with some updated defaults that make it behave more like the Python implementation and which cannot be added to umap without breaking backwards compatibility_".
- My thoughts on implementation here would be to add additional acceptable method for `umap.method` parameter (e.g., `umap.method = "uwot2"`) to allow for use of `uwot::umap2` vs. `uwot::umap`.

If any/all of these sound ok to the team please let me know and I'll be happy to work up a PR.

Thanks as always!
Sam

AtSourav:
Hi @samuel-marsh 

Yes, these look like totally reasonable updates, and no.3 would be especially nice to have. So we would any PRs from you addressing these.

samuel-marsh:
Hi @AtSourav,

Sounds good!  I deleted and reforked the other day so I’ll resubmit PR on this next week!

Best,
Sam

PR #9752, opened 2025-03-15
Updates to `RunUMAP`

samuel-marsh:
Hi Seurat Team,

This PR implements changes 1 and 3 mentioned in #9709.

1. It allows for users to pass `method = "uwot2"` which will implement `uwot::umap2`.  No changes made to parameters that are passed to `umap2` just added as additional method.  
2. It removes old warning that uwot does not support `metric = correlation` since it now does support that.
3. Updated manual entry to remove mention of `method = uwot-learn` since that was deprecated and no longer exists (warning already present in code if user supplies it) and replaces it with entry for `metric = "uwot2"`.

If there are any other changes you think should be made or updates I need to make please let me know.

Thanks as always!
Best,
Sam


samuel-marsh:
rebase to main 4/23/25

PR #9787, opened 2025-03-31
update Doheatmap

Liripo:
You can get possible gene names through Features. When there are more than 20,000 genes and 40,000 cells, the speed of LayerData (GetAssayData) is slowly.


Liripo:
My problem was that I had the wrong branch.

Issue #9768, opened 2025-03-21
Feature Request: DoHeatmap is slowly,may be faster


Liripo:
### Motivation

`Doheatmap` drawing takes more than 5 seconds

### Feature Description

I found it very slow when running `Doheatmap`, and found that the main reason was that it was slow to get the gene name of `scale.data`
![Image](https://github.com/user-attachments/assets/bc8d1cce-3104-4be4-a5ed-ab3e686c1d8e)

code is https://github.com/satijalab/seurat/blob/f8ae979cf2cd6fc51ad70baa47cc8f61f5e6bf94/R/visualization.R#L268

My `scale.data` has 2w gengs, 4w cells. Get only the gene name, which should be faster.

Liripo:
Can it be directly changed to：
```R
possible.features <- Features(x = object, layer = layer)
```

cyrillustan:
Looks like an interesting idea. Would you be interested in opening a PR for such a change? We can then test it, making sure it will work for all the occasions.

cyrillustan:
Included in #9835

PR #9788, opened 2025-03-31
Update visualization.R

Liripo:
You can get possible gene names through Features. When there are more than 20,000 genes and 40,000 cells, the speed of LayerData (GetAssayData) is very slow.

Fix the issue: https://github.com/satijalab/seurat/issues/9768


dcollins15:
Thanks for the PR @Liripo! I ended up merging this tweak as a part of #9835 👍

PR #9807, opened 2025-04-09
Fix a typo in `DietSeurat`

pormr:
Hello, I checked the code of `R/objects.R` and found a typo on line 655:
https://github.com/satijalab/seurat/blob/95ed7691a25365338848e3532733454915b475da/R/objects.R#L655-L657

This segment of code checks if `cells.keep` is lower than the number of columns in `object`; however, the assignment operator `<-` is used here. It may have unintended effects on the final Seurat object when the assays you wish to retain contain fewer cells than the total, since if you assign a number greater than the current length of the `cells.keep` vector, it will be extended with NA at the end.


dcollins15:
Thanks for this @pormr—I ended up merging your fix as a part of #9835 👍

PR #9812, opened 2025-04-10
Improves DimHeatmap documentation and default behavior when fast=FALSE (Issue 9810)

dkcoxie:
Clarifies documentation to specify that figure legends are currently not displayed when `fast=TRUE` but added titles with PC numbers to ggplot outputs so that both PC number and figure legends can are display on the same plot for default outputs, plus option to move position of shared patchwork legend when `combine=TRUE`

# Note


dkcoxie:
In teaching a beginners oriented workshop, we found that the default behavior of the `DimHeatmap` function does not allow for both a title with the PC number and a plot figure legend to be displayed at the same time. The revisions to `DimHeatmap` make the hidden figure legend with `fast=TRUE` more clear and also ensure that the PC number is displayed on each plot (combined or not) when `fast=FALSE`.

Issue #8428, opened 2024-02-06
SketchData + SCTransform bug


GreenGilad:
Hi Seurat team,

Using Seurat v5.0.1, calling SCTransform and then SketchData does not work

1. Running the Sketch pipeline as in the tutorial works fine.
2. Replacing the normalization, scaling and finding variable features with SCTransform results in the following error:
    ```
    > obj <- SCTransform(obj, verbose = FALSE)
    > obj <- SketchData(obj, method = "LeverageScore", verbose = FALSE)    
    Calcuating Leverage Score
    Error in `components()`:
    ! `model` must be one of "counts", not "data".
    ```
4. Line 513 in `sketching.R` is the one to throw the error. With `"SCT"` being the default assay, the `SketchData` calls `LeverageScore.StdAssay` with `object` in line 513 referring to the assay object. Within this line the error comes from `Cells(object, layer="data")`, and can be reproduced by simply calling:
    ```
    > Cells(obj[["SCT"]], layer="data")
    Error in `components()`:
    ! `model` must be one of "counts", not "data".
    ```
    So the problem is not even with the `LeverageScore` itself, but the assignment to the `scores` data-frame.
4. Calling `LeverageScore.Seurat` setting `layer="counts"` works
    ```
    > LeverageScore(obj, layer="counts", verbose=TRUE)
    Running LeverageScore for layer counts
    sampling 5000 cells for random sketching
    Performing QR decomposition
    Performing random projection
    An object of class Seurat 
    36362 features across 86612 samples within 2 assays 
    Active assay: SCT (17179 features, 3000 variable features)
    3 layers present: counts, data, scale.data
    1 other assay present: RNA
    ````
5. Calling `SketchData` with `layer="counts"` still fails.
    ````
    > SketchData(obj, layer="counts", method = "LeverageScore", verbose = FALSE)
    Error in `components()`:
    ! `model` must be one of "counts", not "data".
    ````
6. This time the error occurs on line 96, again calling `Cells(x=object[[assay]], ....)` because `layers.data` is set to be `"data"` on line 91.

Should line 91 be changed to use the passed `layer` argument? Why does `Cells(obj[["SCT"]])` cannot retrieve cell names when layer is "data"? but only "counts"?

As a minor side note:
- Notice typo in `SketchData` message "Calcuating" instead of "Calculating"
- `verbose = TRUE` is not propagated from `SketchData` to inner call to `LeverageScore`

GreenGilad:
[...]

PR #9830, opened 2025-04-21
Fix `Cells.SCTAssay` for `LeverageScore`/`SketchData`

dcollins15:
This PR makes two changes:
1. Updates the `Cells.SCTAssay` method to allow "regular" layer names (i.e., `"counts"`, `"data"`, and `"scale.data"`).
2. Drops the call to the `VariableFeatures` setter at the end of `SketchData`.

The first update makes `SCTAssay` inputs compatible with `SketchData` and `LeverageScore`. The second change is mostly a speedup, but in this context also avoids a warning from `VariableFeatures.SCTAssay` about unused arguments. 

Resolves:
- https://github.com/satijalab/seurat/issues/8428

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.2.99.9010
+Version: 5.2.99.9011
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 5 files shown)


longmanz (approved):
All changes look good!

PR #9831, opened 2025-04-21
Optimize `LeverageScore` for small BPCells inputs

dcollins15:
Huge thanks to @yuhanH for this update 🚀 

---

Optimize the LeverageScore computation to address the long processing time for small BPCell matrices (fewer than 7,500 cells). Convert the input into a sparse matrix and then compute accurate leverage scores.  

In the current implementation, sketching pbmc3k takes approximately 100 seconds. With sparse conversion, this time is reduced to 30 seconds. For larger datasets like hcabm40k, the computation remains at 40 seconds.

```
library(Seurat)
library(SeuratData)
library(BPCells) 
data('pbmc3k')
data('hcabm40k')
 

obj <- UpdateSeuratObject(pbmc3k)
# obj <- UpdateSeuratObject(hcabm40k)

obj[['RNA']] <- as(obj[['RNA']], Class = 'Assay5')
obj[['RNA']]$counts <- write_matrix_dir(obj[['RNA']]$counts, dir = '~/test/pbmc3k_BPCells', overwrite = T)
obj <- NormalizeData(obj)
obj <- FindVariableFeatures(obj)
 
system.time(obj <- SketchData(obj))

```

---

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.2.99.9011
+Version: 5.2.99.9012
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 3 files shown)


erlun1:
However, if my SeuratObject was created by merging single samples where each retains its independent BPCells directory, this design would cause significant timeouts in LeverageScore calculations. In such cases, should I rewrite a complete BPCells file directly from the merged SeuratObject?

PR #9841, opened 2025-04-23
fix LabelClusters split.by warning

kew24:
Just a quick fix to use the _splitting_ variable in the warning, not the _coloring_ variable, for a more accurate/useful warning message.


anashen:
Thank you @kew24 for pointing this out & for your interest in contributing! However, I believe this fix was implemented as part of commit 70da0cb (released with v5.4.0).

Issue #9543, opened 2024-12-11
Kmeans does not converge when running BuildNicheAssay


AlbertoFabbri93:
Run `BuildNicheAssay()` with a high number of niches:

```r
breast_cancer_patients_analyze <- BuildNicheAssay(
    object = breast_cancer_patients_analyze,
    fov = "LU001FFP03",
    group.by = col_cell_tumor_cluster_subtype,
    niches.k = 15,
    neighbors.k = 30
 )
```

Output:
```
Computing nearest neighbor graph
Centering and scaling data matrix
  |========================| 100%
Warning: did not converge in 10 iterations
```
`kmeans` did not converge ans there is not way to set `iter.max` to a higher value.

vdsukhov:
I encountered the same issue, likely due to working with a fairly large matrix. As a workaround, I decided to perform clustering independently using the `MiniBatchKmeans` function from the **ClusterR** package. Here's the approach I used:

```r
xobj <- BuildNicheAssay(
    xobj,
    fov = "fov",
    group.by = "cell_type_res_l2",
    niches.k = 10, neighbors.k = 10
)



# Let's use mini batch k-means since in my case the default Seurat implementation didn't converge
NN <- t(GetAssayData(xobj, "niche", layer="scale.data"))


set.seed(42)
mb <- MiniBatchKmeans(
    data          = NN,
    clusters      = 10,
    batch_size    = 10000,
    num_init      = 5,
    max_iters     = 200,
    init_fraction = 1,
    initializer   = "kmeans++",
    verbose       = TRUE
)

clusters <- predict_MBatchKMeans(data = NN, CENTROIDS = mb$centroids)
xobj@meta.data$neighborhood <- factor(clusters, levels = sort(unique(clusters)))
```

Essentially, I extract the scaled data from the `niche` assay and apply clustering manually. This avoids the limitations caused by the hardcoded k-means implementation within the [`BuildNicheAssay`](https://github.com/satijalab/seurat/blob/9354a78887e66a3f7d9ba6b726aa44123ad2d4af/R/utilities.R#L3075-L3079) function in Seurat.

zskylarli:
Hello - thank you for posting this issue, and appreciate @vdsukhov for making the PR/suggesting a solution! Happy to merge in the PR to resolve this problem with some fixes. Thanks again!

PR #9898, opened 2025-05-20
Add arguments to kmeans call in BuildNicheAssay function

vdsukhov:
### Summary

This PR adds support for passing additional arguments to the `kmeans` function within `BuildNicheAssay`. The goal is to help address convergence issues such as the warning:

```
Warning: did not converge in 10 iterations
```

In some cases, the default settings of `kmeans` may not be sufficient for convergence, particularly with complex or noisy datasets. By allowing users to specify parameters like `iter.max`, `nstart`, and `algorithm`, the clustering behavior can be tuned for better performance.

With the added flexibility, I was able to resolve the convergence warning using the following code:

```r
set.seed(1)
xobj <- BuildNicheAssay(
    xobj,
    fov = "fov",
    group.by = "cell_type_res_l2",
    niches.k = 10, neighbors.k = 10,
    algorithm = "Lloyd", iter.max = 1000, 
    nstart = 1
)
```

Addresses #9543

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -3076,6 +3076,7 @@ BuildNicheAssay <- function(
   results <- kmeans(
     x = t(object[[assay]]@scale.data),
     centers = niches.k,
+    nstart = 30,
     ...
   )
   object[[cluster.name]] <- results[["cluster"]]
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -3027,6 +3027,10 @@ CreateCategoryMatrix <- function(
 #' @param cluster.name Name of output clusters
 #' @param neighbors.k Number of neighbors to consider for each cell
 #' @param niches.k Number of clusters to return based on the niche assay
+#' @param iter.max The maximum number of iterations allowed (see \code{\link[stats]{kmeans}})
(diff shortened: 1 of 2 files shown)


zskylarli (changes requested):
Hello - thank you for making the PR! Happy to merge it in if you could remove the added arguments from the main BuildNicheAssay function (along with changes in documentation) and instead pass in `...` like so:

```
BuildNicheAssay <- function(
  object,
  fov,
  group.by,
  assay = "niche",
  cluster.name = "niches",
  neighbors.k = 20,
  niches.k = 4,
  ...
) 
```
and 
```
results <- kmeans(
  x = t(object[[assay]]@scale.data),
  centers = niches.k,
  nstart = 30,
  ...
)
```

This would allow us to support any user that wants to explicitly pass in these arguments while avoiding bloating the function signature. Thank you!

vdsukhov:
@zskylarli Hi. 

I have added one more commit that wrap arguments in ellipsis.

zskylarli (approved):

zskylarli:
@vdsukhov Merged - thank you for submitting this PR!

Issue #9909, opened 2025-05-28
BUG:  5.3.0 Seurat::UMAPPlot error


Liripo:
### Issue Description

Cannot use `Seurat::UMAPPlot` directly，but `UMAPPlot` can work

### Reproducing Code Example

```R
> Seurat::UMAPPlot(obj,group.by = "cell_type")
Error in object[[reduction]] : 
  ‘seurat::umap’ not found in this Seurat object
 
> UMAPPlot(obj,group.by = "cell_type")
```

### Error Message

```shell

```

### Session Info

```shell
> sessionInfo()
R version 4.4.1 (2024-06-14)
Platform: x86_64-conda-linux-gnu
Running under: Red Hat Enterprise Linux 9.4 (Plow)

Matrix products: default
BLAS/LAPACK: /data03/lead/userdata/huanghuichang/Software/miniconda3/lib/libmkl_rt.so.2;  LAPACK version 3.9.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

time zone: Asia/Shanghai
tzcode source: system (glibc)

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] Seurat_5.3.0       SeuratObject_5.0.2 sp_2.1-4          

loaded via a namespace (and not attached):
  [1] deldir_2.0-4           pbapply_1.7-2          gridExtra_2.3         
  [4] rlang_1.1.4            magrittr_2.0.3         RcppAnnoy_0.0.22      
  [7] matrixStats_1.3.0      ggridges_0.5.6         compiler_4.4.1        
 [10] spatstat.geom_3.2-9    png_0.1-8              vctrs_0.6.5           
 [13] reshape2_1.4.4         stringr_1.5.1          pkgconfig_2.0.3       
 [16] fastmap_1.2.0          labeling_0.4.3         utf8_1.2.4            
 [19] promises_1.3.0         purrr_1.0.2            jsonlite_1.8.8        
[...]

PR #9964, opened 2025-06-25
Fix SpecificDimPlot when call includes Seurat::

samuel-marsh:
Hi Seurat Team,

This is one line fix to address #9909.  It just adds a second `gsub` call to check if function call is being passed with `Seurat::`.

I merged this with fix/5.3.1 per previous PRs since this is quick and easy one but can change to main if preferable.

Thanks as always!
Sam

### DIFF
--- a/R/convenience.R
+++ b/R/convenience.R
@@ -442,6 +442,8 @@ SpecificDimPlot <- function(object, ...) {
   funs <- sys.calls()
   name <- as.character(x = funs[[length(x = funs) - 1]])[1]
   name <- tolower(x = gsub(pattern = 'Plot', replacement = '', x = name))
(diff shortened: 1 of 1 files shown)


rharao:
I think the whole idea of diving into the call stack text to figure out what convenience function the user requested is bizarre. In #9493 I proposed something that I think is more sensible.

mccormickceci:
Hello, thank you so much for contributing to the Seurat codebase! I think you are absolutely right that this would likely fix the issue referenced here. At this time we are no longer going to be making changes to these convenience functions (TSNEPlot, UMAPPlot, PCAPlot) as we have replaced them with DimPlot() in our vignettes.

samuel-marsh:
Makes sense, totally agree!

PR #9977, opened 2025-07-01
Add support for Vizgen objects for InteractiveSpatialPlot

YoukaiFromAccounting:
This small change let's the InteractiveSpatialPlot be used for Vizgen objects (albeit very slowly due to the large size of many Vizgen datasets); users can now run InteractiveSpatialPlot on Vizgen objects and perform subsetting to retrieve a selection of cells.

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -3912,14 +3912,21 @@ InteractiveSpatialPlot <- function(
   image_obj <- object[[image]]
 
   #Detect image type based on slot names
-  if ("boundaries" %in% slotNames(image_obj)) {
+  #Currently supported: VISIUM V2, SLIDESEQ, VIZGEN
+  img_class <- class(image_obj)[1]
+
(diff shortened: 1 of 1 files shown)


anashen:
Thanks! Can you provide a small reproducible example?

YoukaiFromAccounting:
@anashen, here is a small working example on the fix-interactivespatial branch:

[VizgenSnippet.Rmd.zip](https://github.com/user-attachments/files/21919873/VizgenSnippet.Rmd.zip)

Let me know if there are any issues.

anashen (commented):
Can confirm that the interactivity works! However, the image is flipped horizontally on the plot (though the logic has not been changed from the currently existing version). We'll need to discuss how to address this, as well as how we want to move forward with support for this functionality, seeing as the Vizgen objects that we are using to test these functions are of an outdated format.

anashen:
Converting to draft PR so we can put this back in development.

PR #9978, opened 2025-07-01
Feature: Updates for Space Ranger 4.0

stephenwilliams22:
Updates include helpers and readers for Space Ranger cell segmentation. 

Integrates 10x cell segmentations of visium HD data into the Seurat object. 

Issues

- [ ] Converting the SF object into a data.frame and back into an SF object in Read10X_Segmentations for CreateFOV seems like it should be unnecessary but is required
- [ ] Should probably make the code for loading the segmentations in Load10X_Spatial a function? 
- [ ] Associating a tissue image with the segmentations 
- [ ] Need clarity between counts and data slots for the segmented cell matrix
- [ ] ImageFeaturePlot and ImageFeaturePlot  do not plot the cell polygons correctly. If we could use `geom_sf` in `SpatialFeaturePlot` I think that would be more desirable

### DIFF
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -445,9 +445,8 @@ GetResidual <- function(
     new.residuals <- lapply(
       X = sct.models,
       FUN = function(x) {
-        FetchResidualSCTModel(object = object,
-                              assay = assay,
-                              umi.assay = umi.assay,
+        FetchResidualSCTModel(object = object[[assay]],
+                              umi.object = object[[umi.assay]],
                               SCTModel = x,
                               new_features = features,
                               replace.value = replace.value,
@@ -499,6 +498,7 @@ GetResidual <- function(
 #' @param filename Name of H5 file containing the feature barcode matrix
 #' @param slice Name for the stored image of the tissue slice
 #' @param bin.size Specifies the bin sizes to read in - defaults to c(16, 8)
+#' @param load.segmentations Should the cell segmentation from space ranger be read in?
 #' @param filter.matrix Only keep spots that have been determined to be over
 #' tissue
 #' @param to.upper Converts all feature names to upper case. Can be useful when
@@ -528,6 +528,7 @@ Load10X_Spatial <- function (
   assay = "Spatial",
   slice = "slice1",
   bin.size = NULL,
+  load.segmentations = FALSE,
   filter.matrix = TRUE,
   to.upper = FALSE,
   image = NULL,
@@ -540,7 +541,7 @@ Load10X_Spatial <- function (
     # but also raise a warning
     warning(
       paste0(
-        "`data.dir` expects a single value but recieved multiple - ",
+        "`data.dir` expects a single value but received multiple - ",
         "continuing using the first: '",
         data.dir,
         "'."
@@ -640,6 +641,22 @@ Load10X_Spatial <- function (
     object.list[[1]],
     y = object.list[-1]
   )
+
+  # Possibly make this into a new function
+  if(load.segmentations) {
+    segmentation.assay.name <- "Segmentations"
+    seg.data.dir <- paste0(data.dir, "/segmented_outputs/")
+    seg.counts.path <- paste0(seg.data.dir, "filtered_feature_cell_matrix.h5")
+    segmentation.counts <- Read10X_h5(seg.counts.path)
+    segmentation.count.cellIDs <- colnames(segmentation.counts)
+    #segmentation.image <- Read10X_Image(image.dir = paste0(seg.data.dir, "spatial"), assay = "Segmentations", slice = "slice1.segmentations")
+    segmentation.object <- CreateSeuratObject(segmentation.counts, assay = segmentation.assay.name)
+    #segmentation.object[["slice1.segmentation.image"]] <- segmentation.image
+    segmentation.fov <- Read10X_Segmentations(image.dir = paste0(seg.data.dir, "spatial"), data.dir = data.dir, cell.names = segmentation.count.cellIDs)
+    segmentation.object[["slice1.polygons"]] <- segmentation.fov
+    object <- merge(x = object, y = segmentation.object)
+    DefaultAssay(object = object) <- segmentation.assay.name
+  }
 
   return(object)
 }
@@ -1264,8 +1281,8 @@ Read10X_Image <- function(
       image = image
     )
 
-    # As of v5.1.0 `Radius.VisiumV1` no longer returns the value of the 
-    # `spot.radius` slot and instead calculates the value on the fly, but we 
+    # As of v5.1.0 `Radius.VisiumV1` no longer returns the value of the
+    # `spot.radius` slot and instead calculates the value on the fly, but we
     # can populate the static slot in case it's depended on.
     visium.v1@spot.radius <- Radius(visium.v1)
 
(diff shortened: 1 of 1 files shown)


stephenwilliams22:
@jsicherman it would be great if you could have a look at this.

R/preprocessing.R
```diff
@@ -1390,6 +1408,97 @@ Read10X_ScaleFactors <- function(filename) {
...
+#' @param segmentation.type Which segmentations to load, cell or nucleus. If using nucleus the full matrix from cells is still used
+#' @param scale.factor If scaling the segmentations coordinates for the associated tissue image. "lowres" or "highres"
+#'
+#' @return A FOV
+#'
+#' @export
+#' @concept preprocessing
+#'
+Read10X_GeoJson <- function(data.dir, segmentation.type = "cell", scale.factor = NULL) {
+  segmentation_polygons <- read_sf(file.path(data.dir,"segmented_outputs", paste0(segmentation.type, "_segmentations.geojson")))
```
jsicherman, line 1499: nit: since this function doesnt imply it is HD specific, I would refrain from hardcoding these paths and pass them through as parameters instead. Or rename the function

R/preprocessing.R
```diff
@@ -1390,6 +1408,97 @@ Read10X_ScaleFactors <- function(filename) {
...
+  if (!is.null(scale.factor)) {
+    scale.factor.path <- file.path(data.dir,"segmented_outputs/spatial/scalefactors_json.json")
+    scale.factors <- jsonlite::fromJSON(scale.factor.path)
```
jsicherman: nit: `Read10X_ScaleFactors`?

R/preprocessing.R
```diff
@@ -1390,6 +1408,97 @@ Read10X_ScaleFactors <- function(filename) {
...
+{
+
+  image <- png::readPNG(source = file.path(image.dir, image.name))
+  scale.factors <- Read10X_ScaleFactors(filename = file.path(image.dir, "scalefactors_json.json"))
+  coordinates <- Read10X_GeoJson(data.dir = data.dir, segmentation.type = segmentation.type)
+  key <- Key(slice, quiet = TRUE)
+
+  st_coords <- merge(
+    as.data.frame(sf::st_coordinates(coordinates)),
+    data.frame(
```
jsicherman, line 1451: Agree with OP that it would be nice to have a `CreateFOV.sfc` or similar that allows this construction without having to do two round trips through polygon construction. Not sure if such a method already exists in Seurat

R/preprocessing.R
```diff
@@ -1390,6 +1408,97 @@ Read10X_ScaleFactors <- function(filename) {
...
+                                   data.dir,
+                                   image.name = "tissue_lowres_image.png",
+                                   assay = "Segmentations",
+                                   slice = "slice1.segmentations",
+                                   segmentation.type = "cell",
+                                   cell.names
+)
+{
+
+  image <- png::readPNG(source = file.path(image.dir, image.name))
```
jsicherman, line 1440: This is unused?

R/preprocessing.R
```diff
@@ -640,6 +642,22 @@ Load10X_Spatial <- function (
...
+    seg.counts.path <- paste0(seg.data.dir, "filtered_feature_cell_matrix.h5")
+    segmentation.counts <- Read10X_h5(seg.counts.path)
+    segmentation.count.cellIDs <- colnames(segmentation.counts)
+    #segmentation.image <- Read10X_Image(image.dir = paste0(seg.data.dir, "spatial"), assay = "Segmentations", slice = "slice1.segmentations")
+    segmentation.object <- CreateSeuratObject(segmentation.counts, assay = segmentation.assay.name)
+    #segmentation.object[["slice1.segmentation.image"]] <- segmentation.image
+    segmentation.fov <- Read10X_Segmentations(image.dir = paste0(seg.data.dir, "spatial"), data.dir = data.dir, cell.names = segmentation.count.cellIDs)
+    segmentation.object[["slice1.polygons"]] <- segmentation.fov
+    object <- merge(x = object, y = segmentation.object)
+    DefaultAssay(object = object) <- segmentation.assay.name
```
jsicherman, line 661: I'm not sure if it's intuitive to have this create two assays and merge them together, rather than just one or the other. Will defer to you since you look at this data more, though.

Also, this output won't exist for Visium SD, which is also loaded through this preprocessing function, right? Should we raise a more informative error in those cases?

R/preprocessing.R
```diff
@@ -640,6 +642,22 @@ Load10X_Spatial <- function (
     y = object.list[-1]
   )
 
+  # Possibly make this into a new function
+  if(load.segmentations) {
+    segmentation.assay.name <- "Segmentations"
+    seg.data.dir <- paste0(data.dir, "/segmented_outputs/")
+    seg.counts.path <- paste0(seg.data.dir, "filtered_feature_cell_matrix.h5")
```
jsicherman, line 649: nit: `filename`

jsicherman:
@stephenwilliams22 what do you mean when you say the polys aren't plotted correctly? Are they not closed, or merged together or something?

R/preprocessing.R
stephenwilliams22: Yeah, we need to get the tissue image in there at some point so I left it in.

stephenwilliams22:
> @stephenwilliams22 what do you mean when you say the polys aren't plotted correctly? Are they not closed, or merged together or something?

I'm not totally sure but if you look here they just look like balls of twine.

image

jsicherman:
@stephenwilliams22 looks like the vertices aren't ordered sequentially, which is weird if you could construct and plot them without issue in your original sfc? But also, how many vertices do you have per cell segmentation? Might be worth simplifying them if they're as complex as that screenshot seems to imply

stephenwilliams22:
> @stephenwilliams22 looks like the vertices aren't ordered sequentially, which is weird if you could construct and plot them without issue in your original sfc? But also, how many vertices do you have per cell segmentation? Might be worth simplifying them if they're as complex as that screenshot seems to imply

fixed. ordering was the issue

YoukaiFromAccounting:
Hey @stephenwilliams22, I am trying to test the new Load10x_Spatial and Read10x_Segmentations functions with a sample V3D object (https://www.10xgenomics.com/datasets/visium-hd-three-prime-mouse-embryo-fresh-frozen). Currently I am running into the following issue:

If I do not specify an image in Load10X_Spatial ie:
```{R}
v3d <- Load10X_Spatial(
  data.dir = data.dir,
  filename = "filtered_feature_cell_matrix.h5",
  slice = "slice1",
  image = NULL,
  load.segmentations = TRUE
)
```
Then I get the following error when trying to run Load10X_Spatial:
```{R}
Error in if (tools::file_ext(filename) == "parquet") { : 
  argument is of length zero
```

It seems that without an image, the function tries to automatically call Read10X_Image() which attempts to locate a spatial/tissue_positions.parquet file, which does not exist for V3D objects.  I was able to get Read10X_Segmentations working on this V3D object however:

image

stephenwilliams22:
> v3d <- Load10X_Spatial(
>   data.dir = data.dir,
>   filename = "filtered_feature_cell_matrix.h5",
>   slice = "slice1",
>   image = NULL,
>   load.segmentations = TRUE
> )

@YoukaiFromAccounting you do not need to use `filename = "filtered_feature_cell_matrix.h5",` that is implied when `load.segmentations = TRUE`. Do you think we should make that another parameter to the function? I don't think people go and change these file names very often (if ever). Also, does your `data.dir,` point to the top level of the pipeline outputs like this? 
```
binned_outputs
segmented_outputs
spatial
Visium_HD_3prime_Mouse_Embryo_barcode_mappings.parquet
Visium_HD_3prime_Mouse_Embryo_binned_outputs.tar.gz
Visium_HD_3prime_Mouse_Embryo_cloupe_008um.cloupe
Visium_HD_3prime_Mouse_Embryo_cloupe_cell.cloupe
Visium_HD_3prime_Mouse_Embryo_feature_slice.h5
Visium_HD_3prime_Mouse_Embryo_metrics_summary.csv
Visium_HD_3prime_Mouse_Embryo_molecule_info.h5
Visium_HD_3prime_Mouse_Embryo_segmented_outputs.tar.gz
Visium_HD_3prime_Mouse_Embryo_spatial.tar.gz
Visium_HD_3prime_Mouse_Embryo_web_summary.html
```

The `Load10X_Spatial` function is intended to point to the top level. Maybe you pointed it to `segmented_outputs`?

YoukaiFromAccounting:
@stephenwilliams22 Thanks for your help; I was able to get it working (I also switched to a smaller V3D dataset https://www.10xgenomics.com/datasets/visium-hd-three-prime-zebrafish-head-fresh-frozen). I am able to get a visualization if I run the following: 

```{R}
#Complete dataset path
data.dir <- "~/spaceranger_zebrafish"

#Load expression data and spatial image
v3d <- Load10X_Spatial(
  data.dir = data.dir,
  slice = "slice1",
  load.segmentations = TRUE
)

#Check existing boundary types
Boundaries(v3d[["slice1.polygons"]])
seg <- v3d[["slice1.polygons"]]@boundaries[["segmentation"]]
seg

#Inspect segmentations
DefaultBoundary(v3d[["slice1.polygons"]]) <- "segmentation"

ImageDimPlot(
  v3d,
  fov = "slice1.polygons",
  border.color = "white",
  border.size = 0.1
)
```

which gets me the following visualization:
image

Zoomed in visualization:
image

stephenwilliams22:
> @stephenwilliams22 Thanks for your help; I was able to get it working (I also switched to a smaller V3D dataset https://www.10xgenomics.com/datasets/visium-hd-three-prime-zebrafish-head-fresh-frozen). I am able to get a visualization if I run the following:


Very cool! Do you think you all can take it from here to get the tissue image in the background? I think we also need to test if the regular data normalization workflow is working properly. Also, I'm wondering if this should be rewritten to use `geom_sf` which is much easer.

YoukaiFromAccounting:
@stephenwilliams22 I'm not quite sure what you mean by using geom_sf; do you mean extracting the segmentation polygons from an sf object and plotting those with ggplot instead of using ImageDimPlot?

stephenwilliams22:
> @stephenwilliams22 I'm not quite sure what you mean by using geom_sf; do you mean extracting the segmentation polygons from an sf object and plotting those with ggplot instead of using ImageDimPlot?

Sorry, what i meant was in reference to my last checkbox at the top of the page. Right now we are implementing this in a Xenium fashion which does some back and forth making the polygons which are already in an `sf` object to a data.frame and then back to `sf`. This is necessary because xenium segmentations are not provided as a geojson. What would be nice is if we provided a path to the user to go through `SpatialFeaturePlot` where we could update that code to use `geom_sf` instead of drawing the polygons "by hand". I think that users would find this flow more intuitive since this is how the original visium assay works for visualization

rsatija:
Hi Stephen - we've been looking over this. Do you have example code (doesn't have to use Seurat at all) for 
1. how to load an HD data,set [this one ](https://www.10xgenomics.com/datasets/visium-hd-three-prime-zebrafish-head-fresh-frozen) for example, and plot the segmentations with geom_sf in a way that avoids the conversion you mentioned?
2. ideally this example plots a tissue image as well

If you have this, we can certainly take the lead as using that as a prototype for updating the Seurat functions.

dixinSJ:
How is this work progressing?

YoukaiFromAccounting:
We recommend that individuals interested in this update check out our Discussions board at (https://github.com/satijalab/seurat/discussions/9991) for questions/concerns users may have.

PR #10024, opened 2025-07-28
Add support for Xenium protein data

pryvkin10x:
- Xenium 4.0 will produce a new feature type called "Protein Expression"
- In order to limit breakage of downstream tools, it stores values as integers multiplied by 10 (essentially fixed point).
- This scaling factor is stored in the HDF5 file as an attribute called `protein_scaling_factor`
- This PR loads the new feature type and applies the scaling factor when it is available.

### DIFF
--- a/R/convenience.R
+++ b/R/convenience.R
@@ -260,7 +260,8 @@ LoadXenium <- function(
     `Unassigned Codeword` = 'BlankCodeword',
     `Negative Control Codeword` = 'ControlCodeword',
     `Negative Control Probe` = 'ControlProbe',
-    `Genomic Control` = 'GenomicControl'
+    `Genomic Control` = 'GenomicControl',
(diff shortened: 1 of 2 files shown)


anashen:
Thank you @pryvkin10x! Tested with Xenium 4.0 data & output looks good.

- - -
Closing & reopening to trigger checks, since the action appears stuck.

shoebridges:
Hi, 
Is this still being implemented, or has this been completed, as I am having issues finding the protein expression in the seurat object when using the LoadXenium function.

anashen:
Hi @shoebridges,

Yes, this will be available on the main branch very soon. In the meantime, you can install via

```R
devtools::install_github("pryvkin10x/seurat", ref = "paul/xenium-protein", force = TRUE)
```

Also, be sure to restart your R session after installation.

Issue #10076, opened 2025-09-08
BUG: <SpatialFeaturePlot alpha> Not recognization of alpha parameter in SpatailFeaturePlot


yifanfu01:
### Issue Description

Hi team,

I do apprecriate your work on keeping Seurat update.

I occurred HD segment results and visualization. I found it can't recognize the `alpha` parameter.

I run:

```
SpatialFeaturePlot(
  object = v3d,
  features = "nCount_Spatial.Polygons",
  images = "slice1.polygons",
  alpha=c(0.2),max.cutoff = 3000,
  plot_segmentations = TRUE,
  stroke = 0.01,
  crop = TRUE
) +
  coord_sf(
    xlim = x_crop,
    ylim = y_crop,
    expand = FALSE
  )
```

<img width="2100" height="2100" alt="Image" src="https://github.com/user-attachments/assets/a70d16ce-56e8-4ed9-94dc-c2180d4714c1" />

### Reproducing Code Example

```R
SpatialFeaturePlot(
  object = v3d,
  features = "nCount_Spatial.Polygons",
  images = "slice1.polygons",
  alpha=c(0.2),max.cutoff = 3000,
  plot_segmentations = TRUE,
  stroke = 0.01,
  crop = TRUE
) +
  coord_sf(
    xlim = x_crop,
    ylim = y_crop,
    expand = FALSE
  )
```

### Error Message

```shell
No error. And also no warning.
```

### Session Info

```shell
Seurat = 5.3.1.999
```

anashen:
@yifanfu01 thank you for reporting this issue! This has been resolved in the PR created by @YoukaiFromAccounting at #10079, which I have just merged into the `spaceranger-4.0` branch.

Please be sure to restart your R session after reinstalling the branch from GitHub, and feel free to follow up with any questions. Thanks!

PR #10079, opened 2025-09-08
Fix incorrect alpha parameter being used for SpatialFeature plots

YoukaiFromAccounting:
# Issues

- https://github.com/satijalab/seurat/issues/10076

## Issue 1

SingleSpatialPlot plots transparency in cells using the _alpha_ parameter. This parameter is derived in two different ways:

For DimPlots, alpha is set to be equal to pt.alpha, which is a number between 0 and 1. 

For FeaturePlots, alpha is set to be equal to alpha.by, which is set to be name of a feature (ie: alpha.by = "rho"). Then, when plotting feature plots, ggplot uses the actual values in this feature column (ie: rho expression levels), scaled between the alpha range provided (0-1) to plot the alphas. 

For objects containing sf.data, plotting was previously done only using pt.alpha. This meant that for FeaturePlots, it was not possible to set the alpha levels of cells. This fix addresses this issue by adding a check for alpha.by for plotting of objects containing sf.data; if alpha.by is found, then the plot is a FeaturePlot and ggplot will use alpha.by instead of pt.alpha to define cell transparency. 

image

With this fix, the following should work:
- SpatialDimPlot with alpha range defined (ie: c(0.2-0.6)), plot_segmentations = TRUE
- SpatialDimPlot with alpha range defined, plot_segmentations = FALSE
- SpatialFeaturePlot with alpha range defined, plot_segmentations = TRUE
- SpatialFeaturePlot with alpha range defined, plot_segmentations = FALSE

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -9501,73 +9501,63 @@ SingleSpatialPlot <- function(
       #Plot (currently independently of switch/case)
       if(!plot_segmentations){
         #If plot_segmentations FALSE, then plot just the centroids from sf data
-        if(is.null(pt.alpha)){
+        
+        if (is.null(pt.alpha)) {
           #If pt.alpha not provided, then alpha parameter is derived from group/cluster data
           #Use alpha.by instead of pt.alpha
-          ggplot(sf.cleaned, aes(x = x, y = y)) +
-            annotation_custom(
-                grob = image.grob,
-                xmin = 0,
-                xmax = image.width,
-                ymin = 0,
-                ymax = image.height
-              ) +
-            geom_point(
-                shape = 21, 
-                stroke = stroke, 
-                size=pt.size.factor, 
-                aes_string(fill = col.by, alpha = alpha.by)
-              ) +
-            coord_fixed() +
-            xlab("x") +
-            ylab("y") +
-            theme_minimal()
-        }else{
+          geom_point_layer <- geom_point(
+            shape = 21, 
+            stroke = stroke, 
+            size = pt.size.factor, 
+            aes_string(fill = col.by, alpha = alpha.by)
+          )
+        } else {
           #If pt.alpha is indeed provided, then use that to define alpha
-          ggplot(sf.cleaned, aes(x = x, y = y)) +
-            annotation_custom(
-                grob = image.grob,
-                xmin = 0,
-                xmax = image.width,
-                ymin = 0,
-                ymax = image.height
-              ) +
-              geom_point(shape = 21, 
-              stroke = stroke, 
-              size=pt.size.factor, 
-              aes_string(fill = col.by), 
-              alpha = if (is.null(pt.alpha)) 1 else pt.alpha
-            ) +
-            coord_fixed() +
-            xlab("x") +
-            ylab("y") +
-            theme_minimal()
+          geom_point_layer <- geom_point(
+            shape = 21,
+            stroke = stroke,
+            size = pt.size.factor,
+            aes_string(fill = col.by), 
+            alpha = if (is.null(pt.alpha)) 1 else pt.alpha
+          )
         }
+        ggplot(sf.cleaned, aes(x = x, y = y)) +
+          annotation_custom(
+              grob = image.grob,
+              xmin = 0,
+              xmax = image.width,
+              ymin = 0,
+              ymax = image.height
+            ) +
+          geom_point_layer +
+          coord_fixed() +
+          xlab("x") +
+          ylab("y") +
+          theme_minimal()
+      
       }else{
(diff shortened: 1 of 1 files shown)


R/visualization.R
```diff
@@ -9500,39 +9500,82 @@ SingleSpatialPlot <- function(
...
+          #Use alpha.by instead of pt.alpha
+          ggplot(sf.cleaned, aes(x = x, y = y)) +
+            annotation_custom(
+                grob = image.grob,
+                xmin = 0,
+                xmax = image.width,
+                ymin = 0,
+                ymax = image.height
+              ) +
+            geom_point(shape = 21, stroke = stroke, size=pt.size.factor, aes_string(fill = col.by, alpha = alpha.by)) +
```
anashen: Lint - please put each parameter of `geom_point` on its own line to avoid long line width

R/visualization.R
```diff
@@ -9500,39 +9500,82 @@ SingleSpatialPlot <- function(
...
+          #If pt.alpha is indeed provided, then use that to define alpha
+          ggplot(sf.cleaned, aes(x = x, y = y)) +
+            annotation_custom(
+                grob = image.grob,
+                xmin = 0,
+                xmax = image.width,
+                ymin = 0,
+                ymax = image.height
+              ) +
+            geom_point(shape = 21, stroke = stroke, size=pt.size.factor, aes_string(fill = col.by), alpha = if (is.null(pt.alpha)) 1 else pt.alpha) +
```
anashen: Lint - please put each parameter of `geom_point` on its own line to avoid long line width

R/visualization.R
```diff
@@ -9500,39 +9500,82 @@ SingleSpatialPlot <- function(
...
-          scale_fill_viridis_d(option = "plasma", alpha = 0.5) +
-          coord_sf() +
-          theme_void()
+            geom_sf(
+              data = sf.cleaned,
+              aes_string(fill = col.by, alpha = alpha.by),
+              color = "black",
+              linewidth = stroke
+            ) +
+            scale_fill_viridis_d(option = "plasma", alpha = 0.5) +
```
anashen: I think you can drop the `alpha` parameter of `scale_fill_viridis_d` here. Tested without it, no change to functionality

R/visualization.R
```diff
@@ -9500,39 +9500,82 @@ SingleSpatialPlot <- function(
...
+              ymax = image.height
+            ) +
+            geom_sf(
+              data = sf.cleaned,
+              aes_string(fill = col.by),
+              alpha = if (is.null(pt.alpha)) 1 else pt.alpha,
+              color = "black",
+              linewidth = stroke
+            ) +
+            scale_fill_viridis_d(option = "plasma", alpha = 0.5) +
```
anashen: Drop the `alpha` parameter of `scale_fill_viridis_d` here

YoukaiFromAccounting:
@anashen Thanks for the review and the suggestions!

I’ve addressed all your comments in the latest commit. Let me know if you want to take another look, or feel free to merge if everything looks good.

anashen:
Looks good, thanks @YoukaiFromAccounting!

anashen:
On second thought, I think we can condense the logic so we can reuse code as much as possible :) Will make some suggestions.

anashen:
My suggestion is to define a layer that is defined differently depending on whether `pt.alpha` is `NULL`.

Here's an example of what I mean for the case that `plot_segmentations = FALSE` (I've just rearranged your code):
- - -
```R
if (is.null(pt.alpha)) {
  #If pt.alpha not provided, then alpha parameter is derived from group/cluster data
  #Use alpha.by instead of pt.alpha
  geom_point_layer <- geom_point(
    shape = 21, 
    stroke = stroke, 
    size = pt.size.factor, 
    aes_string(fill = col.by, alpha = alpha.by)
  )
} else {
  #If pt.alpha is indeed provided, then use that to define alpha
  geom_point_layer <- geom_point(
    shape = 21,
    stroke = stroke,
    size = pt.size.factor,
    aes_string(fill = col.by), 
    alpha = if (is.null(pt.alpha)) 1 else pt.alpha
  )
}
```

You can then add this layer in _one_ call to `ggplot`:
```R
ggplot(sf.cleaned, aes(x = x, y = y)) +
  annotation_custom(
      grob = image.grob,
      xmin = 0,
      xmax = image.width,
      ymin = 0,
      ymax = image.height
    ) +
  geom_point_layer +
  coord_fixed() +
  xlab("x") +
  ylab("y") +
  theme_minimal()
```
- - -
Thanks!

YoukaiFromAccounting:
@anashen Great suggestion! I've implemented this approach - defining the layer conditionally upfront is definitely cleaner than separate ggplot calls.

Issue #10070, opened 2025-09-03
[Visium HD 3' beta support] Spatial subsetting issue


YoukaiFromAccounting:
This issue is a continuation of this comment on Visium HD 3' support here:

- https://github.com/satijalab/seurat/pull/10028#issuecomment-3244291423

_Originally posted by @BioinfSina in https://github.com/satijalab/seurat/issues/10028#issuecomment-3244291423_

- - -

I have been trying to subset the zebrafish head dataset to only the eye area. 
Plotting the tissue coordinates I got this overview:
<img width="1224" height="812" alt="zebrafish_head" src="https://github.com/user-attachments/assets/bf49dc06-f431-4c0c-b15b-c28704198645" />

Using this I estimated my eye to be between  200<x<400 and 100<y<250. That does not return the expected area though, instead get's me this:
<img width="1224" height="812" alt="zebrafish_eye_expectedcoords" src="https://github.com/user-attachments/assets/ab9720aa-f533-4703-a5ed-e31565ed8262" />

I realised the x coordinates were fine, but the y coordinates seem incorrect. After playing around with y coordinates I found that this code gave me the results I wanted:

`tissue <- GetTissueCoordinates(v3d, image = "slice1.polygons")`

`tissue_filtered <- tissue[tissue$x>=200 & tissue$x<=400 & tissue$y>=250 & tissue$y<=400, ]`

`cropped_cells <- tissue_filtered$cell`

`eye <- subset(v3d, cells = cropped_cells)`

`SpatialFeaturePlot(eye, images = "slice1.polygons", features = "nCount_Spatial.Polygons", plot_segmentations = TRUE, crop = TRUE) 
`

Resulting in this cropped subset:
<img width="1224" height="812" alt="zebrafish_eye_adaptedCoords" src="https://github.com/user-attachments/assets/546765c7-034b-4a17-97b9-db7eb063afc2" />

That doesn't match the previously plotted coordinates though. I had used this code to plot the full dataset:
`SpatialDimPlot(v3d, images = "slice1.polygons",pt.size.factor = 0.1, plot_segmentations = FALSE) +`
  `theme(axis.text.x = element_text(size = 10, color = "black"),
        axis.text.y = element_text(size = 10, color = "black"),
        axis.title.x = element_text(size = 12, color = "black"),
        axis.title.y = element_text(size = 12, color = "black"),
        axis.line = element_line(color = "black"),
        axis.ticks = element_line(color = "black"),
[...]

PR #10082, opened 2025-09-09
Realign polygon coordinates with ggplot coordinate system

YoukaiFromAccounting:
# Issues

- https://github.com/satijalab/seurat/issues/10070

## Issue 1

For Spatial plots involving segmentations and an sf.data slot, there exists an inconsistency between the ggplot coordinate system and the coordinate system used by the segmentation polygons. While ggplot uses a traditional cartesian coordinate system with the origin (0,0) treated as the bottom left point of quadrant 1, segmentation coordinates are read in while treating (0,0) as their top left point. This meant that while cells could be reoriented correctly during plotting, their inherent coordinates (as used by GetTissueCoordinates) would be upside-down compared to the actual plot.

Example of what a SpatialFeaturePlot would look like if the original coordinates of cells were plotted without any transformations: 

image

This lead to issues when attempting to subset Visium HD 3' datasets, as the actual locations of cells differed from where they appeared in Spatial plots:

Below figures show the actual (left) vs expected (right) cell populations when subsetting based on ggplot coordinates:
imageimage

This fix addresses this issue by adjusting the Read10x_HD_GeoJson function in preprocessing.R to invert polygonal coordinates to match those of the ggplot coordinate system immediately as they are loaded into a Seurat object. With these changes, users will be able to subset objects by first determining a region of interest from SpatialPlots, then using GetTissueCoordinates to define cells of interest within this region.

**Reproducible Subsetting Example** (this example will produce the actual vs expected plots shown above if run on this [dataset](https://cf.10xgenomics.com/samples/spatial-exp/4.0.1/Visium_HD_3prime_Zebrafish_Head/Visium_HD_3prime_Zebrafish_Head_web_summary.html), with the actual plot resulting if you run this example before these changes):

```{R}
#Extract cellular coordinates for all cells
coords <- GetTissueCoordinates(object)

# Subset cell barcodes by x and y ranges
cells_keep <- rownames(coords)[
  coords[, "x"] >= 200 & coords[, "x"] <= 400 &
  coords[, "y"] >= 100 & coords[, "y"] <= 250
]

# Subset Seurat object to just those cells
object_subset <- subset(object, cells = cells_keep)
```

With this fix, the following should work:
- Subsetting Seurat objects containing Visium HD 3' data using GetTissueCoordinates
- Previous spatial plots involving other datasets (ie: Visium V2, Xenium) are unchanged

### DIFF
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -1538,11 +1538,12 @@ Read10X_Segmentations <- function (image.dir,
 {
 
   image <- png::readPNG(source = file.path(image.dir, image.name))
+  image.height <- dim(image)[1] 
   scale.factors <- Read10X_ScaleFactors(filename = file.path(image.dir,
                                                              "scalefactors_json.json"))
   key <- Key(slice, quiet = TRUE)
 
-  sf.data <- Read10X_HD_GeoJson(data.dir = data.dir, image.dir = image.dir, scale.factor = "lowres", image = image)
+  sf.data <- Read10X_HD_GeoJson(data.dir = data.dir, image.dir = image.dir, scale.factor = "lowres", image.height = image.height)
 
   # Create a Segmentation object based on sf, populate sf.data and polygons
   segmentation <- CreateSegmentation(sf.data)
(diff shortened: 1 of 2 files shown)


YoukaiFromAccounting:
@anashen Thanks for the suggestion! I’ve updated Read10X_HD_GeoJson to accept image.height as a parameter instead of the full image, and adjusted the documentation accordingly.

PR #10105, opened 2025-09-26
Fix incorrect box colors when `DimPlot()` with `split.by` and `label.box`

Billsfriend:
Due to improper color vector handling with splited `DimPlot()`, the color of `label.box` can look funny with `split.by`.
Example: `SeuratObject::pbmc_small |> DimPlot(label = T, label.box = T, split.by = 'groups')`
before_correction

It can be fixed with adding an `unique()` function to group-duplicated color vector.
after_correction

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -5834,7 +5834,7 @@ LabelClusters <- function(
       mapping = aes_string(x = xynames['x'], y = xynames['y'], label = id, fill = id),
       show.legend = FALSE,
       ...
-    ) + scale_fill_manual(values = labels.loc$color[order(labels.loc[, id])])
+    ) + scale_fill_manual(values = labels.loc$color[!duplicated(labels.loc[, id])])
   } else {
     geom.use <- ifelse(test = repel, yes = geom_text_repel, no = geom_text)
     plot <- plot + geom.use(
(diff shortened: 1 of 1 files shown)


R/visualization.R
```diff
@@ -6133,7 +6133,7 @@ LabelClusters <- function(
       mapping = aes_string(x = xynames['x'], y = xynames['y'], label = id, fill = id),
       show.legend = FALSE,
       ...
-    ) + scale_fill_manual(values = labels.loc$color[order(labels.loc[, id])])
+    ) + scale_fill_manual(values = unique(labels.loc$color[order(labels.loc[, id])]))
```
YoukaiFromAccounting, line 6136: scale_fill_manual() was replaced with scale_fill_identity() in the recent shift from aes_string to aes with the release of ggplot2 v4.0.0

YoukaiFromAccounting (commented):
With recent changes to aes interactions with ggplot2 v4.0.0, changing scale_fill_manual actually doesn't fix the issue of incorrect label box colours; testing these changes with aes rather than aes_string results in label boxes still being miscoloured:

image

For now, we anticipate sticking to scale_fill_identity rather than scale_fill_manual. Thank you for helping notice this issue!

PR #10115, opened 2025-10-02
Fix to have Load10X_Spatial recognize use.names for segmented data

chlige:
Currently, Load10X_Spatial will not pass the use.names argument when loading segmented data.   The result is that feature IDs are always gene symbols regardless if use.names is set or not.  This fix updates the Load10X_Spatial function so that use.names will be passed to Read10X_h5 when loading segmented data.


anashen (approved):
Thank you @chlige for catching this & for the PR!

PR #10142, opened 2025-10-17
Fix colouring by cells.highlight in SingleSpatialPlot

YoukaiFromAccounting:
Minor fix in SingleSpatialPlot logic; previously, the variable **col.by.plot** was created from the original col.by value (variable critical for determining which data column is used for colouring cell points). However, with the shift of plotting functions to support ggplot2 v4.0.0 (https://github.com/satijalab/seurat/pull/10116), col.by.plot was introduced to safely use tidyeval syntax in ggplot aesthetics without affecting downstream string operations.

### Error

Previously, when cells.highlight is specified in plotting, the function SingleSpatialPlot creates a new "highlight" column in the data and updates col.by <- "highlight". However, since col.by.plot was already created from the original col.by value, it never reflected this change. This would result in cryptic plotting errors like:
```{R}
Error in `palette()`:
! Insufficient values in manual scale. 14 needed but only 2 provided.
```

### Fix 
Moved the cells.highlight handling block to execute before creating col.by.plot, ensuring both variables reference the same column name and synchronizing plot aesthetics and color scales.

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -9463,13 +9463,7 @@ SingleSpatialPlot <- function(
     warning("Cannot find '", col.by, "' in data, not coloring", call. = FALSE, immediate. = TRUE)
     col.by <- NULL
   }
-  col.by.plot <- col.by %iff% data_sym(col.by) #had to create second variable to safely use tidyeval in plotting but not effect subsetting in gsub call later in function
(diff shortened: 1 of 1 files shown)


anashen (approved):
Thanks @YoukaiFromAccounting!

PR #10145, opened 2025-10-20
Change default value of verbose parameter for LeverageScore within sketchdata

Liuy12:
Change default value of verbose parameter for LeverageScore to use the supplied verbose. Otherwise, user will have no idea of the process especially when the sample size is large.


anashen (approved):
Thanks for catching this @Liuy12!

PR #10151, opened 2025-10-21
Update loading & visualization for Visium data

anashen:
## **A detailed description will be added here once marked ready for review.**

These changes (should) only apply to Visium objects.

- `preprocessing.R` - Update spatial object creation for binned data to ensure data matches image coordinate system
- `visualization.R` - Update visualization for spatial data to ensure both binned and segmented data match image coordinate system. Switch to using `geom_polygon` to plot segmentations.

To test, the `update-segmentation-methods` branch of SeuratObject must be installed from GitHub.

### DIFF
--- a/tests/testthat/test_load_10X.R
+++ b/tests/testthat/test_load_10X.R
@@ -83,7 +83,7 @@ test_that("Read10X_Image works as expected", {
     coordinates.expected <- Read10X_Coordinates(
       file.path(path.to.image, coordinate.filename),
       filter.matrix = TRUE
-    )[, c("imagerow", "imagecol")]
+    )[, c("imagecol", "imagerow")]
     colnames(coordinates.expected) <- c("x", "y")
     # read in the scale factors as an S3 object
     scale.factors.expected <- Read10X_ScaleFactors(
@@ -147,9 +147,9 @@ test_that("Read10X_Image works as expected", {
     # check that `coordinates` contains values scaled for the low resolution PNG
     # also make sure that it has the expected column names
     coordinates.expected.v1 <- coordinates.expected
-    colnames(coordinates.expected.v1) <- c("imagerow", "imagecol")
+    colnames(coordinates.expected.v1) <- c("imagecol", "imagerow")
     expect_equal(
-      coordinates[, c("imagerow", "imagecol")] / scale.factors[["lowres"]],
+      coordinates[, c("imagecol", "imagerow")] / scale.factors[["lowres"]],
       coordinates.expected.v1
     )
     # check that the spot size is similarly scaled


R/visualization.R
```diff
@@ -9551,119 +9545,90 @@ SingleSpatialPlot <- function(
...
-      #Plot (currently independently of switch/case)
-      if(!plot_segmentations){
-        #If plot_segmentations FALSE, then plot just the centroids from sf data
-        
+      
+      # Merge sf.data with expression data and centroid coordinates
+      
+      sf.plot <- sf.data %>%
+        left_join(data, by = c("barcodes" = "cell")) %>%
+        filter(barcodes %in% centroid_barcodes)
```
YoukaiFromAccounting: Currently, attempting to plot an object using SingleSpatialPlot with segmentations loaded in results in the following error:

```{R}
Error in filter(., barcodes %in% centroid_barcodes) : 
  ensembldb::filter requires an 'EnsDb' object as input. To call the filter function from the stats or dplyr package use stats::filter and dplyr::filter instead
```

R/preprocessing.R
```diff
@@ -1426,7 +1426,7 @@ Read10X_Image <- function(
 
   # Create an `sp` compatible `FOV` instance.
   fov <- CreateFOV(
-    coordinates[, c("imagerow", "imagecol")],
```
YoukaiFromAccounting, line 1429: Since coordinates are altered when loading in, this may necessitate a change to UpdateSeuratObject to prevent older Visium objects from having incorrect coordinate systems.

ie: loading SeuratData's mouse brain object gives:
image

anashen: Great point, thanks for bringing this up.

R/visualization.R
anashen: Will fix. Might just use base R instead of requiring/specifying dplyr (especially as it is not in imports, suggests, etc).

anashen: Fixed in f003bc0.

YoukaiFromAccounting (approved):
Plotting seems to work, all looks good on my end!

Tested the following:

- [x] Plotting VisiumV2, HD (binned), HD (centroids), and HD (segmentations) objects using SpatialDimPlot and SpatialFeaturePlot, on both hires and lowres tissue images
- [x] Subsetting VisiumV2, HD (binned), HD (centroids), and HD (segmentations) objects
- [x] Zooming in on SpatialDimPlots and SpatialFeaturePlots using coord_cartesian and coord_fixed 

The following are likely to require adjustments in the future following this update:

- [ ] Changes to existing coords used in GetTissueCoordinates in previous workflows
- [ ] Change to InteractiveSpatialPlot plotting system

anashen:
Closing. An updated PR (off of a more recent branch) will be created.

PR #10172, opened 2025-10-29
Fix Bug in IntegrateEmbeddings when Ref is specified

lldelisle:
Dear developpers,
I noticed that if I want to use IntegrateEmbeddings but I specify `dims.to.integrate` and `reference` I get an error:
```txt
Error in dimnames(x) <- dn : 
  length of 'dimnames' [2] not equal to array extent
```
I could find the origin of the problem. I also so that there was a similar commit: https://github.com/satijalab/seurat/commit/c9f8b5e9fec5c6d6e23ad021e9b9143195b605e9 but unfortunately only the case where reference was not specified was fixed.

You find here the fix for the case where `reference` is specified.


AtSourav (approved):

anashen (approved):
Thank you for catching this & for the fix @lldelisle!

lldelisle:
You're welcome, thanks for the great package.

Issue #10156, opened 2025-10-23
BUG: DimPlot `label.box` color fill broken with ggplot2 4.0.0


samuel-marsh:
### Issue Description

Hi Seurat Team (tagging @anashen @YoukaiFromAccounting here),

Following the merge of #10028 the `label.box` parameter in `DimPlot` (passed to `LabelClusters` no longer properly colors the filled boxes.  Given the changes made spatial plotting there I wanted to leave this one to Seurat team for how you would like to handle fix.  Issue comes in that the `scale_fill_manual` line was removed from `LabelClusters` if `box = TRUE`.  This is because the color column in the "labels.loc" data.frame no longer contains valid color values.  

<img width="2442" height="352" alt="Image" src="https://github.com/user-attachments/assets/76de979e-0fc1-4758-8698-3105ae7d89ec" />

Hence just restoring that line then spits the following error:

```
Error in `geom.use()`:
! Problem while converting geom to grob.
ℹ Error occurred in the 2nd layer.
Caused by error:
! Unknown colour name: -5.69291337379421
```

Again just wanted to leave this one with you all to figure out how to handle with the updates/changes to spatial plotting.

Thanks!
Best,
Sam

### Reproducing Code Example

```R
library(Seurat)
DimPlot(pbmc_small, label.box = T, pt.size = 0.25, label.size = 5, label = T, cols = DiscretePalette(n = 3, palette = "polychrome"))
```

### Error Message

```shell

```

### Session Info

```shell
> sessionInfo()
R version 4.5.0 (2025-04-11)
Platform: aarch64-apple-darwin20
Running under: macOS Sequoia 15.6.1

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.12.1

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: America/New_York
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[...]

PR #10198, opened 2025-11-10
Add color retrieval logic in LabelClusters for consistent cluster label coloring

YoukaiFromAccounting:
Minor fix to LabelClusters logic that addresses this [issue](https://github.com/satijalab/seurat/issues/10156). Previously, when generating cluster labels, the function relied on the data passed from ggplot_build() without ensuring that point colour information was retrieved correctly. With the shift from ggplot2 v3.5.2 to ggplot2 v4+, the colour values contained in the data column are retrieved incorrectly with LabelClusters, breaking functions that use LabelClusters (DimPlot, FeaturePlot, etc.).

### Error
Previously, when LabelClusters was called within functions like DimPlot that specify label.box or custom label colours:
Ie:
```{R}
DimPlot(pbmc_small, reduction = "tsne", label.box = T, label = T)
DimPlot(pbmc_small, label.box = T, pt.size = 2, label.size = 4, label = T, cols = DiscretePalette(n = 3, palette = "polychrome"))
```

The plot would either fail to generate, generate as a blank plot, or result in the following error (or similar):
```{R}
Error in `geom.use()`:
! Problem while converting geom to grob.
ℹ Error occurred in the 2nd layer.
Caused by error:
! Unknown colour name: -5.69291337379421
```

### Fix
This fix adds a small block in LabelClusters() to retrieve the correct colour mapping from the built ggplot object (pb$data[[1]]) and append it to the label data frame. This checks for either "colour" or "color" keys and binds the appropriate column as colour, instead of looking in the first dataframe column as was done previously. If no such column exists, an NA placeholder is used to maintain compatibility.

With these changes, the following commands work on both R version 4.4.2 and 4.5.1, running either ggplot2 v3.5.2 or v4+:
```{R}
DimPlot(pbmc_small, reduction = "tsne", label.box = T, label = T)
DimPlot(pbmc_small, label.box = T, pt.size = 2, label.size = 4, label = T, cols = DiscretePalette(n = 3, palette = "polychrome"))
```
image

image

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -6229,6 +6229,7 @@ LabelClusters <- function(
       data = labels.loc,
       mapping = aes(x = .data[[xynames['x']]], y = .data[[xynames['y']]], label = .data[[id]], fill = .data[["color"]]),
       show.legend = FALSE,
+      inherit.aes = FALSE,
       ...
     ) + scale_fill_identity()
   } else {
@@ -6237,6 +6238,7 @@ LabelClusters <- function(
       data = labels.loc,
       mapping = aes(x = .data[[xynames['x']]], y = .data[[xynames['y']]], label = .data[[id]]),
       show.legend = FALSE,
+      inherit.aes = FALSE,
       ...
     )
   }


anashen (approved):
Thanks @YoukaiFromAccounting, tested with pbmc dataset on both ggplot2 v3.5.2 and v4.0.0.

PR #10217, opened 2025-11-20
Fix typos in spatial and Visium HD vignettes

liuzhenyu-yyy:
Corrected several typos in the vignettes of spatial and Visium HD analysis.


anashen:
Thanks @liuzhenyu-yyy for the PR, your commit has been bundled into #10423.

Issue #10233, opened 2025-12-09
latest umap-learn 0.5.9.post2 crashing version control in RunUMAP


daniel-spies:
### Issue Description

Dear Seurat Team,

I've encountered a special case in which RunUMAP with 'umap-learn' is crashing as the check for the version >= 0.5.0 is failing as the latest conda umap version is "0.5.9.post2" and the non-numerical part is messing up the version comparison.

best
daniel

### Reproducing Code Example

```R
so <- RunUMAP(so,dims = pc.num, metric = "correlation",umap.method = "umap-learn")
```

### Error Message

```shell
Error: invalid version specification ‘0.5.9.post2’
```

### Session Info

```shell
R version 4.5.2 (2025-10-31)
Platform: x86_64-pc-linux-gnu
Running under: Debian GNU/Linux 12 (bookworm)

Matrix products: default
BLAS:   /nfs/scistore07/clustersw/debian/bookworm/R/4.5.2/lib/R/lib/libRblas.so 
LAPACK: /nfs/scistore07/clustersw/debian/bookworm/R/4.5.2/lib/R/lib/libRlapack.so;  LAPACK version 3.12.1

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8      
 [8] LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

time zone: Europe/Vienna
tzcode source: system (glibc)

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] future_1.67.0      patchwork_1.3.2    Seurat_5.3.1       SeuratObject_5.2.0 sp_2.2-0           lubridate_1.9.4    forcats_1.0.1      stringr_1.6.0      dplyr_1.1.4        purrr_1.2.0       
[11] readr_2.1.5        tidyr_1.3.1        tibble_3.3.0       ggplot2_4.0.0      tidyverse_2.0.0   

loaded via a namespace (and not attached):
[...]

PR #10246, opened 2025-12-15
Fix UMAP package versioning issue

YoukaiFromAccounting:
# Issues

- https://github.com/satijalab/seurat/issues/10233

## Issue 1

RunUMAP relies on the UMAP package (https://github.com/lmcinnes/umap), which has its version determined by the base R utils function numeric_version(). This function cannot parse any version names with letters in them, which the latest UMAP release has: 

```{R}
> reticulate::py_run_string("import umap; print(umap.__version__)") 
0.5.9.post2
```

This results in the following error when trying to use RunUMAP:

```{R}
Error: invalid version specification ‘0.5.9.post2’
```

This fix addresses this issue by simply parsing the version to remove any versioning other than the major/minor/patch version numbers before calling numeric_version(). With this fix, RunUMAP should work with newer release versionings for the UMAP package.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.4.0.9000
+Version: 5.4.0.9001
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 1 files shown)


anashen:
Thanks @YoukaiFromAccounting! 

Tested with `pbmc_small`:

```R
library(reticulate)
conda_create("r-reticulate", python_version = "3.9")
conda_install("r-reticulate", "umap-learn=0.5.9.post2")

# In new R session -
use_condaenv("r-reticulate")
data("pbmc_small")
pbmc_small <- RunUMAP(object = pbmc_small, dims = 1:5, metric = "correlation", umap.method = "umap-learn")
```

Previously, pre-releases for `umap-learn` were tagged with suffix "rc" (ex. `0.5.0rc1`), which is also parsed correctly with the new changes.

PR #10257, opened 2025-12-23
Fix link to NEWS file in README

mihem:
The link to NEWS file in README was still ponting to the master branch instead of the main branch


anashen:
Thanks for the PR, but looks like we fixed this in a later commit (b4e5b10)

Issue #10261, opened 2025-12-27
SpatialPlot error on VisiumV1 object


assaron:
### Issue Description

Hi,

Recently the following code stopped working for me:
```{r}
library(Seurat)

obj <- readRDS(url("https://alserglab.wustl.edu/files/fgsea/275_T_seurat.rds"))
SpatialPlot(obj)
```

It gives an error:
```
Please run `UpdateSeuratObject` on your Seurat object first to ensure that data aligns to the image slice1 when plotting.
```

However, when I use the update function, nothing changes:
```
> obj1 <- UpdateSeuratObject(obj)
Validating object structure
Updating object slots
Ensuring keys are in the proper structure
Updating matrix keys for DimReduc ‘pca.rev’
Ensuring keys are in the proper structure
Ensuring feature names don't have underscores or pipes
Updating slots in Spatial
Updating slots in SCT
Updating slots in pca.rev
Updating slots in slice1
Setting assay used for SCTransform.Spatial to Spatial
Setting assay used for RunPCA.SCT to SCT
Validating object structure for Assay ‘Spatial’
Validating object structure for SCTAssay ‘SCT’
Validating object structure for DimReduc ‘pca.rev’
Validating object structure for VisiumV1 ‘slice1’
Object representation is consistent with the most current Seurat version
> SpatialPlot(obj1)
Error in SpatialPlot(obj1) : 
  Please run `UpdateSeuratObject` on your Seurat object first to ensure that data aligns to the image slice1 when plotting.

```

### Reproducing Code Example

```R
library(Seurat)

obj <- readRDS(url("https://alserglab.wustl.edu/files/fgsea/275_T_seurat.rds"))
SpatialPlot(obj)
```

### Error Message

```shell
Please run `UpdateSeuratObject` on your Seurat object first to ensure that data aligns to the image slice1 when plotting.
```

### Session Info

```shell
R version 4.5.1 (2025-06-13)
Platform: x86_64-pc-linux-gnu
Running under: Debian GNU/Linux 13 (trixie)

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3 
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.21.so;  LAPACK version 3.11.0

locale:
[...]

PR #10267, opened 2026-01-06
Fix spatial plots with VisiumV1 + Seurat v5.4.0

anashen:
Fixes https://github.com/satijalab/seurat/issues/10261.

With the changes in Seurat v5.4.0 to load Visium (V2) data in the correct order with `imagecol` corresponding to the "y" position and `imagerow` corresponding to the "x" position, a check was added for spatial plots that would alert the user to run `UpdateSeuratObject` if their object appeared to have the old orientation of coordinates.

This is not applicable to Visium V1 images, as the `coordinates` dataframe uses the `imagecol` and `imagerow` columns directly without re-assigning to an `x` or `y` column.

To fix this error (1) and make sure spatial plots appear in the correct orientation (2), this PR proposes the following:

1) Update `SpatialPlot` to restrict the coordinate flip check to Visium V2 images only.
2) Update `GetTissueCoordinates.VisiumV1` to return the coordinate columns in the correct order (`imagecol`, `imagerow`)

```R
# Seurat object sourced from https://github.com/satijalab/seurat/issues/10261 
> v1_obj <- readRDS("~/visiumv1_275.rds")
> SpatialPlot(v1_obj, pt.size.factor = 100)
> ggsave("1.png")
# Saving 7 x 7 in image
> SpatialDimPlot(v1_obj, group.by = "orig.ident", pt.size.factor = 100)
> ggsave("2.png")
# Saving 7 x 7 in image
> SpatialFeaturePlot(v1_obj, features = "nCount_Spatial", pt.size.factor = 100)
> ggsave("3.png")
# Saving 7 x 7 in image
```
<p float="left">
  
   
  
</p>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.4.0.9001
+Version: 5.4.0.9002
 Title: Tools for Single Cell Genomics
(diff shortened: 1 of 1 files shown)


YoukaiFromAccounting (approved):
Looks good to merge! Tested with two older existing VisiumV1 objects, plotting and coordinates are set correctly. _Requires small change to UpdateSeuratObject in SeuratObject to account for updating non-FOV-type objects like VisiumV1._

EDIT: UpdateSeuratObject does not need to be changed as a result of this PR, VisiumV1 objects now do not rely on the coords_x_orientation slot to be updated for correct plotting.

PR #10285, opened 2026-01-21
Fix error in `FindSpatiallyVariableFeatures` from use of `slot` in function cascade

anashen:
Pass `layer` instead of `slot` to `FindSpatiallyVariableFeatures.Assay` call in `FindSpatiallyVariableFeatures`. 

Currently because the call uses `slot`, the default layer (`scale.data`) is used instead of the layer originally passed, resulting in an error if that layer is empty. See below (example from spatial vignette):

```R
> cortex <- FindSpatiallyVariableFeatures(cortex, 
+                                         assay = "predictions", 
+                                         selection.method = "moransi", 
+                                         features = rownames(cortex), 
+                                         r.metric = 5, 
+                                         layer = "data")
# Warning: Layer ‘scale.data’ is empty
# Error in `colnames<-`(x = `*tmp*`, value = paste0("MoransI_", colnames(x = svf.info))) : 
#  attempt to set 'colnames' on an object with less than two dimensions
> traceback()
# 6: stop("attempt to set 'colnames' on an object with less than two dimensions")
# 5: `colnames<-`(x = `*tmp*`, value = paste0("MoransI_", colnames(x = svf.info))) at preprocessing.R#4757
# 4: FindSpatiallyVariableFeatures.Assay(object = object[[assay]], 
       # slot = slot, features = features, spatial.location = tc, 
       # selection.method = selection.method, r.metric = r.metric, 
       # x.cuts = x.cuts, y.cuts = y.cuts, nfeatures = nfeatures, 
       # verbose = verbose, ...) at generics.R#244
# 3: FindSpatiallyVariableFeatures(object = object[[assay]], slot = slot, 
       # features = features, spatial.location = tc, selection.method = selection.method, 
       # r.metric = r.metric, x.cuts = x.cuts, y.cuts = y.cuts, nfeatures = nfeatures, 
       # verbose = verbose, ...) at preprocessing.R#4809
# 2: FindSpatiallyVariableFeatures.Seurat(cortex, assay = "predictions", 
       # selection.method = "moransi", features = rownames(cortex), 
       # r.metric = 5, layer = "data") at generics.R#244
# 1: FindSpatiallyVariableFeatures(cortex, assay = "predictions", 
       # selection.method = "moransi", features = rownames(cortex), 
       # r.metric = 5, layer = "data")
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.4.0.9007
+Version: 5.4.0.9008
 Title: Tools for Single Cell Genomics
 Description: A toolkit for quality control, analysis, and exploration of single cell RNA sequencing data. 'Seurat' aims to enable users to identify and interpret sources of heterogeneity from single cell transcriptomic measurements, and to integrate diverse types of single cell data. See Satija R, Farrell J, Gennert D, et al (2015) <doi:10.1038/nbt.3192>, Macosko E, Basu A, Satija R, et al (2015) <doi:10.1016/j.cell.2015.05.002>, Stuart T, Butler A, et al (2019) <doi:10.1016/j.cell.2019.05.031>, and Hao, Hao, et al (2020) <doi:10.1101/2020.10.12.335331> for more details.
 Authors@R: c(


anashen:
Just adding a note here that lines 4795-4802 take care of non-NULL `slot` arguments in the call, so the fix to specify `layer` instead of `slot` in subsequent calls will not break existing code.

https://github.com/satijalab/seurat/blob/afb7654e3ace55d413dbbfde814c8156ed6c138c/R/preprocessing.R#L4795-L4802

PR #10298, opened 2026-02-05
Allow unnamed cols vector to be used in SpatialPlots

YoukaiFromAccounting:
Fixes https://github.com/satijalab/seurat/discussions/6083

Previously, when plotting with SpatialPlot or SpatialDimPlot, users would have to pass in a named vector to the cols parameter, which is a vector of colours with each colour corresponding to an identity class (eg: 4 colours corresponding to 4 different clusters in the object). However, if this vector was not named, then the user would encounter the following error:

```
Error in `palette()`:
! Insufficient values in manual scale. 10 needed but only 0 provided.
Run `rlang::last_trace()` to see where the error occurred.
```

This is because SingleSpatialPlot specifically checks for named values in cols via ```cols <- cols[names(cols) %in% vals]```, with cols being set to NULL if no names are provided. These changes allow the user to pass a non-named vector to cols, with the colours being set to identities in order. 

Reproducible example: 
```
library(SeuratData)
brain <- LoadData("stxBrain", type = "anterior1")

DefaultAssay(brain) <- "Spatial"
brain <- SCTransform(brain, assay = "Spatial", verbose = FALSE)
brain <- RunPCA(brain, verbose = FALSE)
brain <- FindNeighbors(brain, dims = 1:20, verbose = FALSE)
brain <- FindClusters(brain, resolution = 0.3, verbose = FALSE)

#Set of custom colours
custom <- c(
  "#1f77b4",
  "#ff7f0e",
  "#2ca02c",
  "#d62728",
  "#9467bd",
  "#8c564b",
  "#e377c2",
  "#7f7f7f",
  "#bcbd22",
  "#17becf"
)

#Plotting would error without these changes
SpatialPlot(brain, cols = custom)
```

With these changes, SpatialPlot(cols = custom) now achieves the same result as using scale_fill_manual(values = custom):
image

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -9734,21 +9734,19 @@ SingleSpatialPlot <- function(
       # cols must be a named vector 
       # If cols is unnamed, then names(cols) is NULL
       # Cols is then subsetted down to length 0 
-      if (has_valid_names) {
-        cols <- cols[names(cols) %in% vals]
-      } else {
-        if (length(cols) < n_groups) {
-          warning(
-                    "SpatialPlot: `cols` is being treated as a manual mapping, but it has no valid names. ",
-                    "Seurat will match colors to groups by name, so you should pass a named vector whose names ",
-                    "match the values of `", col.by, "`, for example: setNames(cols, sort(unique(data[[col.by]])))."
-          )
-          cols <- rep_len(cols, n_groups)
-        } else {
-          cols <- cols[seq_len(n_groups)]
+     if (has_valid_names) {
+        cols <- cols[vals]
+        if (anyNA(cols)) {
+          warning("Missing color mappings for ", paste(vals[is.na(cols)], collapse = ", "))
         }
+      } else {
+        if (length(cols) != n_groups) {
+          warning("Number of colors does not match number of groups; adjusting.")
+        }
+        cols <- rep_len(cols, n_groups)
         names(cols) <- vals
       }
+      scale <- scale_fill_manual(values = cols, na.value = na.value)
       scale <- scale_fill_manual(values = cols, na.value = na.value)
     }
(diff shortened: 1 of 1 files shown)


R/visualization.R
```diff
@@ -9720,12 +9720,35 @@ SingleSpatialPlot <- function(
...
+                    "SpatialPlot: `cols` is being treated as a manual mapping, but it has no valid names. ",
+                    "Seurat will match colors to groups by name, so you should pass a named vector whose names ",
+                    "match the values of `", col.by, "`, for example: setNames(cols, sort(unique(data[[col.by]])))."
+          )
+          cols <- rep_len(cols, n_groups)
+        } else {
+          cols <- cols[seq_len(n_groups)]
+        }
+        names(cols) <- vals
+      }
```
anashen:
```suggestion
     if (has_valid_names) {
        cols <- cols[vals]
        if (anyNA(cols)) {
          warning("Missing color mappings for ", paste(vals[is.na(cols)], collapse = ", "))
        }
      } else {
        if (length(cols) != n_groups) {
          warning("Number of colors does not match number of groups; adjusting.")
        }
        cols <- rep_len(cols, n_groups)
        names(cols) <- vals
      }
      scale <- scale_fill_manual(values = cols, na.value = na.value)
```

anashen: Changing to `cols <- cols[vals]` produces NAs where color mappings are not defined, so that we can alert the user accordingly (this refers to the example in my review).

Pasted Graphic 1

YoukaiFromAccounting: Thanks @anashen, these changes look good!

anashen:
Thanks @YoukaiFromAccounting!

PR #10302, opened 2026-02-09
fix issues with image orientation for curio-seeker data

Liuy12:
Fix issues with image orientation for curio-seeker data for the function LoadCurioSeeker. According to the source code from curio-seeker's pipeline, to match the tissue coordinates with the original image, the following is implemented:

1. Flip x and y coordinates from the *MatchedBeadLocation.csv file. 
2. Update x to be max(x) - x


anashen:
Hi @Liuy12, are you able to provide an example for testing?

Issue #8225, opened 2023-12-27
Featureplot not generated right plot when using raster=T


shanshenbing:
```r
#when raster=T
FeaturePlot(object = srobj,
            features =c('SOX2'),
            order = T,cols = c(c('lightblue','white', 'orangered')),
            min.cutoff = 'q10',max.cutoff = 'q90',raster = T)
#when raster=F
FeaturePlot(object = srobj,
            features =c('STAT3'),
            order = T,cols = c('lightblue','white', 'orangered'),
            min.cutoff = 'q10',max.cutoff = 'q90',raster = F)

```
![image](https://github.com/satijalab/seurat/assets/42245493/66342b67-a3bb-4c16-9ffa-c11d07411b7c)
![image](https://github.com/satijalab/seurat/assets/42245493/df7d5b32-4e9d-4f06-a171-67133a9e5a84)

Gesmira:
Hi, 
Can you create a reproducible example of this issue with a SeuratData object? Do you only see this with large objects?

shanshenbing:
You can try a SeuratData object with more than 10w cells, and run Featureplot. This error occurs in large SeuratData object.

domi84:
I have exactly the same issue on Seurat 5.2/5.3.
FeaturePlot of a Seurat object with more than 100,000 cells gives a completely different result if raster=T or F. It seems order is ignored when raster is T.

`FeaturePlot(complete.2, "HPGD", order = T, pt.size = 3, raster = T)`

![Image](https://github.com/user-attachments/assets/982ebc24-90e9-47ed-8168-ea9fe02349bd)

`FeaturePlot(complete.2, "HPGD", order = T, pt.size = 3, raster = F)`

![Image](https://github.com/user-attachments/assets/dd0f1dd2-3646-4e2b-aca6-3a2b098b7600)

pormr:
You can try to plot positive cells separately (replace GENE, CUTOFF and PT.SIZE with appropriate values):
``` R
FeaturePlot(object = srobj, features = 'GENE', order = TRUE, raster = TRUE) +
  scattermore::geom_scattermore(
    aes(color = GENE, x = umap_1, y = umap_2),
    data = ~ filter(., GENE > CUTOFF),
    pointsize = PT.SIZE
  )
  ...
```
To prevent duplication of cells, you need to pull necessary data from the dataset and create the plot manually.

carolienv:
[...]

PR #10305, opened 2026-02-13
Add singledimplot rasterwarning

YoukaiFromAccounting:
Related to: https://github.com/satijalab/seurat/issues/8225

Small change adding warning to SingleDimPlot when users set both raster and order to TRUE; scattermore cannot resolve orderings of cells during the rasterization process, and thus points cannot be simultaneously ordered by expression value and rasterized. 

Example of plotting with DimPlot without rasterization (LEFT) and with rasterization (RIGHT). Points lose their order when plotting with rasterization:
image image

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -8661,6 +8661,11 @@ SingleDimPlot <- function(
   raster <- raster %||% (nrow(x = data) > 1e5)
   pt.size <- pt.size %||% AutoPointSize(data = data, raster = raster)
 
+  if (is.null(raster) && is.null(order)){
+    message("Points cannot be ordered and rasterized at the same time.",
+            "\nPoint order will be randomized in raster image.")
+  }
+
   if (is.null(x = stroke.size)) {
     stroke.size <- 0.600075815011372
   }


pormr:
Could you please check [#9943 (discussion)](https://github.com/satijalab/seurat/discussions/9943)? A new global option can be introduced to choose default rasterization engine (just like `options(Seurat.object.assay.version = "v3")`). Alternatively, we can also suggest users to apply rasterization to an existing ggplot object using `+ ggrastr::rasterize(layers = "Point", dpi = 300)`. Thank you for considering my suggestion!

YoukaiFromAccounting:
@pormr Thank you very much for the suggestion! In lieu of adding additional global options which may be confusing to newer users, we will add the option whereby users can plot Dim/FeaturePlots with ggrastr::geom_point_rastr if both raster = T and order = T. In this way, if users specify that they want to maintain cell expression ordering, Seurat will use the plotting mechanism that maintains ordering, albeit at slower plotting speeds.

YoukaiFromAccounting:
Closing for now, see: https://github.com/satijalab/seurat/pull/10307

PR #10307, opened 2026-02-18
Add alternative ggrastr-based rasterization for Dim/Feature Plots

### LINKED ISSUE #8225: Featureplot not generated right plot when using raster=T

YoukaiFromAccounting:
Fixes: https://github.com/satijalab/seurat/issues/8225 

Previously, when plotting with DimPlot or FeaturePlot, users are able to set the "order" flag to True, which would ensure that points with higher values (eg: cell expression levels) are plotted atop those with lower values. However, with the adoption of geom_scattermore-based rasterization, which does not support cell order within its blending algorithm, users are unable to have both a rasterized plot with correct expression-level ordering. 

This PR addresses this issue by using ggraster::geom_point_rastr; this command rasterizes the point layer into an image when rendering the plot, and importantly maintains point order based on values. Users are able to use the geom_point_rastr render within DimPlot/FeaturePlot calls by specifying both raster = T and order = T. 

Below are plots of single-cell gene expression with these changes. The left plot was generated with ``` FeaturePlot(object, features = "rho", order = T)``` and the right plot was generated with ```FeaturePlot(object, features = "rho")```. Importantly, since the number of points within these plots exceed 100k, rasterization is used by default. With our changes, the left plot uses ggrastr::geom_point_rastr and correctly orders points by expression, while the right plot continues to use geom_scattermore and fails to order points correctly (many cells with high expression are hidden).

image

Reproducible example:

```
library(Seurat)
library(SeuratData)
library(ggrastr)
library(ggplot2)

InstallData("stxBrain")
brain <- LoadData("stxBrain", type = "anterior1")

brain <- SCTransform(brain, assay = "Spatial", verbose = FALSE)
brain <- RunPCA(brain, assay = "SCT", verbose = FALSE)

p1 <- FeaturePlot(brain, features = "Hpca", pt.size = 3, raster = T, order = T)
p2 <- FeaturePlot(brain, features = "Hpca", pt.size = 3, raster = T)
p1 + p2
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.4.0.9011
+Version: 5.4.0.9012
 Title: Tools for Single Cell Genomics
 Description: A toolkit for quality control, analysis, and exploration of single cell RNA sequencing data. 'Seurat' aims to enable users to identify and interpret sources of heterogeneity from single cell transcriptomic measurements, and to integrate diverse types of single cell data. See Satija R, Farrell J, Gennert D, et al (2015) <doi:10.1038/nbt.3192>, Macosko E, Basu A, Satija R, et al (2015) <doi:10.1016/j.cell.2015.05.002>, Stuart T, Butler A, et al (2019) <doi:10.1016/j.cell.2019.05.031>, and Hao, Hao, et al (2020) <doi:10.1101/2020.10.12.335331> for more details.
 Authors@R: c(
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -8654,6 +8654,16 @@ SingleDimPlot <- function(
   raster = NULL,
   raster.dpi = NULL
 ) {
+  # IF both raster and order are TRUE, points are plotted with ggraster::geom_point_rast to maintain 
+  # correct ordering of points
+  if (isTRUE(raster) && isTRUE(order)){
+    # Check if ggrastr installed correctly
+    if (!is.null(x = raster) && isTRUE(x = raster)){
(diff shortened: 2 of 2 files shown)


anashen (approved):
Thanks @YoukaiFromAccounting - changes look good.

EduardAnsaldo:
The new code works to fix the order, however it ignores the 'raster.dpi' parameter if I'm not mistaken. Adding it as a dpi parameter to the ggrastr::rasterize calls should restore the correct dpi behavior.

pormr:
> The new code works to fix the order, however it ignores the 'raster.dpi' parameter if I'm not mistaken. Adding it as a dpi parameter to the ggrastr::rasterize calls should restore the correct dpi behavior.

@EduardAnsaldo ggrastr's `dpi` parameter is one-dimensional compared to `raster.dpi` in `SingleDimPlot()`, which is a two-dimensional vector (corresponding to `pixels` in `scattermore::geom_scattermore()`). It seems a little confusing, but it's in accordance with the document.

https://github.com/satijalab/seurat/blob/79c466c0e61e89aa299ee767fec3b69afda47301/R/visualization.R#L8727-L8728

PR #10329, opened 2026-03-13
Warning message for reference SCT model counts correction

### LINKED ISSUE #10304: SCTransform with reference.SCT.model and UMI correction
Hi,
I realized that when using SCTransform using a reference model (using `reference.SCT.model`), the UMI counts, and the normalized `data` slot is not corrected even if we use the option do.correct.umi=TRUE. The lines responsible of this behaviour are 
[here](https://github.com/satijalab/seurat/blob/1693636c68cc9e2fcd495da2f8df58ec75e60d6c/R/preprocessing.R#L4160-L4161).
This is misleading when doing downstream analysis, e.g. Using AddModuleScore on the SCTransform normalized `data`, regarding in this case, the `data` slot of SCTTransform assay is not normalized at all.

Thank you for your help, and for this great package and continuous maintenance.

mccormickceci:
Adding a warning message so that users know if they use a reference SCT model we will hard set do.correct.umi to FALSE. We do this since count correction across models is not statistically sound, but we will send a message so that users know their counts will not be corrected.

### DIFF
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -4160,6 +4160,9 @@ SCTransform.Assay <- function(
   if (!is.null(reference.SCT.model)){
     do.correct.umi <- FALSE
     do.center <- FALSE
+    message(
+      "NOTE:A reference SCT model was provided, therefore counts are NOT corrected (Regardless of argument passed to do.correct.umi)"
+    )
   }
 
   umi <- GetAssayData(object = object, layer = 'counts')
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -4160,8 +4160,8 @@ SCTransform.Assay <- function(
   if (!is.null(reference.SCT.model)){
     do.correct.umi <- FALSE
     do.center <- FALSE
-    message(
-      "NOTE:A reference SCT model was provided, therefore counts are NOT corrected (Regardless of argument passed to do.correct.umi)"
+    warning(
+      "A reference SCT model was provided, therefore counts are NOT corrected (Regardless of argument passed to do.correct.umi)"
     )
   }
 


anashen (changes requested):
@mccormickceci Thanks for the PR! Can you use `warning` instead of `message`? You can also remove the "NOTE:" prefix in the warning text.

anashen (approved):

anashen:
Just noting here that one of the tests in `test_integration.R` now gives this warning - we should consider modifying it to catch and expect this.

```
Warning ('test_integration.R:307:3'): FindTransferAnchors with default SCT works
A reference SCT model was provided, therefore counts are NOT corrected (Regardless of argument passed to do.correct.umi)
Backtrace:
     ▆
  1. └─Seurat::FindTransferAnchors(...) at test_integration.R:307:3
  2.   └─Seurat:::ValidateParams_FindTransferAnchors(...) at seurat/R/integration.R:779:3
  3.     ├─Seurat::SCTransform(...) at seurat/R/integration.R:6101:7
  4.     └─Seurat:::SCTransform.Seurat(...) at seurat/R/generics.R:761:3
  5.       ├─Seurat::SCTransform(...) at seurat/R/preprocessing.R:4274:3
  6.       └─Seurat:::SCTransform.StdAssay(...) at seurat/R/generics.R:761:3
  7.         └─base::lapply(...) at seurat/R/preprocessing5.R:1150:3
  8.           └─Seurat (local) FUN(X[[i]], ...)
  9.             ├─Seurat::SCTransform(...) at seurat/R/preprocessing5.R:1154:7
 10.             └─Seurat:::SCTransform.Assay(...) at seurat/R/generics.R:761:3
 ```

Issue #10101, opened 2025-09-22
BUG: VlnPlot -- Error in deparse(substitute(e2, env = caller_env(2))) : 'S4SXP'...


daddabi:
I wanted to check the quality of my data with violinplot: 
`VlnPlot(obj, features = 'nCount_RNA')`
and I got the error:
> Error in deparse(substitute(e2, env = caller_env(2))) : 
>   'S4SXP': should not happen - please report
> Error during wrapup: 'S4SXP': should not happen - please report
> Error: no more error handlers available (recursive errors?); invoking 'abort' restart

I have never encountered this problem before. I am also using my raw data, I did it before and it worked. But now for every seurat object (integrated/raw) I got this error when I make violinplot. I have no problems with other commands.

ksoleary:
I was having trouble with this yesterday and just realized ggplot2 was updated to 4.0.0 a few weeks ago. I inadvertently updated it when installing another package and started getting this error. It seems ggplot2 4.0.0 is incompatible with Seurat right now. Downgrading to 3.5.2 worked for me! I have a feeling this ggplot update will break a lot of packages with internal plotting functions...

Here is what I ran to downgrade the package 

`install.packages("remotes")  #if needed
library(remotes)
install_version("ggplot2", version = "3.5.2", repos = "https://cran.r-project.org")`

Then restart R and you should be good to go.

anashen:
Thank you for the report @daddabi & @ksoleary! We are currently looking into this. 

Would you be able to provide the result of running `sessionInfo`? It would also be very helpful if you could describe the type & structure of the object you are working with.

- - -
In the meantime, if any users encounter this error, please report here (and try downgrading ggplot2 as in the above post).

alizeevd:
Hello,
I had the same issue.
Downgrading to ggplot2  3.5.2 worked for me. 
Thank you for the tips !

HXD3D0323:
Hello,
I had the same issue.
Downgrading to ggplot2 3.5.2 worked for me. （my pkg version: seurat 4.4.0 seuratobject 4.1.4 or >5.0.0）
Thank you for the tips !

bilibili-fuyou:
Hello,
If I downgrade ggplot2 to 3.5.2, then when I load the clusterProfiler package, I get the following error. That means I may also need to downgrade other R packages that depend on ggplot2, which would make my R environment quite messy. Is there another way to solve this?
R version 4.3.3
library(clusterProfiler)
[...]

PR #10333, opened 2026-03-16
Update layer-building logic in E x I plots for ggplot2 v4+ compatibility

anashen:
Expression-by-identity plots (VlnPlot, RidgePlot, etc) are built with `SingleExIPlot` and `MultiExIPlot`. The addition of plot layers has been handled in previous versions dynamically via `do.call("+", ...)`.

ggplot2 v4 (September 2025) [transitioned to the S7 OOP system](https://tidyverse.org/blog/2025/09/ggplot2-4-0-0/). With the new update, some users (#10101) began encountering an error when running `VlnPlot` or `RidgePlot` with the following message[^1]:

```
Error in deparse(substitute(e2, env = caller_env(2))) :
'S4SXP': should not happen - please report
Error during wrapup: 'S4SXP': should not happen - please report
```

Since `+` is a builtin primitive operator whose method dispatch is handled in the R evaluator, `do.call` can behave differently from a direct call (`plot + geom`) when working with S4/S7 objects. So the fix here is to build plots iteratively, instead of passing all layers to be added to `do.call`. 

**Below: Reprex 1 demonstrates where/how exactly the problem occurs, and reprex 2 demonstrates that the change proposed in this branch fixes the issue in `VlnPlot` and `RidgePlot`.**

[^1]: Seemed to only occur in older R versions; there were no known `deparse` errors when using the `do.call` method for plot generation with ggplot2 v4+ in R 4.5.0+.

# reprex 1
``` r
library(ggplot2)
library(sessioninfo)

###
R.version$version.string
#> [1] "R version 4.3.2 (2023-10-31)"
packageVersion("ggplot2")
#> [1] '4.0.2'
###

p <- ggplot(iris, aes(x = Species, y = Petal.Length, fill = Species))

S7::S7_inherits(p)
#> [1] TRUE
class(p)
#> [1] "ggplot2::ggplot" "ggplot"          "ggplot2::gg"     "S7_object"      
#> [5] "gg"

layer1 <- geom_violin(scale = 'width', trim = TRUE)
layer2 <- theme(axis.text.x = element_text(angle = 45, hjust = 1))

S7::S7_inherits(layer1)
#> [1] FALSE
class(layer1)
#> [1] "LayerInstance" "Layer"         "ggproto"       "gg"

S7::S7_inherits(layer2)
#> [1] TRUE
class(layer2)
#> [1] "theme"          "ggplot2::theme" "gg"             "S7_object"

geom <- list(layer1, layer2)

do.call(what = '+', args = list(p, geom))
#> Error in deparse(substitute(e2, env = caller_env(2))): 'S4SXP': should not happen - please report

for (layer in geom) {
  p <- p + layer
}

p
```

<sup>Created on 2026-03-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

# reprex 2

``` r
test_ggplot2_v4_compat <- function(path_to_seurat) {
  devtools::load_all(path_to_seurat, quiet = TRUE)
  print(git2r::repository_head(repo = path_to_seurat))
  p1 <- try(VlnPlot(pbmc_small, features = "nCount_RNA"))
  p2 <- try(RidgePlot(object = pbmc_small, features = 'PC_1'))
  return(list("vln" = p1, "ridge" = p2))
}

invisible(test_ggplot2_v4_compat("~/seurat-testing-clone/seurat"))
#> [350dfa] (Local) (HEAD) main
#> Error in deparse(substitute(e2, env = caller_env(2))) : 
#>   'S4SXP': should not happen - please report
#> Error in deparse(substitute(e2, env = caller_env(2))) : 
#>   'S4SXP': should not happen - please report

res <- test_ggplot2_v4_compat("~/seurat")
#> [9e70e0] (Local) (HEAD) fix/ggplot2-v4-compat
gridExtra::grid.arrange(res$vln, res$ridge, ncol = 2)
#> Picking joint bandwidth of 0.387
```

<sup>Created on 2026-03-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

More info:

- https://stat.ethz.ch/R-manual/R-devel/library/methods/html/Methods_for_S3.html
  - > The R evaluator will ‘dispatch’ a method from a function call either when the body of the function calls the special primitive [UseMethod](https://stat.ethz.ch/R-manual/R-devel/library/base/html/UseMethod.html) or when the call is to one of the builtin primitives such as the math functions or the binary operators.
- https://adv-r.hadley.nz/base-types.html#base-types-2
- https://cran.r-project.org/doc/manuals/r-devel/R-ints.html#Argument-evaluation-1

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.4.0.9012
+Version: 5.4.0.9013
 Title: Tools for Single Cell Genomics
 Description: A toolkit for quality control, analysis, and exploration of single cell RNA sequencing data. 'Seurat' aims to enable users to identify and interpret sources of heterogeneity from single cell transcriptomic measurements, and to integrate diverse types of single cell data. See Satija R, Farrell J, Gennert D, et al (2015) <doi:10.1038/nbt.3192>, Macosko E, Basu A, Satija R, et al (2015) <doi:10.1016/j.cell.2015.05.002>, Stuart T, Butler A, et al (2019) <doi:10.1016/j.cell.2019.05.031>, and Hao, Hao, et al (2020) <doi:10.1101/2020.10.12.335331> for more details.
 Authors@R: c(


anashen:
Thanks @YoukaiFromAccounting!

PR #10344, opened 2026-03-24
Follow-up to #10307

anashen:
See #10307 for more details.

This follow up accounts for `DimPlot` taking `order` as a vector rather than a logical. Also added a warning that is shown once per session. Appears as `Seurat uses ggrastr::geom_point_rast to maintain point order with rasterization. This warning is displayed once per session.`

```R
group1 <- c("A", "B", "C")
pbmc_small$group1 <- sample(group1, ncol(pbmc_small), replace = TRUE)

dp <- DimPlot(pbmc_small, group.by="group1", order=c("C", "A", "B"), raster=T, pt.size=7, cols=c("black", "blue", "lightgray"))

fp <- FeaturePlot(pbmc_small, features=head(rownames(pbmc_small)), order=T, raster=T)
```

### DIFF
--- a/R/visualization.R
+++ b/R/visualization.R
@@ -8786,10 +8786,10 @@ SingleDimPlot <- function(
         pixels = raster.dpi
       )
     } else {
-      rlang::warn(message = "Seurat uses ggrastr::geom_point_rast to maintain point order with rasterization.", 
+      rlang::warn(message = "Seurat uses ggrastr::rasterise to maintain point order with rasterization.", 
                   .frequency = "once",
-                  .frequency_id = "Seurat-geom_point_rast")
-      plot + ggrastr::geom_point_rast(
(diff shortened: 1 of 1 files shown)


anashen:
Thanks @YoukaiFromAccounting! Recent change switches to using `ggrastr::rasterise` for rasterization instead of `geom_point_rast`, as this is recommended by ggrastr documentation.
- - -

Two things to note:
- (in general, unrelated to changes) (for DimPlot) coloring is not preserved between ordered and unordered plots **unless the `cols` argument is a named vector that maps group->color**
- as noted above, ggrastr is slower than geom_scattermore, but the behavior also differs slightly as the aesthetics are more similar to ggplot2. in the case of `rasterise`, since it wraps around `geom_point`, we use the same aesthetics. these can be adjusted with the parameters of `DimPlot`/`FeaturePlot`.

Issue #10314, opened 2026-02-27
Calls to Rf_error


Enchufa2:
We've detected calls to `Rf_error` in your C++ code, which may be problematic, see RcppCore/Rcpp/issues/1247. Rcpp will warn about these calls in the near future, and will error later on. The purpose is to force package developers to think about their uses of `Rf_error` to:

- substitute them with `Rcpp::stop` in most cases, which does proper unwinding, calling destructors in the process to avoid memory leaks; or
- protect them with parenthesis as in `(Rf_error)(...)` if these uses are justified (e.g. in callbacks that are passed on to R functions).

Please review your `Rf_error` calls and adjust them accordingly.

anashen:
Thank you @Enchufa2 for the heads up - will fix!

PR #10353, opened 2026-03-28
Remove redundant exception handling in RModularityOptimizer

Enchufa2:
As a follow-up for #10314, Rcpp already try/catches exceptions for you, so this old construct is redundant.


anashen (approved):
Thanks @Enchufa2!

PR #10398, opened 2026-05-22
Fixed feature subsetting in PrepDR5

### LINKED ISSUE #8627: Seurat v5: RunPCA discards supplied features if there are other non-supplied features with zero variance in scale.data matrix
When RunPCA is used on a Seurat v5 object, it calls Seurat:::PrepDR5 internally. Due to a bug in PrepDR5, supplied features (i.e. RunPCA(seurat_object, features = supplied_features)) are discarded if there are non-supplied features with zero variance.

The bug occurs on lines 2435 and 2436 of dimensional_reduction.R. Line 2435 should be changed from

`features.var <- apply(X = data.use, MARGIN = 1L, FUN = var)`

to:

`features.var <- apply(X = data.use[features,], MARGIN = 1L, FUN = var)`

So that in line 2436 the supplied features are subsetted correctly. Currently, they are subsetted by a logical which is longer than the length of the supplied features and isn't in the same order as them:

`features.keep <- features[features.var > 0]`

Reproducible example on simulated data:

```r
count_mat <- do.call(rbind, lapply(rep(1000,1000), rnorm)) + 100
rownames(count_mat) <- paste0("gene", 1:1000)
count_mat[1:100,] <- 0
seurat_object <- CreateSeuratObject(count_mat)
seurat_object <- NormalizeData(seurat_object)
#does not matter if scaled to reproduce bug
seurat_object <- ScaleData(seurat_object, do.scale = F)
#note that zero-variance features were not supplied here
seurat_object <- RunPCA(seurat_object, features = rownames(count_mat)[101:1000])
#should be 900 but is only 800
dim(seurat_object@reductions$pca@feature.loadings)

#bug does not occur if only features with non-zero variance given to Seurat
seurat_object <- CreateSeuratObject(count_mat[101:1000,])
seurat_object <- NormalizeData(seurat_object)
#does not matter if scaled to reproduce bug
seurat_object <- ScaleData(seurat_object, do.scale = F)
#note that zero-variance features were not supplied here
seurat_object <- RunPCA(seurat_object, features = rownames(count_mat)[101:1000])
#should be 900 and is 900
dim(seurat_object@reductions$pca@feature.loadings)
```

igrabski:
To fix GitHub issue #8627 -- previously, `RunPCA` incorrectly discards features if any zero-variance features are present, due to improper subsetting. This is now fixed through a simple modification to `PrepDR5`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.5.0.9003
+Version: 5.5.0.9004
 Title: Tools for Single Cell Genomics
 Description: A toolkit for quality control, analysis, and exploration of single cell RNA sequencing data. 'Seurat' aims to enable users to identify and interpret sources of heterogeneity from single cell transcriptomic measurements, and to integrate diverse types of single cell data. See Satija R, Farrell J, Gennert D, et al (2015) <doi:10.1038/nbt.3192>, Macosko E, Basu A, Satija R, et al (2015) <doi:10.1016/j.cell.2015.05.002>, Stuart T, Butler A, et al (2019) <doi:10.1016/j.cell.2019.05.031>, and Hao, Hao, et al (2020) <doi:10.1101/2020.10.12.335331> for more details.
 Authors@R: c(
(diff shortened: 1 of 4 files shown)


anashen (approved):
Thanks @igrabski! I added two test cases to prevent future regressions.

PR #10429, opened 2026-07-23
Fix FindConservedMarkers meta method labels

### LINKED ISSUE #9100: meta.method name (not itself) not actually used in FindConservedMarkers function
```r
When I'm trying different non-default metap package methods for FindConservedMarkers function, I found it never reports any other p-values other than the default mimimum. I checked the source code and looks like the meta.method is not correctly used for getting its names. The current function tries to grep its function name by: 
meta.method.name <- as.character(x = formals()$meta.method)
However, after tested, this method always get the DEFAULT method name, namely metap::minimump, therefore, FindConservedMarkers always uses the "mimimump" to generate the output. I think the correct way of getting the user-provided function name should be:
meta.method.name <- as.character(substitute(meta.method))
Hoe this bug can be verified and fixed soon!
```

xHaMMaDy:
## Summary

- name the combined p-value column after the supplied `metap` method instead of the default formal argument
- resolve exported `metap` functions by identity so stored function references receive the correct label
- add regression coverage for direct and stored non-default methods while preserving the default label

## Tests

- focused `FindConservedMarkers` regression test with `pbmc_small`
- parsed the modified source and test files successfully

Closes #9100

### DIFF
--- a/R/differential_expression.R
+++ b/R/differential_expression.R
@@ -420,9 +420,24 @@ FindConservedMarkers <- function(
         return(meta.method(clean_x)$p)
       }
     ))
-    meta.method.name <- as.character(x = formals()$meta.method)
-    if (length(x = meta.method.name) == 3) {
-      meta.method.name <- meta.method.name[3]
+    meta.method.name <- getNamespaceExports(ns = "metap")
+    meta.method.name <- meta.method.name[vapply(
+      X = meta.method.name,
+      FUN = function(method) {
+        identical(
+          x = meta.method,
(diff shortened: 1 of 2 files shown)


xHaMMaDy:
Hi! Just checking whether this approach to resolving the supplied meta-analysis method name looks appropriate. I'm happy to simplify or adjust it if preferred.

cyrillustan:
Hi @xHaMMaDy, yes this approach looks good! We'll test it and get it merged

anashen (approved):
Thank you @xHaMMaDy for the fix, and appreciate the added test!

PR #10451, opened 2026-08-27
Fix GetResidual() for multi-model SCT assays 

### LINKED ISSUE #10441: GetResidual() fails / silently drops features for multi-model SCTAssay when features are not modeled in every model
## Bug description

`GetResidual()` on an SCTAssay containing **multiple SCT models** (e.g. per-sample `SCTransform` results merged into one object) cannot compute residuals for features that are not modeled in *every* model:

- with the default `na.rm = TRUE`, the feature is silently dropped (only a `Residuals not computed for the following requested features: ...` warning is issued) — there is no way to obtain residuals for it even though it is modeled in some samples;
- with `na.rm = FALSE`, the call errors out. On a real 12-model object (per-sample SCTransform, joined layers) it fails with `Error: object 'new_residual' not found`; on the minimal 2-model repro below it fails with `Error: number of rows of matrices must match (see arg 2)` (in some layer states also `subscript out of bounds`).

## Steps to reproduce

```r
library(Seurat)
data("pbmc_small", package = "SeuratObject")
set.seed(1)
obj1 <- SCTransform(pbmc_small[, 1:40], verbose = FALSE)
obj2 <- SCTransform(pbmc_small[, 41:80], verbose = FALSE)
merged <- merge(obj1, obj2)

# a gene modeled in obj1 only, and one modeled in both objects
only1 <- setdiff(VariableFeatures(obj1), VariableFeatures(obj2))[1]  # e.g. "TCL1A"
both  <- intersect(VariableFeatures(obj1), VariableFeatures(obj2))[1]

# na.rm = FALSE: crashes
GetResidual(merged, features = c(only1, both), assay = "SCT", na.rm = FALSE)
# Error: number of rows of matrices must match (see arg 2)
```

With the default `na.rm = TRUE`, the same call completes but `only1` is absent from `scale.data` afterwards (warning only).

## Root cause

In `FetchResidualSCTModel`, the `diff_features` / `intersect_features` handling is incomplete: when a model contains only *some* of the requested features (`diff_features > 0` while `intersect_features > 0`), the `else` branch issues a warning but then falls through to `return(new_residual)`, where `new_residual` is only ever defined inside the `diff_features == 0` branch — hence `object 'new_residual' not found`. `GetResidualSCTModel` has the same structure. When `intersect_features == 0`, an NA matrix over `model.cells` is returned instead; with different models returning different row counts, the subsequent `Reduce(cbind, ...)` in `GetResidual` fails with the matrix-mismatch error.

Related: `GetResidual` never passes `layer.cells` to the per-model helpers, so every model computes residuals over all cells of the UMI assay rather than its own `model.cells`. The per-model matrices therefore overlap, and the `new.scale[rownames, colnames] <- ...` merge can either go out of bounds or silently end up last-model-wins — i.e. cells receive residuals computed with another sample's model parameters. `PrepSCTFindMarkers` avoids this by restricting each model to its own cells (`cells <- rownames(cell_attr[[model_name]])`).

## Expected behavior

For a multi-model SCTAssay, residuals should be computed per model, each restricted to its own cells and its own modeled features, returning NA for cells whose model does not include the feature — the pattern already used in `PrepSCTFindMarkers`.

## Environment

- R 4.4.3
- Seurat 5.5.1
- SeuratObject 5.4.0
- sctransform 0.4.3

YoukaiFromAccounting:
Fixes #10441 

This PR fixes GetResidual() behaviour for SCT assays that contain multiple SCT models that resulted from separate runs of SCTransform(). 

Previously, GetResidual() behaved incorrectly when requested features were not modeled in every SCT model in a Seurat object; when its parameter na.rm was set to FALSE, this should normally remove all genes with NAs from the residual matrix. However, without these changes, this would instead result in a crash with the following error: `# Error: number of rows of matrices must match (see arg 2)`. 

Reproducible example (thank you @pwwang!): 
```
library(Seurat)
data("pbmc_small", package = "SeuratObject")
set.seed(1)
obj1 <- SCTransform(pbmc_small[, 1:40], verbose = FALSE)
obj2 <- SCTransform(pbmc_small[, 41:80], verbose = FALSE)
merged <- merge(obj1, obj2)

# a gene modeled in obj1 only, and one modeled in both objects
only1 <- setdiff(VariableFeatures(obj1), VariableFeatures(obj2))[1]  # e.g. "TCL1A"
both  <- intersect(VariableFeatures(obj1), VariableFeatures(obj2))[1]

# na.rm = FALSE: crashes
GetResidual(merged, features = c(only1, both), assay = "SCT", na.rm = FALSE)
# Error: number of rows of matrices must match (see arg 2)
```

This PR introduces some changes to GetResidual(), GetResidualSCTModel(), and FetchResidualSCTModel() that addresses these issues. 

GetResidual() is normally supposed to loop over each SCT model in an object and call the GetResidual/FetchResidual helpers to recompute any missing residuals. However the previous code just passed the object's entire set of cells, making the output structurally inconsistent. GetResiduals() now passes the current model's cells into the residual calculator helpers, such that each SCT model correctly computes residuals for its own cells.

GetResidualSCTModel() and FetchResidualSCTModel() now explicitly handle partial feature overlap better by only computing residuals for features present in a model, and padding any missing features with NAs. 

With these changes, the residual matrix for the above reprex now generates correctly:
```
#Print residuals for 4 cells, of which 2 cells lack the TCL1A feature in their SCT model
> residual_mat[c(only1, both), 39:42]
      GGCATATGCTTATC CATTACACCAACTG TAGGGACTGAACTC GCTCCATGAGAAGT
TCL1A     -0.2626143      -0.325272             NA             NA
GNLY       1.3417045       1.341705       1.302976     -0.1981292
```

Simple example of the problem: 
Suppose GetResidual is called for features **A** and **B** across an SCT assay with two SCT models. model1 contains both A and B, but model2 contains only B. model1 contains cells c1, c2, while model2 contains cells c3, c4. Under healthy GetResidual na.rm = FALSE behaviour, we would expect the residual matrix to look something like:

```text
      c1   c2   c3   c4
A    0.2  0.1   NA   NA
B    0.1  0.2  0.3  0.1
```
Without these changes, running Get/FetchResidualSCTModel would return

```text
      c1   c2
A    0.2  0.1
B    0.1 0.2
```

for model1 and

```text
      c3   c4
B    0.3 0.1
```
for model two, with inconsistent row sets that cause GetResidual to error. After this change, each model computes residuals only for the features it contains, however the helper functions now also pad missing features with NA:

```text
      c1   c2
A    0.2  0.1
B    0.1  0.2
```
for model1 and 

```text
      c3  c4
A    NA  NA
B    0.3  0.1
```
for model 2. This would eventually combine into the full residual matrix that can now be returned by GetResidual():

```text
      c1   c2   c3   c4
A    0.2  0.1   NA   NA
B    0.1  0.2  0.3 0.1
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.5.1.9003
+Version: 5.5.1.9004
 Title: Tools for Single Cell Genomics
 Description: A toolkit for quality control, analysis, and exploration of single cell RNA sequencing data. 'Seurat' aims to enable users to identify and interpret sources of heterogeneity from single cell transcriptomic measurements, and to integrate diverse types of single cell data. See Satija R, Farrell J, Gennert D, et al (2015) <doi:10.1038/nbt.3192>, Macosko E, Basu A, Satija R, et al (2015) <doi:10.1016/j.cell.2015.05.002>, Stuart T, Butler A, et al (2019) <doi:10.1016/j.cell.2019.05.031>, and Hao, Hao, et al (2020) <doi:10.1101/2020.10.12.335331> for more details.
 Authors@R: c(
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,8 @@
 
 ### Fixes
 
-- Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
+- Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix ([#10448](https://github.com/satijalab/seurat/pull/10448))
+- Fixed `GetResidual()` to correctly handle multi-model SCT assays with partial feature overlap ([#10541](https://github.com/satijalab/seurat/pull/10451))
 - Fixed bug in `PercentageFeatureSet` where layer data was incorrectly retrieved prior to finding features with requested pattern ([#10438](https://github.com/satijalab/seurat/pull/10438))
 - Updated `as.SingleCellExperiment` to address conversion case where an object has both original and sketched assay / reductions (differing numbers of cells) ([#10419](https://github.com/satijalab/seurat/pull/10419))
 - Fixed bugs in behavior of `RidgePlot` parameters `fill.by`, `y.max`, and `same.y.lims` ([#10424](https://github.com/satijalab/seurat/pull/10424))
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -445,8 +445,10 @@ GetResidual <- function(
     new.residuals <- lapply(
       X = sct.models,
       FUN = function(x) {
+        model.cells <- Cells(x = slot(object = object[[assay]], name = "SCTModel.list")[[x]])
         FetchResidualSCTModel(object = object[[assay]],
                               umi.object = object[[umi.assay]],
+                              layer.cells = model.cells,
                               SCTModel = x,
                               new_features = features,
                               replace.value = replace.value,
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -5972,9 +5972,8 @@ GetResidualSCTModel <- function(
   }
   diff_features <- setdiff(x = features_to_compute, y = model.features)
   intersect_features <- intersect(x = features_to_compute, y = model.features)
-  features_to_compute_model <- features_to_compute
   if (length(x = diff_features) == 0) {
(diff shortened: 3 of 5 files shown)


anashen (commented):
Thanks @YoukaiFromAccounting, can you take a look at the new commits and let me know what you think?

bb9d0e1e2d9321425575529d74a454ba42646149 removes the redundant `features_to_compute_model` state - when only some requested features are modeled for a given SCT model, we can compute residuals for `intersect_features` only, then pad the result back to the full `features_to_compute` shape by assigning with `rownames(new_residual)` / `colnames(new_residual)`

a13d123922dac2aef86efbb34101b917de16ab2b computes `diff_features` after `vst_out` is built

4ff2b79e39003b5c6d29aeb9c9af3ba38a350940 adds a test verifying the fix

YoukaiFromAccounting:
@anashen Thank you for the catches; those commits look great, I tested that the new diff_features logic now correctly handles both missing and zero-overlap feature cases. The new testing function also looks good. All good on my end.

PR #10017, opened 2025-07-25
feat: Ultra-large matrix support and Read10X performance optimization

BenjaminDEMAILLE:
This PR introduces comprehensive support for ultra-large matrices and significantly optimizes Read10X performance.

Key Features:
- Ultra-large matrix support with spam integration
- 2-10x Read10X performance improvement via data.table
- Batch processing automation with CreateMultipleSeurat
- SoupX integration for doublet decontamination
- Full backward compatibility maintained

Technical improvements:
- Integrated data.table::fread for faster I/O
- Vectorized file processing and memory optimization  
- Enhanced support for matrices exceeding 2^31 elements
- Smart auto-detection for workflow automation

All changes maintain full API compatibility.

Issue #9798, opened 2025-04-02
Enable Seurat to Handle Ultra-Large Matrices (>2^31 values)


BenjaminDEMAILLE:
# Motivation

As the scale of single-cell and spatial transcriptomics datasets continues to grow, particularly in the context of large national or international consortia (e.g., Human Cell Atlas, BICCN, HuBMAP, Allen Brain Atlas), we are increasingly working with datasets that exceed the current technical limitations of R's internal matrix representations.

Seurat currently relies on in-memory matrix formats such as `matrix` or `dgCMatrix` (from the **Matrix** package), which use **32-bit signed integers** to represent indices and pointers. This imposes a hard limit of **2^31 - 1 elements (~2.1 billion entries)** per matrix. Once a matrix exceeds this number of values, it becomes impossible to represent or manipulate it within these formats.

##Example: Integration of 10 large scRNA-seq samples

Suppose we are analyzing 10 human scRNA-seq samples, each with:

- ~300,000 cells  
- ~30,000 genes  

This results in a combined gene expression matrix of size:

300,000 cells/sample × 10 samples = 3,000,000 cells
3,000,000 cells × 30,000 genes = 90,000,000,000 values

Even if the matrix is sparse and only 1% of the values are non-zero, that still gives **900 million entries** — already nearing the 2^31 limit. With more samples, higher cell counts, or multi-modal assays (e.g., CITE-seq or spatial transcriptomics), we can easily cross this threshold.

When trying to create a `dgCMatrix` or pass this matrix into Seurat with `CreateSeuratObject()`, users may receive errors such as:

`Error in validObject(.Object) : 
  invalid class “dgCMatrix” object: 'i' slot is too large (> 2^31 - 1 elements)` 

## Real-World Consequences

- Users are forced to **downsample** their data or split it into batches, losing valuable biological information.
- Large institutions with powerful compute infrastructures are **unable to take advantage** of their resources within the Seurat framework due to this internal bottleneck.
- Other tools (e.g., Scanpy in Python) are more flexible in this regard, making users consider migrating away from Seurat despite its rich functionality.

## Why This Feature Matters

Projects like the **Allen Brain Atlas** and **BICCN (Brain Initiative Cell Census Network)** regularly release **ultra-large single-cell and spatial transcriptomics datasets**, often exceeding millions of cells and covering diverse brain regions and modalities. These datasets are invaluable for neuroscience research but are difficult to analyze at full scale using Seurat under current constraints.

[...]

PR #10452, opened 2026-08-30
Make Seurat objects with on-disk (BPCells) layers portable

BenjaminDEMAILLE:
## Motivation

Addresses the second half of #9798: an object with BPCells layers cannot be shared.

A Seurat object with on-disk layers records only the **absolute path** of each BPCells matrix directory. The `.Rds` is therefore not self-contained. Sending it to a collaborator, uploading it, or simply moving or renaming the matrix directory leaves the layers unreadable, and the failure is quiet: `LoadSeuratRds()` warns "Cannot find any of the layer files specified" and hands back an object with **no layers at all**.

This is the gap between a Seurat object and an AnnData `.h5ad`, which people can just send.

## What this adds

Two routes, depending on whether the data fits in memory.

**`AsInMemory(object)`** converts every on-disk layer (`IterableMatrix`, `DelayedMatrix`) to an in-memory `dgCMatrix`. The object then depends on nothing outside itself, so a plain `saveRDS()` writes one portable file. When a layer is too large to materialize, the error points at `SaveSeurat()` rather than surfacing the raw `2^31` `dgCMatrix` failure.

```r
obj <- AsInMemory(obj)
saveRDS(obj, "object.rds")   # single, self-contained file
```

**`SaveSeurat()` / `LoadSeurat()`** write the object together with copies of its on-disk stores into one gzip-compressed archive plus a small JSON manifest. `LoadSeurat()` extracts it and reconnects each layer from the extracted store, so the archive stands alone regardless of where it was created and keeps the data on disk.

```r
SaveSeurat(obj, "atlas.seurat")
obj <- LoadSeurat("atlas.seurat")   # works after any move, on any machine
```

A fully in-memory object is simply serialized inside the archive, so the same two calls work for every object.

## Notes

- No new hard dependencies. `jsonlite` is already in `Imports`; `HDF5Array` is added to `Suggests` and is only touched for `DelayedMatrix` layers. All BPCells tests are guarded with `skip_if_not_installed`.
- Complementary, root-cause fix proposed separately in satijalab/seurat-object#297: `LoadSeuratRds()` resolves its cached layer paths against the working directory rather than the directory holding the `.Rds`, which is why both `relative = TRUE` and `relative = FALSE` break on a move. That fix makes existing saved objects portable; this PR gives a single file to hand over.
- Kept separate from #10017 and from the 64-bit matrix backend PR.

## Tests

New `tests/testthat/test_portable_objects.R` covers the failure itself (a plain rds of a BPCells-backed object becomes unreadable once its store is removed) and both remedies, including a bundle that is renamed while the original store is deleted. 12 passing locally with BPCells installed.


BenjaminDEMAILLE:
cc @anashen @igrabski

Companion to #10453, and independent of it: either can be reviewed on its own.

This one is the practical complaint from #9798. An object with BPCells layers stores only the absolute path of each on-disk matrix directory, so the `.Rds` is not self-contained: send it to a collaborator, or just move the store, and `LoadSeuratRds()` warns once and hands back an object with **no layers**. That is the gap between a Seurat object and an `.h5ad` someone can simply email.

Two additions, no new hard dependencies: `AsInMemory()` when the data fits in RAM, and `SaveSeurat()`/`LoadSeurat()` for a single self-contained archive when it does not. The test file reproduces the breakage first, then both remedies. `R CMD check` is clean (Status: OK) locally.

There is also a root-cause fix on the SeuratObject side in satijalab/seurat-object#297, which makes objects saved by the existing `SaveSeuratRds()` survive a move. The two are complementary and can land in either order.

---

BenjaminDEMAILLE:
Superseded by #10493, which carries both halves of the same problem: holding a matrix past the dgCMatrix limit, and being able to move the object that holds it. The commits are unchanged; closing here to keep the queue down rather than because anything was wrong with it.

PR #10405, opened 2026-06-08
Ultra-large DelayedMatrix backend and portable single-file Seurat objects (#9798)

BenjaminDEMAILLE:
## Motivation

Closes the actionable part of #9798: enabling Seurat to work with matrices larger than the `dgCMatrix` 2³¹ limit, and making such objects portable as a single file — "a Seurat object in one file, the way you don't fuss over an AnnData `.h5ad`."

> **Companion PR:** [satijalab/seurat-object#295](https://github.com/satijalab/seurat-object/pull/295) adds `StitchMatrix.DelayedMatrix` to SeuratObject (alongside `StitchMatrix.IterableMatrix`). This PR requires `SeuratObject (>= 5.4.0.9002)` and will go red on CI until that ships.

## What's in this PR

**1. `DelayedMatrix` (DelayedArray/HDF5Array) layer backend** mirroring the BPCells `IterableMatrix` backend (`as.DelayedMatrix`, `as.sparse`, `.CalcN`, `LogNormalize`, `VST`). `DelayedMatrix` carries dimnames and 64-bit indexing, so it works as a Seurat layer past 2³¹.

- **In RAM, 64-bit:** for sparse input, `as.DelayedMatrix(x)` uses a `SparseArray::SVT_SparseMatrix` seed (per-column tree, no flat 2³¹ index vector), so a layer can hold >2³¹ non-zeros **in memory** and a plain `saveRDS` writes one self-contained file — an in-memory AnnData equivalent.
- **On disk, 64-bit:** `as.DelayedMatrix(x, ondisk = TRUE)` realizes to a 10x-style HDF5 `TENxMatrix` (64-bit `indptr`).

**2. "Total" pipeline coverage** on `DelayedMatrix`-backed objects, verified numerically equal to `dgCMatrix`:
`NormalizeData` (LogNormalize native; **CLR/RC** via materialize-when-fits), `FindVariableFeatures` (**vst** native; **mvp/dispersion** materialize-when-fits), `ScaleData`, `RunPCA`, neighbors/clusters/UMAP, `SCTransform`, `FindMarkers`/`FindAllMarkers`/`FoldChange`, **`AverageExpression`/`AggregateExpression`/`PseudobulkExpression`** (block-wise), and multi-layer `split` → `JoinLayers` → `ScaleData` → `RunPCA` → `IntegrateLayers` (**CCA, RPCA, Harmony**). Non-block-native steps materialize-when-fits with a warning (never silent subsampling), matching BPCells.

[...]

PR #10453, opened 2026-08-30
Add a 64-bit DelayedMatrix backend for matrices past the dgCMatrix 2^31 limit

BenjaminDEMAILLE:
## Motivation

Addresses the first half of #9798 (currently closed, happy for it to be reopened): `dgCMatrix` indexes its non-zeros with 32-bit signed integers, so a matrix stops being representable at `2^31 - 1` entries and users hit

```
invalid class "dgCMatrix" object: 'i' slot is too large (> 2^31 - 1 elements)
```

Real datasets already cross that line: the 10x GEM-X Flex 1M human K562 CRISPR aggregate has 4,355,024,689 non-zeros. Today BPCells is the only way around it.

This PR adds a second backend built on `DelayedArray`/`HDF5Array`, mirroring the existing `IterableMatrix` one, and nothing else. It is deliberately kept separate from #10017 and from the portability PR #10452.

## What this adds

**`as.DelayedMatrix(x, ondisk = FALSE, file = NULL)`** wraps a matrix for use as a layer:

- In memory: sparse input gets an `SVT_SparseMatrix` seed (`SparseArray`), which stores non-zeros in a per-column tree rather than one flat index vector, so a layer can hold more than `2^31` non-zeros in RAM and still serialize to a single `saveRDS` file.
- On disk (`ondisk = TRUE`): realized to a 10x-style HDF5 `TENxMatrix`. Verified that the written `indptr` dataset has an 8-byte native HDF5 integer type, so the same limit is lifted while the data stays out of memory.

**Backend methods** so the standard pipeline stays block-wise and never materializes the matrix: `LogNormalize.DelayedMatrix`, `VST.DelayedMatrix`, `.CalcN.DelayedMatrix`, `as.sparse.DelayedMatrix`, plus `DelayedMatrix` branches in `RowMeanSparse`/`RowSumSparse`/`RowVarSparse`.

**A materialization fallback that cannot be silenced.** Operations with no block-wise form (CLR and RC normalization, `mvp`/`dispersion` feature selection) load the layer into memory. Because someone reaches for this backend precisely when the data may not fit in RAM, that always emits an immediate `warning()` naming the operation, the matrix dimensions and the block-wise alternative. It is deliberately not gated on `verbose`.

**Two pre-existing rough edges fixed along the way**, both of which also affect BPCells today:

- `PseudobulkExpression` multiplies block-wise instead of erroring with "does not support DelayedArray objects".
- `as.Seurat()` on a `SingleCellExperiment` with on-disk assays failed with an opaque `assignment of an object of class "RenameDims" is not valid for slot 'data'`. A v3 `Assay` keeps its matrices in slots typed for in-memory classes, so it now converts explicitly at that boundary with the warning above.

**`CheckMatrixSize()`** reports how close a matrix is to `.Machine$integer.max`; **`as.sparse.spam()`** coerces `spam` matrices, which carry 64-bit indices but no dimnames and so cannot back a layer directly.

## Verification

**At the actual limit.** Built an `SVT_SparseMatrix` of **2,250,000,000 non-zeros** (20,000 x 450,000, 18 distinct generated chunks, 36.6 GB resident):

| step | result |
| --- | --- |
| control: `as(big, "dgCMatrix")` | errors, `too many nonzero values (2250000000) to "fit" in a CsparseMatrix derivative` |
| `CreateSeuratObject` | 19.4 s; `nCount_RNA`/`nFeature_RNA` exact against a directly materialized slice |
| `NormalizeData` | 19.4 s; `LogNormalize` values exact to 8 significant digits against a hand computation |
| `FindVariableFeatures` (`VST.DelayedMatrix`) | 96.4 s |

**Coverage.** 37 public Seurat entry points were run twice, once on a `DelayedMatrix`-backed object and once on an identical `dgCMatrix` one, to separate backend gaps from inherent limits: **no function works on `dgCMatrix` and fails on `DelayedMatrix`**. `ScaleData`, `RunPCA`, neighbours/clusters, UMAP/t-SNE, `FindMarkers`/`FindAllMarkers`/`FoldChange`, `AverageExpression`/`AggregateExpression`/`PseudobulkExpression`, `merge`, `subset`, `JoinLayers` and the plotting functions all run.

**Ecosystem, against `main` as the control.** UCell, SingleCellExperiment, scater, scran, scDblFinder, harmony, glmGamPoi and slingshot: **no regression on the `dgCMatrix` path**, and no name collision for the new exports across 16 packages. One third-party limit worth stating: `UCell::ScoreSignatures_UCell()` rejects a bare `DelayedMatrix` ("Unrecognized input format"), since its dispatch predates these classes; the normal path, `UCell::AddModuleScore_UCell()` on the Seurat object, works.

## Tests and checks

`tests/testthat/test_large_matrix.R` (44 assertions, all guarded with `skip_if_not_installed`): backend equivalence to `dgCMatrix`, in-RAM SVT seed and single-rds round-trip, 64-bit on-disk `TENxMatrix`, PCA, the materialization warning firing under `verbose = FALSE` and staying quiet on block-native paths, `as.Seurat` interop, split layers through `JoinLayers` + CCA/RPCA integration, and `spam` coercion.

Full suite: 571 passing. The 2 failures (`test_load_10X.R`, `Read10X_Image`) reproduce unchanged on `origin/main` and are unrelated. `R CMD check --as-cran`-style run: Status OK.

## Notes

- No new hard dependencies: `DelayedMatrixStats`, `HDF5Array`, `SparseArray`, `spam` are added to `Suggests` (`DelayedArray` was already there).
- Multi-layer (split) `DelayedMatrix` assays additionally need `StitchMatrix.DelayedMatrix`, proposed separately in satijalab/seurat-object#295. This PR does **not** bump the `SeuratObject` requirement, so it is green on its own; that test skips when the method is unavailable.


BenjaminDEMAILLE:
cc @anashen @igrabski

Flagging this for visibility since #9798 was closed. @igrabski, you said on that issue that you would be happy to review a PR, so here it is, split into the small pieces that were missing last time.

This one is the 64-bit half only: a `DelayedMatrix` layer backend so a counts matrix can exceed the `dgCMatrix` 2^31 non-zero limit. No new hard dependencies, everything new is behind `Suggests` and `skip_if_not_installed`, and `R CMD check` is clean (Status: OK) locally.

The portability half is #10452 and is independent of this one; either can be reviewed first. This supersedes #10405, which I closed.

---

BenjaminDEMAILLE:
Superseded by #10493, which carries both halves of the same problem: holding a matrix past the dgCMatrix limit, and being able to move the object that holds it. The commits are unchanged; closing here to keep the queue down rather than because anything was wrong with it.

Issue #10099, opened 2025-09-19
BUG: Segfault in FindSpatiallyVariable Features


swbioinf:
### Issue Description

Hi,

With the latest Seurat (5.3.0), I'm getting a segfault from FindSpatiallyVariableFeatures().

I think it may be to do with the fact if I run GetTissueCoordinates(), the cell identifier is in a column called 'cell', but not rownames. But FindSpatiallyVariableFeatures() gets a list of features from the rownames, and tries to calculate RowVar on a matrix with no columns. [Relevant code here](https://github.com/satijalab/seurat/blob/30f82df52159ac5f0feb80b149698abbd876b779/R/preprocessing.R#L4468)

```
> GetTissueCoordinates(so.sample)
         x         y          cell
1   166300 -64502.29   XXXX_1_21
2   166666 -64486.29   XXXX_2_21
3   168091 -64498.29   XXXX_3_21
```
I can see theres a bunch of GetTissueCoordinates() functions for different techs, but can't figure out what is run for cosmx data, which I am using. I wonder if its particular to that?

### Reproducing Code Example

```R
so.sample <- FindSpatiallyVariableFeatures(
  so.sample,
  assay = "RNA",
  features = VariableFeatures(so.sample),
  selection.method = "moransi",
  layer = "scale.data")
```

### Error Message

```shell
*** caught segfault ***
address (nil), cause 'memory not mapped'

Traceback:
 1: RowVar(x = data)
 2: FindSpatiallyVariableFeatures.StdAssay(object = object[[assay]],     slot = slot, features = features, spatial.location = tc,     selection.method = selection.method, r.metric = r.metric,     x.cuts = x.cuts, y.cuts = y.cuts, nfeatures = nfeatures,     verbose = verbose, ...)
 3: FindSpatiallyVariableFeatures(object = object[[assay]], slot = slot,     features = features, spatial.location = tc, selection.method = selection.method,     r.metric = r.metric, x.cuts = x.cuts, y.cuts = y.cuts, nfeatures = nfeatures,     verbose = verbose, ...)
 4: FindSpatiallyVariableFeatures.Seurat(so.sample, assay = "RNA",     features = VariableFeatures(so.sample), selection.method = "moransi",     layer = "scale.data")
 5: FindSpatiallyVariableFeatures(so.sample, assay = "RNA", features = VariableFeatures(so.sample),     selection.method = "moransi", layer = "scale.data")
An irrecoverable exception occurred. R is aborting now ...
[...]

Issue #9092, opened 2024-07-08
Error in FindSpatiallyVariableFeatures.default(object = data, spatial.location = spatial.location,  :    Please provide the same number of observations as spatial locations.


ankushs0128:
I am encountering an error when trying to find spatially variable features using the FindSpatiallyVariableFeatures function in Seurat. The specific error message which is also raised in closed issue [Issue 4211](https://github.com/satijalab/seurat/issues/4211)

```
Error in FindSpatiallyVariableFeatures.default(object = data, spatial.location = spatial.location,  : 
  Please provide the same number of observations as spatial locations.
```

I have merged several Seurat objects into a single Seurat object (combined_seurat).
I preprocessed the combined Seurat object using PrepSCTFindMarkers and SCTransform.
I performed PCA, found neighbors, and identified clusters.
I attempted to run FindSpatiallyVariableFeatures on the combined Seurat object.
Here is the relevant part of my code:
```
combined_seurat <- PrepSCTFindMarkers(combined_seurat)
combined_seurat <- SCTransform(combined_seurat, assay = "Spatial", verbose = TRUE)
combined_seurat <- RunPCA(combined_seurat, verbose = FALSE)
combined_seurat <- FindNeighbors(combined_seurat, dims = 1:30)
combined_seurat <- FindClusters(combined_seurat, resolution = 0.8, verbose = FALSE)
combined_seurat <- RunUMAP(combined_seurat, dims = 1:30, n.neighbors = 30, min.dist = 0.3)

# Save results
cluster.markers <- FindAllMarkers(combined_seurat, only.pos = FALSE, min.pct = 0.25)
write.csv(cluster.markers, file.path(output_dir, "Clustermarkers.csv"), row.names = TRUE)

# Attempt to find spatially variable features
combined_seurat <- FindSpatiallyVariableFeatures(combined_seurat, 
                                                 features = VariableFeatures(combined_seurat)[1:1000], 
                                                 slot = "scale.data", 
                                                 selection.method = "markvariogram")
```

The function throws an error message indicating a mismatch between the number of observations and spatial locations.

Session info
> Seurat version: [Seurat_5.1.0.9002  SeuratObject_5.0.2]
> R version 4.3.3 (2024-02-29)
> Platform: aarch64-apple-darwin20 (64-bit)
> Running under: macOS Sonoma 14.3
> 

Any pointers resolving the error would be highly appreciated.  

Thanks

LinbeiGithub:
@May be

[...]

Issue #7020, opened 2023-03-06
error in FindSpatiallyVariableFeatures


BenjaminDEMAILLE:
Hi all !

I try to run FindSpatiallyVariableFeatures on a multi slide Seurat object. but I get this

Error in switch(EXPR = selection.method, markvariogram = RunMarkVario(spatial.location = spatial.location, : 
EXPR must be a length 1 vector
I changed the lowres image to the hires and ran SCT on each samples

thanks

BenjaminDEMAILLE:
Any ideas ?

reliscu:
Getting this error as well.

BenjaminDEMAILLE:
Cause: `selection.method` defaults to `c('markvariogram', 'moransi')` and is never resolved to one value, so omitting it passes the whole vector on. In `FindSpatiallyVariableFeatures.default` it reaches `switch()`, and in the assay method it reaches `if (selection.method == "markvariogram")`, which under R 4.2 and later errors with *the condition has length > 1*.

Fix in https://github.com/satijalab/seurat/pull/10457: `match.arg()` in the default, assay and Seurat methods. It also means an unknown method is rejected before the variogram is computed rather than after.

That PR also carries the fix for the crash on FOV-based images (Xenium, Vizgen, VisiumV2), which is what you hit first if you supply the method explicitly.

---
*Outside contribution: I am not a member of the Seurat dev team, and this comment is not an official answer from the maintainers.*

PR #10457, opened 2026-08-30
Fix FindSpatiallyVariableFeatures crashing on FOV-based objects

BenjaminDEMAILLE:
Kept on its own: it is the only crash in the batch, and the fix is in C++ as well as R.

On `main`, `FindSpatiallyVariableFeatures` aborts the R session on an FOV-based object. The Eigen assertion fires in `RowVar`, and in a CRAN build without assertions it segfaults instead ([#10099](https://github.com/satijalab/seurat/issues/10099)).

| Part | Fix |
| --- | --- |
| `src/data_manipulation.cpp` | `RowVar` returns `NA_REAL` per row when there are fewer than two columns, instead of reducing over an empty range |
| `R/preprocessing.R` | Features with zero variance are dropped before `loess`, `nfeatures` is passed through rather than defaulting to the argument's own name, and the `cell` column is moved to rownames so the coordinate frame is numeric ([#7020](https://github.com/satijalab/seurat/issues/7020), [#9092](https://github.com/satijalab/seurat/issues/9092)) |

The C++ guard is the part worth reviewing carefully: the crash is reachable from any caller that hands `RowVar` a single-column matrix, not only this one.

## Verification

Test fails without the fix, by crashing the session, so it is written to run in a separate process. Full suite passes with it.


BenjaminDEMAILLE:
Added a further fix in the same function family, for #7020: `selection.method` defaults to `c('markvariogram', 'moransi')` and nothing resolved it, so omitting the argument passed a length-2 value into `switch()` in the default method, and into `if (selection.method == "markvariogram")` in the assay method, which errors with *the condition has length > 1*.

`match.arg()` in each of the three methods. It also rejects an unknown method up front instead of after the variogram has been computed.

It belongs here rather than in its own PR because the two overlap: reaching the `selection.method` failure at all requires the empty-matrix guard from this PR on any FOV-based object.

Tests added for both the omitted and the invalid case; suite passing.

---

BenjaminDEMAILLE:
One more from the same family, #9092: `FindSpatiallyVariableFeatures()` takes its coordinates from a single image,

```r
image <- image %||% DefaultImage(object = object)
```

and a merged object has one image per sample. So the analysis covers only the cells of whichever image comes first, and says nothing about it. Depending on the image class that is either a quiet partial analysis, or the failure in that issue, *Please provide the same number of observations as spatial locations*.

```
This object has 2 images ('fov.a', 'fov.b'); using 'fov.a', which covers 40 of the 80
cells in assay 'Spatial'. Pass image = to choose another, and run each image separately
to cover them all
```

A warning rather than an error: the single-image analysis is still a legitimate thing to want, it just should not be silent. An object with one image says nothing, as before.

Two more tests; suite passing.

---

BenjaminDEMAILLE:
Superseded by #10492, which now carries this fix alongside the other independent ones. The commits are unchanged, including the C++ guard in RowVar; closing here to keep the queue down rather than because anything was wrong with it.

anashen (commented):
Hi @BenjaminDEMAILLE - the core fix looks good, but these changes seem a bit larger in scope than the original issue. 

I opened PR #10504 with a narrower fix for the cell lookup path and related edge cases (plus the two minor changes with `match.arg` on `selection.method` and the wrong `nfeatures` default).

I'm closing this in favor of that PR, and I’ve added you as co-author on the relevant commits. Feel free to add a review on the other PR if you would like. Thanks!

PR #10461, opened 2026-08-31
Leave cells without usable prediction scores unlabelled

### LINKED ISSUE #10003: NaNs in prediction.scores cause FindTransferAnchors to fail
### Issue Description

Dear developers, we just stumbled upon this issue.

When running the following code and other analyses, NaNs are generated in `prediction.scores` inside `FindTransferAnchors`, causing the following error.

```
Error in possible.ids[apply(X = prediction.scores, MARGIN = 1, FUN = which.max)] : 
  invalid subscript type 'list'
```

This is because `which.max` applied to a vector with NaNs returns a list and not a vector. Adding the following processing steps solved the issue (likely avoiding the NaNs in the first place):

```
pbmc_small <- NormalizeData(pbmc_small)
pbmc_small <- FindVariableFeatures(pbmc_small)
pbmc_small <- ScaleData(pbmc_small)
pbmc_small <- RunPCA(pbmc_small)
pbmc_small <- FindNeighbors(pbmc_small, dims = 1:30)
pbmc_small <- FindClusters(pbmc_small)
```

What is the meaning of a NaN in the scores?
As we are facing the same problem on a real dataset (even after the above steps), we would like to know whether this is likely a normalization issue or how to deal with those NaNs.
Thank you for your help

### Reproducing Code Example

```R
library(Seurat)
data('pbmc_small')
# split into reference and query
reference <- pbmc_small
query     <- pbmc_small


set.seed(89)

anchors_test <- FindTransferAnchors(
  reference = reference,
  query     = query,
  dims      = 1:10,
  reference.reduction = 'pca',
)


# debug(Seurat::TransferData)

test_preds <- TransferData(
  anchorset     = anchors_test,
  refdata       = reference$groups,   # built‑in grouping factor
  query         = query,
  k.weight      = 5,
  store.weights = TRUE
)
```

### Error Message

```shell
Error in possible.ids[apply(X = prediction.scores, MARGIN = 1, FUN = which.max)] : 
  invalid subscript type 'list'
```

### Additional Comments

_No response_

### Session Info

```shell
sessionInfo()
R version 4.4.1 (2024-06-14)
Platform: x86_64-pc-linux-gnu
Running under: Ubuntu 24.10

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.12.0 
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.12.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=it_IT.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=it_IT.UTF-8   
 [6] LC_MESSAGES=en_US.UTF-8    LC_PAPER=it_IT.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=it_IT.UTF-8 LC_IDENTIFICATION=C       

time zone: Europe/Rome
tzcode source: system (glibc)

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] future_1.58.0      Seurat_5.3.0       SeuratObject_5.1.0 sp_2.2-0          

loaded via a namespace (and not attached):
  [1] deldir_2.0-4           pbapply_1.7-2          gridExtra_2.3          rlang_1.1.6            magrittr_2.0.3        
  [6] RcppAnnoy_0.0.22       matrixStats_1.5.0      ggridges_0.5.6         compiler_4.4.1         spatstat.geom_3.4-1   
 [11] png_0.1-8              vctrs_0.6.5            reshape2_1.4.4         stringr_1.5.1          crayon_1.5.3          
 [16] pkgconfig_2.0.3        fastmap_1.2.0          promises_1.3.3         purrr_1.0.4            jsonlite_2.0.0        
 [21] goftest_1.2-3          later_1.4.2            spatstat.utils_3.1-4   irlba_2.3.5.1          parallel_4.4.1        
 [26] cluster_2.1.8.1        R6_2.6.1               ica_1.0-3              stringi_1.8.7          RColorBrewer_1.1-3    
 [31] spatstat.data_3.1-6    reticulate_1.42.0      parallelly_1.45.0      spatstat.univar_3.1-3  lmtest_0.9-40         
 [36] scattermore_1.2        Rcpp_1.0.14            tensor_1.5             future.apply_1.20.0    zoo_1.8-14            
 [41] sctransform_0.4.2      httpuv_1.6.16          Matrix_1.7-3           splines_4.4.1          igraph_2.1.4          
 [46] tidyselect_1.2.1       rstudioapi_0.17.1      dichromat_2.0-0.1      abind_1.4-8            codetools_0.2-20      
 [51] spatstat.random_3.4-1  miniUI_0.1.2           spatstat.explore_3.4-3 listenv_0.9.1          lattice_0.22-7        
 [56] tibble_3.3.0           plyr_1.8.9             shiny_1.10.0           ROCR_1.0-11            Rtsne_0.17            
 [61] fastDummies_1.7.5      survival_3.8-3         polyclip_1.10-7        fitdistrplus_1.2-2     pillar_1.10.2         
 [66] KernSmooth_2.23-24     plotly_4.10.4          generics_0.1.4         RcppHNSW_0.6.0         ggplot2_3.5.2         
 [71] scales_1.4.0           globals_0.18.0         xtable_1.8-4           glue_1.8.0             lazyeval_0.2.2        
 [76] tools_4.4.1            data.table_1.17.4      RSpectra_0.16-2        RANN_2.6.2             dotCall64_1.2         
 [81] cowplot_1.1.3          grid_4.4.1             tidyr_1.3.1            rsthemes_0.5.0         nlme_3.1-168          
 [86] patchwork_1.3.0        cli_3.6.5              spatstat.sparse_3.1-0  spam_2.11-1            viridisLite_0.4.2     
 [91] dplyr_1.1.4            uwot_0.2.3             gtable_0.3.6           digest_0.6.37          progressr_0.15.1      
 [96] ggrepel_0.9.6          htmlwidgets_1.6.4      farver_2.1.2           htmltools_0.5.8.1      lifecycle_1.0.4       
[101] httr_1.4.7             mime_0.13              MASS_7.3-65
```

BenjaminDEMAILLE:
Fixes #10003.

## The bug

`TransferData()` fails when any cell's prediction scores are all `NaN`:

```
Error in possible.ids[apply(X = prediction.scores, MARGIN = 1, FUN = which.max)] :
  invalid subscript type 'list'
```

The reporter had the mechanism right, and it reproduces in three lines with no Seurat objects involved:

```r
which.max(c(NaN, NaN))               #> integer(0)
apply(scores, 1, which.max)          #> a list, once any row is all NaN
ids[.]                               #> Error: invalid subscript type 'list'
```

`apply()` returns a matrix or vector only when every result has the same length. One all-`NaN` row makes the results ragged, so it returns a list, and a list is not a valid index.

## The change

A cell with no usable score has no defensible label, so it now gets none: the index is `NA` and `predicted.id` is `NA`. The other cells are unaffected.

A warning reports how many cells that covers. Silently unlabelled cells would be its own trap, and the question in the report was what a `NaN` score means, so the message says: their anchor weights were all zero or `NaN`.

The rowwise selection moves into an internal `.MaxScoreIndex()`. That is deliberate: the failing expression was inline, and the data condition that triggers it is awkward to provoke on demand, so extracting it is what makes the fix testable rather than only arguable.

## Tests

New `tests/testthat/test_transfer_nan_scores.R`: ordinary rows give the index of their maximum, an all-`NaN` row gives `NA` rather than `integer(0)`, the result is an integer vector and not a list, indexing identities with it works where the old expression raises `invalid subscript type`, all-`NaN` and single-column matrices behave, `NA` scores behave like `NaN`, and `TransferData()` on ordinary data still labels every cell.

Against `origin/main`: 5 errors. With this change: 13 passes. Full suite 540 passing; the 2 failures (`test_load_10X.R`) are present unchanged on `main` and are addressed separately in #10454. `R CMD check` Status OK.

## Scope

This stops the crash and makes the affected cells visible. It does not address *why* the scores are `NaN` in the first place, which the reporter also asked about and which looks like a separate question about anchor weighting on small or poorly-matched references.


BenjaminDEMAILLE:

PR #10462, opened 2026-08-31
Compute residuals from the models that cover a feature

### LINKED ISSUE #10441: GetResidual() fails / silently drops features for multi-model SCTAssay when features are not modeled in every model

BenjaminDEMAILLE:
Fixes #10441.

## The bug

`GetResidual()` on an `SCTAssay` holding several models fails whenever a requested feature is modelled by some of them and not others. That is the normal state of an object merged from per-sample `SCTransform` runs.

Using the reporter's reproducer verbatim, with the bundled `pbmc_small`:

| call | on `main` | this PR |
| --- | --- | --- |
| `na.rm = FALSE` | `Error: number of rows of matrices must match` | returns residuals |
| `na.rm = TRUE` | `Error: subscript out of bounds` | returns residuals, drops the partial feature with the existing warning |

On the reporter's 12-model object `na.rm = FALSE` gives `Error: object 'new_residual' not found`, which is the same defect reached by a different route.

## Root cause

Two defects, one in each half.

**`FetchResidualSCTModel()`** splits the request into features the model has (`intersect_features`) and features it does not (`diff_features`). The computation runs only in the `length(diff_features) == 0` branch. The other branch warns, returns early when *no* requested feature is present, and otherwise falls through to

```r
return(new_residual)
```

for a variable that path never assigned. So a model covering *some* of the requested features computed nothing at all. That is the `object 'new_residual' not found`, and it is why the feature could not be obtained even though it is modelled somewhere.

**`GetResidual()`** then combined the per-model results with `Reduce(cbind, ...)`, which requires matching row counts. Models covering different feature sets cannot be combined that way even once each returns something.

## The change

- The partial branch computes residuals for the features the model does cover, by recursing with `intersect_features`, rather than falling through.
- The results are filled into the destination by name instead of `cbind`, so each model contributes what it has.

## Result

`na.rm = FALSE` now returns a partially modelled feature with residuals where it is modelled:

```
scale.data: 124x80
TCL1A present: TRUE | GNLY present: TRUE
  finite in model-1 cells: 40 / 40
  finite in model-2 cells: 0 / 40      (NA: not modelled there)
  range over model-1 cells: -0.343 .. 1.29
```

which is what the report asks for. `na.rm = TRUE` keeps its documented behaviour of dropping such features, with the existing warning, and no longer errors on the way there.

## Tests

New `tests/testthat/test_getresidual_models.R`, built on `pbmc_small` so no external data is needed: the fixture really does contain a feature only one model covers, `na.rm = FALSE` returns rather than failing, the partially modelled feature is finite exactly where its model applies and `NA` elsewhere, `na.rm = TRUE` keeps only features every model covers and leaves no `NA`, and a shared feature's residuals are identical whether or not a partial feature is requested alongside it.

Against `upstream/main`: 1 failure and 4 errors. With this change: 12 passes. Seurat's suite is unchanged, 2 failures and 527 passes before and after. `R CMD check` Status OK.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- Fixed `FetchResiduals(na.rm = FALSE)` failing with \dQuote{number of rows of matrices must match}: the per-model residuals were combined with `cbind`, which cannot combine models that cover different features, so the NA that `na.rm = FALSE` describes was never reached
+
 - Fixed `GetResidual` failing on an `SCTAssay` with several models when a requested feature is modelled by only some of them, with `object 'new_residual' not found` or `number of rows of matrices must match`; residuals for such a feature are now returned for the cells whose model covers it, and `NA` elsewhere ([#10441](https://github.com/satijalab/seurat/issues/10441))
 
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- Fixed `GetResidual` failing on an `SCTAssay` with several models when a requested feature is modelled by only some of them, with `object 'new_residual' not found` or `number of rows of matrices must match`; residuals for such a feature are now returned for the cells whose model covers it, and `NA` elsewhere ([#10441](https://github.com/satijalab/seurat/issues/10441))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
 - Fixed bug in `PercentageFeatureSet` where layer data was incorrectly retrieved prior to finding features with requested pattern ([#10438](https://github.com/satijalab/seurat/pull/10438))
 - Updated `as.SingleCellExperiment` to address conversion case where an object has both original and sketched assay / reductions (differing numbers of cells) ([#10419](https://github.com/satijalab/seurat/pull/10419))
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -466,12 +466,23 @@ GetResidual <- function(
   if (nrow(x = existing.data) > 0){
     new.scale[1:nrow(x = existing.data), ] <- existing.data
   }
+  # Models covering different feature sets return matrices with different rows,
+  # which cbind cannot combine. Fill by name instead, so each model contributes
+  # the features it actually has
   if (length(x = new.residuals) == 1 & is.list(x = new.residuals)) {
-    new.residuals <- new.residuals[[1]]
-  } else {
-    new.residuals <- Reduce(cbind, new.residuals)
+    new.residuals <- list(new.residuals[[1]])
   }
-  new.scale[rownames(x = new.residuals), colnames(x = new.residuals)] <- new.residuals
+  for (residual in new.residuals) {
+    rdim <- dim(x = residual)
+    if (is.null(x = rdim) || any(rdim == 0L)) {
+      next
+    }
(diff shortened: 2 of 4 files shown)


BenjaminDEMAILLE:
Added a second commit here: the same defect one level up, in `FetchResiduals()`.

With the first commit, `FetchResiduals(object, features = ..., na.rm = FALSE)` gets past `object 'new_residual' not found` and then hits

```
Error in f(init, x[[i]]) : number of rows of matrices must match (see arg 2)
```

from

```r
new.residuals <- Reduce(cbind, new.residuals)
```

Models cover different features once an object has been subset or merged, so their matrices have different rows and `cbind` refuses them. That is the whole of `na.rm = FALSE`, documented as *return NA for residual value in scale.data*: the NA can never be reached, because the matrices that would carry it cannot be bound.

The result is now filled by name, the same way the first commit fills `GetResidual()`, so each model contributes the features it has and the rest stay NA. Measured on a two-model object, asking for two features both models fit and two only one model fits:

| | rows returned | NA rows |
| --- | --- | --- |
| `na.rm = FALSE` | 4 | 2 |
| `na.rm = TRUE` | 2 | 0 |

which is what the documentation describes for each.

One more test; suite 1195 passing.

---

BenjaminDEMAILLE:

Issue #7792, opened 2023-09-13
LoadAkoya function from QuPath table


JessicaP94:
Hello, I am trying to import a table from QuPath but keeping receiving the following error:
``
Error in r[i1] - r[-length(r):-(length(r) - lag + 1L)] : 
  non-numeric argument to binary operator
``
My script:
``
codex.obj <- LoadAkoya(filename = "~/JP/230913_for_seurat.csv", 
                       type = "qupath")
``

Can you help me to solve this problem?

Thank you

alikhuseynov:
Could you please show the whole error output and `traceback()`
What does the `ReadAkoya()` outputs in this case?

JessicaP94:
> Could you please show the whole error output and `traceback()` What does the `ReadAkoya()` outputs in this case?

``
> codex.obj_JP <- LoadAkoya(filename = "~/Jessica-KP/230913_for_seurat.csv", 
+                        type = "qupath")
Error in r[i1] - r[-length(r):-(length(r) - lag + 1L)] : 
  non-numeric argument to binary operator
> traceback()
13: diff.default(newX[, i], ...)
12: FUN(newX[, i], ...)
11: apply(X = apply(X = coords, MARGIN = 2L, FUN = range), MARGIN = 2L, 
        FUN = diff)
10: mean(x = apply(X = apply(X = coords, MARGIN = 2L, FUN = range), 
        MARGIN = 2L, FUN = diff))
9: .AutoRadius(coords = coords)
8: radius %||% .AutoRadius(coords = coords)
7: CreateCentroids.default(coords = coords, nsides = nsides, radius = radius, 
       theta = theta)
6: CreateCentroids(coords = coords, nsides = nsides, radius = radius, 
       theta = theta)
5: CreateFOV.data.frame(coords = data$centroids, type = "centroids", 
       key = "fov", assay = assay)
4: CreateFOV(coords = data$centroids, type = "centroids", key = "fov", 
       assay = assay)
3: withCallingHandlers(expr, warning = function(w) if (inherits(w, 
       classes)) tryInvokeRestart("muffleWarning"))
2: suppressWarnings(expr = CreateFOV(coords = data$centroids, type = "centroids", 
       key = "fov", assay = assay))
1: LoadAkoya(filename = "~/Jessica-KP/230913_for_seurat.csv", type = "qupath")
``

I cannot show the ``ReadAkoya()`` output because it's too long.

Thank you for your help

alikhuseynov:
[...]

PR #10463, opened 2026-08-31
Match QuPath centroid columns as they are actually named

### LINKED ISSUE #9102: Seurat::LoadAkoya produced Error in data.frame(x = mtx[[xpos]], y = mtx[[ypos]], cell = rownames(x = mtx), : arguments imply differing number of rows: 0, 116335
Hi there, 

Thanks for the tool. 

I attempted to load in the csv dataframe as below but encountered an error about dimension 

`Error in data.frame(x = mtx[[xpos]], y = mtx[[ypos]], cell = rownames(x = mtx), :
arguments imply differing number of rows: 0, 116335`

Anyone has a potential solution?

Thank you for your help!

```
Seurat::LoadAkoya(filename = ".../PlanA_Trial4_ROI1234_CelldetectionMeasurements.csv",
    type = "qupath", fov = "PlanA_Trial4", filter = NA, assay = "Polaris")
```

Screenshot 2024-07-10 at 17 23 17

BenjaminDEMAILLE:
Fixes #9102.

## The bug

`ReadAkoya(type = "qupath")` fails on exports whose centroid columns are not spelled exactly `Centroid X` and `Centroid Y`:

```
Error in data.frame(x = mtx[[xpos]], y = mtx[[ypos]], cell = rownames(x = mtx), :
  arguments imply differing number of rows: 0, 116335
```

## Root cause

The columns are located with a single literal grep:

```r
xpos <- sort(grep(pattern = 'Centroid X', x = colnames(x = mtx), value = TRUE), decreasing = TRUE)[1L]
```

With no match that returns `character(0)`, `sort(...)[1L]` makes it `NA`, and `mtx[[NA]]` is a zero-length vector. The frame is then built from one column of length zero and one of length *n*, which is the reported error. It names neither the column that is missing nor what was expected, so there is nothing to act on.

QuPath does not spell these consistently: units may be appended, case varies between versions, and `read.csv(check.names = TRUE)` turns the space into a dot before Seurat ever sees the name.

| column name | `main` | this PR |
| --- | --- | --- |
| `Centroid X` | OK | OK |
| `Centroid X um` | OK | OK |
| `centroid x` | **error** | OK |
| `Centroid.X` | **error** | OK |

## The change

Match on `centroid[ ._]*x` ignoring case, which covers all four spellings. When there is genuinely no such column, say so rather than letting it become a row-count mismatch:

```
Could not find the centroid X and Y columns in this QuPath export; a column
named like 'Centroid X' is expected. Columns present: 'id', 'area', 'posX',
'posY', 'Cell: Mean CD3', 'Cell: Mean CD8'
```

The neighbouring grep for `Cell: Mean` already uses `ignore.case = TRUE`, so this also removes an inconsistency rather than introducing a new convention.

## Tests

New `tests/testthat/test_read_akoya_qupath.R`, which writes small QuPath-style CSVs so no vendor export is needed: each of the four spellings is read, the coordinates match the values in the file rather than merely being non-empty, and a file with no centroid column at all raises the explanatory error.

Against `upstream/main`: 2 failures and 2 errors. With this change: 9 passes. Suite unchanged at 536 passing, `R CMD check` Status OK.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- Fixed `ReadAkoya(type = "qupath")` failing with `arguments imply differing number of rows` when the centroid columns are not named exactly `Centroid X`/`Centroid Y`; the match now tolerates units, case and the dot that `read.csv` substitutes, and a genuinely missing column is reported by name ([#9102](https://github.com/satijalab/seurat/issues/9102))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
 - Fixed bug in `PercentageFeatureSet` where layer data was incorrectly retrieved prior to finding features with requested pattern ([#10438](https://github.com/satijalab/seurat/pull/10438))
 - Updated `as.SingleCellExperiment` to address conversion case where an object has both original and sketched assay / reductions (differing numbers of cells) ([#10419](https://github.com/satijalab/seurat/pull/10419))
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- `ReadAkoya` now reads QuPath centroid columns that come back as text, converting a decimal comma and naming the column when the values are not numbers, instead of failing later with \dQuote{non-numeric argument to binary operator} ([#7792](https://github.com/satijalab/seurat/issues/7792))
+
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:
Added a second QuPath failure to this PR, #7792, since it is the same few lines and the same export.

A QuPath table can carry its centroids as **text**: the decimal separator is a comma under some locales, and units are sometimes written into the values. Either way the column reads as character, and the failure surfaces much later, from inside `diff()`:

```
Error in r[i1] - r[-length(r):-(length(r) - lag + 1L)] :
  non-numeric argument to binary operator
```

which names neither the column nor the file, and is the traceback in that issue.

Now a decimal comma is converted, with a warning saying so, and anything else stops with what is actually in the column:

```
The Centroid X um column does not hold numbers: '45.53 um', '41.01 um', '81.09 um'.
Centroids must be plain numbers, without units
```

Two more tests, both failing without the change. Suite 1192 passing.

---

BenjaminDEMAILLE:

Issue #9182, opened 2024-08-02
Incorrect nCount_Spatial and nFeature_Spatial for 10x Visium Spatial Gene + Protein Expression


joonlee3:
We have recently received a 10x Visium Spatial **Gene** + **Protein Expression** data. When I created Seurat objects for the dataset. Seurat does not provide nCount_Spatial and nFeature_Spatial for gene expression and protein expression data individually.

Here is the UMI count for the gene expression
```r
> head(colSums(mat.gene))
AACAATGGAACCACAT-1 AACAATGTGCTCCGAG-1 AACAATTATCTGCGCC-1 AACACCAGCCTACTCG-1
              4127               1861               2162               6802
```

Here is the UMI count for the protein (antibody) expression
```r
> head(colSums(mat.antibody))
AACAATGGAACCACAT-1 AACAATGTGCTCCGAG-1 AACAATTATCTGCGCC-1 AACACCAGCCTACTCG-1
            742623             440772             305467             650371
```

Here is the total UMI count (Gene + Protein)
```r
> head(colSums(mat.gene))[1:4] + head(colSums(mat.antibody))[1:4]
AACAATGGAACCACAT-1 AACAATGTGCTCCGAG-1 AACAATTATCTGCGCC-1 AACACCAGCCTACTCG-1 
            746750             442633             307629             657173 
```

These total UMI counts match  "nCount_Spatial" in the Seurat object `obj1`
```r
> head(obj1)
                      orig.ident nCount_Spatial nFeature_Spatial sample.id project
AACAATGGAACCACAT-1 SeuratProject         746750             2658      outs   NSCLC
AACAATGTGCTCCGAG-1 SeuratProject         442633             1543      outs   NSCLC
AACAATTATCTGCGCC-1 SeuratProject         307629             1601      outs   NSCLC
AACACCAGCCTACTCG-1 SeuratProject         657173             3980      outs   NSCLC
```

```r
> sessionInfo()
R version 4.4.1 (2024-06-14)
Platform: x86_64-pc-linux-gnu
Running under: Rocky Linux 8.7 (Green Obsidian)

Matrix products: default
BLAS/LAPACK: /usr/lib64/libopenblas-r0.3.15.so;  LAPACK version 3.9.0

locale:
 [1] LC_CTYPE=C.UTF-8       LC_NUMERIC=C           LC_TIME=C.UTF-8        LC_COLLATE=C.UTF-8    
[...]

PR #10464, opened 2026-08-31
Read 10x HDF5 with rhdf5 when hdf5r is unavailable

BenjaminDEMAILLE:
## The problem

`Read10X_h5()` requires `hdf5r`, which builds against the *system* HDF5 and does not compile against 1.12 or newer:

```
error: call to undeclared function 'H5FD_family_init'
ERROR: compilation failed for package 'hdf5r'
```

1.14 is what current package managers ship (Homebrew is on 1.14.6), so on an otherwise working, up-to-date system `hdf5r` simply cannot be installed. Every 10x `.h5` file then becomes unreadable, and `Read10X_h5()`, `Load10X_Spatial()` and `LoadXenium()` all stop at

```
Please install hdf5r to read HDF5 files
```

There is no way round it from Bioconductor either: `Rhdf5lib` bundles 1.14.6, the same version.

I hit this trying to reproduce #9182 and could not read a Visium `.h5` at all. It seems worth fixing on its own, since the same wall stands between a lot of users and their data.

## The change

`rhdf5` bundles its own HDF5 through `Rhdf5lib` and therefore always installs. This adds `.Read10Xh5Rhdf5()`, used **only when `hdf5r` is absent**, so nothing changes for anyone who has `hdf5r` working today. `hdf5r` stays the preferred reader.

It mirrors the existing reader: CellRanger 2 and 3 layouts, `use.names`, `unique.features`, the multimodal split, and the protein scaling factor.

One deliberate difference, in the direction of robustness: the modality split keys off the `feature_type` **dataset** rather than the `features` **group**. The current code checks for the group and then reads `feature_type` unconditionally, so a file carrying feature names without types errors; this path reads it.

## Verification

Matrices written out in the 10x layout and read back:

| case | result |
| --- | --- |
| single modality | values and dimnames identical to the source matrix |
| Gene Expression + Antibody Capture | split into the same two matrices, correct dims and values |
| `use.names = FALSE` | feature ids used |
| duplicate feature names | `make.unique` applied, or preserved with `unique.features = FALSE` |
| `protein_scaling_factor` | applied to Protein Expression only |

## Tests

New `tests/testthat/test_read10x_h5_rhdf5.R`, which writes its own 10x-layout HDF5 files with `rhdf5`, so no vendor data is needed. Guarded with `skip_if_not_installed("rhdf5")`. 13 assertions.

`rhdf5` added to `Suggests`. Suite 540 passing, `R CMD check` Status OK.

## Note

This does not address #9182 itself, which is about `nCount_Spatial` summing gene and protein counts into one assay. It removes the barrier to working on it, and to reading 10x data at all on a current system.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -116,6 +116,7 @@ Suggests:
     mixtools,
     monocle,
     presto,
+    rhdf5,
     rsvd,
     R.utils,
     Rfast2,
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ### Additions
 
+- `Read10X_h5` falls back to `rhdf5` when `hdf5r` is not installed, so 10x HDF5 files remain readable on systems where `hdf5r` cannot be built against a current system HDF5 ([#9182](https://github.com/satijalab/seurat/issues/9182))
+
 ### Fixes
 
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -2709,7 +2709,11 @@ ReadXenium <- function(
 
   has_dt <- requireNamespace("data.table", quietly = TRUE) && requireNamespace("R.utils", quietly = TRUE)
   has_arrow <- requireNamespace("arrow", quietly = TRUE)
-  has_hdf5r <- requireNamespace("hdf5r", quietly = TRUE)
+  # Read10X_h5() reads through hdf5r or, failing that, rhdf5; gate on whether
+  # either is available rather than on hdf5r alone, or the .h5 is skipped on a
+  # system where it is perfectly readable
+  has_h5 <- requireNamespace("hdf5r", quietly = TRUE) ||
+    requireNamespace("rhdf5", quietly = TRUE)
 
   binary_to_string <- function(arrow_binary) {
     if(typeof(arrow_binary) == 'list') {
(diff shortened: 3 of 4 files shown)


BenjaminDEMAILLE:
I have opened #10467, which replaces `hdf5r` with `rhdf5` outright rather than falling back to it.

These two are alternatives, not companions. **Take whichever you prefer and I will close the other**; I am not asking for both.

- **This PR (#10464)** is conservative: `hdf5r` stays the reader when it is present, and nothing changes for anyone who has it working today. Only `Read10X_h5` gains the fallback, so `Read10X_probe_metadata` and the Vizgen segmentation reader still require `hdf5r`.
- **#10467** ports all four HDF5 readers, so a system without `hdf5r` can use every one of them.

The case for the stronger version is that `hdf5r` does not compile against HDF5 1.12 or newer, which is what current package managers ship, so it can be uninstallable on an otherwise working system, and `Rhdf5lib` bundles the same 1.14 so Bioconductor is no way round it. The case for this one is that `hdf5r` works fine for anyone who already has it, and CRAN-only users may prefer not to pull from Bioconductor.

That trade-off is yours rather than mine, which is why both are on the table.

---

BenjaminDEMAILLE:
Closing in favour of #10467, which ports all four HDF5 readers rather than adding rhdf5 as a fallback for one of them.

The reason for preferring the fuller version: hdf5r does not build against HDF5 1.12 or newer, which is what current package managers ship, so it can be uninstallable on an otherwise working system, and Rhdf5lib bundles the same 1.14 so Bioconductor is no way round it. A fallback for Read10X_h5 alone would still leave Read10X_probe_metadata and the Vizgen segmentation reader unusable there.

If you would rather have the conservative version after all, say so and I will reopen this one and close #10467 instead.

PR #10466, opened 2026-08-31
Load Xenium panels that carry a single feature type

### LINKED ISSUE #9142: LoadXenium() issue on the 480genes standalone panel, Xenium V1
Hi, 
      I got this error when I was using Seurat 5.1.0 to load Xenium data, which was generated with 480genes standalone panel.
      It showed
"Error in `CreateAssayObject()`:
! No cell names (colnames) names present in the input matrix
Run `rlang::last_trace()` to see where the error occurred."
image

And when I manually loaded the cell_feature_matrix folder into R, I realized that the reason is this standalone panel has no "Unassigned Codeword":
image
It only has "Gene expression", "Negative Control Codeword", "Negative Control Probe". 
I think maybe for the panel with 480genes, the 10X Genomics team didn't include the "Unassigned Codeword" for some reason. (such like all Codewords were used for the genes and negative probes.)

Anyway, is there a way to skip that error?

Thanks!

BenjaminDEMAILLE:
Fixes #9142. **Builds on #10464** (rhdf5 fallback), which is its base branch.

## The bug

`LoadXenium()` fails on a standalone gene panel:

```
Error in `CreateAssayObject()`:
! No cell names (colnames) names present in the input matrix
```

`Read10X_h5()` returns a **named list** only when the file holds more than one feature type. A panel with control probes alongside the genes gets a list; a standalone gene panel, such as the 480-gene V1 panel in the report, gets a bare matrix. `LoadXenium()` indexes it as a list either way:

```r
CreateSeuratObject(counts = data$matrix[["Gene Expression"]], assay = assay)
```

On a bare `dgCMatrix` that yields nothing usable. On the reporter's version it surfaced as the missing-colnames error above; on current `main` as `this S4 class is not subsettable`. Same cause.

| panel | `main` | this PR |
| --- | --- | --- |
| gene expression only | **error** | loads, assay `Xenium` |
| genes + control probes | loads, assays `Xenium`, `ControlProbe` | unchanged |

## The change

Take the matrix as-is when it is not a list. The loop that adds control probes as their own assays already iterates over `names(data$matrix)`, which is `NULL` for a bare matrix, so it does nothing and needs no change.

## Tests

New `tests/testthat/test_loadxenium_panels.R`, which writes Xenium-style outputs so no vendor data is needed: a gene-only panel loads with a single assay, its counts match the file, a panel with control probes still splits into two assays, and the gene features are the same either way.

Against the base branch: 3 errors. With this change: 8 passes. Suite 548 passing.

## Why it also needed #10464

`ReadXenium()` decided whether to read `cell_feature_matrix.h5` by checking for `hdf5r` specifically, so on a system where `hdf5r` cannot be built it skipped a perfectly readable `.h5` and stopped with `Xenium outputs were incomplete: missing cell_feature_matrix`. That gate is fixed in #10464, which is why this branch sits on top of it.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -116,6 +116,7 @@ Suggests:
     mixtools,
     monocle,
     presto,
+    rhdf5,
     rsvd,
     R.utils,
     Rfast2,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+### Fixes
+
+- Fixed `LoadXenium` failing on panels that carry a single feature type, such as the standalone gene panels, where `Read10X_h5` returns a bare matrix rather than a named list ([#9142](https://github.com/satijalab/seurat/issues/9142))
+
 ### Additions
 
 - `Read10X_h5` falls back to `rhdf5` when `hdf5r` is not installed, so 10x HDF5 files remain readable on systems where `hdf5r` cannot be built against a current system HDF5 ([#9182](https://github.com/satijalab/seurat/issues/9182))
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ### Additions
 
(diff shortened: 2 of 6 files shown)


BenjaminDEMAILLE:

PR #10468, opened 2026-08-31
Recorrect SCT counts when a model sits above the shared depth

### LINKED ISSUE #9130: Warning to PrepSCTFindMarkers when I already have
Hi, trying to run `FindAllMarkers` on SCT transformed data:
```
# find markers for each cluster
seu <- PrepSCTFindMarkers(seu, assay="SCT_Scaled_Centered")
seu <- SetIdent(seu, value="SNN_PCA_SCT_Scaled_Centered_res0.1")
all_markers <- FindAllMarkers(seu, assay="SCT_Scaled_Centered",
                              logfc.threshold=0.1,
                              test.use="wilcox",
                              min.pct=0.01,
                              min.cells.feature=3,
                              min.cells.group=3,
                              return.thresh=0.01,
                              recorrect.umi=F)
```

```
Warning: The following tests were not performed: 
Warning: When testing harmony3 versus all:
	Object contains multiple models with unequal library sizes. Run `PrepSCTFindMarkers()` before running `FindMarkers()`.
```

I've already run `PrepSCTFindMarkers` and I also added `recorrect.umi=F` just in case this counted as "subsetting" but it hasn't helped.

Any advice please?

BenjaminDEMAILLE:
Fixes #9130, where `PrepSCTFindMarkers()` and `FindMarkers()` contradict each other and leave no way to run differential expression on the object:

```r
> sub <- PrepSCTFindMarkers(sub)
Minimum UMI unchanged. Skipping re-correction.
> FindMarkers(sub, ident.1 = "a", ident.2 = "b")
Error: Object contains multiple models with unequal library sizes.
Run `PrepSCTFindMarkers()` before running `FindMarkers()`.
```

## Cause

`PrepSCTFindMarkers()` returns early when

```r
all(unlist(x = model_median_umis) > min_median_umi)
```

`min_median_umi` is the smallest median UMI observed across the models' cells, and the models sitting **above** it are precisely the ones whose counts are on the wrong scale. So this skips the objects that need recorrecting. The test for "nothing to do" is equality: every model already at the shared depth. It was `==` until d97a5f6f.

`FindMarkers.SCTAssay()` meanwhile requires `model_median_umis == min(observed_median_umis)`, so an object that takes the early return is rejected by the very function it was being prepared for.

## When it is reached

Whenever cells are removed after `SCTransform`. The observed medians are recomputed from the remaining cells, while each model's `median_umi` keeps the depth recorded at fit time, so dropping low-UMI cells from one sample can put every stored median above the new minimum. That matches the reports in #9130: merge, SCTransform, then subset, and every subsequent `PrepSCTFindMarkers()` is a no-op.

## Test

Two tests on the existing two-model `sct_merged` fixture: subsetting to the ten lowest-UMI cells of one model puts both stored medians above the shared depth; after the fix `PrepSCTFindMarkers()` brings both to it and `FindMarkers()` runs. The second test pins the early return that should happen: an already-prepared object still reports `Skipping re-correction`.

Both fail without the change (3 failures) and pass with it. Full suite `NOT_CRAN=true`: 1183 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- Fixed `PrepSCTFindMarkers` skipping the objects that need it most: after cells are removed, a model's observed median UMI can fall below the depth its `median_umi` records, and those objects were returned untouched with \dQuote{Minimum UMI unchanged}, after which `FindMarkers` refused to run on them with \dQuote{Object contains multiple models with unequal library sizes} ([#9130](https://github.com/satijalab/seurat/issues/9130))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
 - Fixed bug in `PercentageFeatureSet` where layer data was incorrectly retrieved prior to finding features with requested pattern ([#10438](https://github.com/satijalab/seurat/pull/10438))
 - Updated `as.SingleCellExperiment` to address conversion case where an object has both original and sketched assay / reductions (differing numbers of cells) ([#10419](https://github.com/satijalab/seurat/pull/10419))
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

PR #10470, opened 2026-08-31
Put cells back in order after scaling by split

BenjaminDEMAILLE:
`ScaleData(split.by = )` gives cells another cell's scaled values. Nothing warns, and the resulting matrix looks perfectly well formed.

## Cause

Each split is scaled separately and the pieces are bound back together:

```r
scaled.data <- do.call(what = 'cbind', args = scaled.data)
...
dimnames(x = scaled.data) <- object.names
```

`cbind` puts the cells in split order (all of split a, then all of split b), and the dimnames are then overwritten with the object's own order. Every cell whose position moved is labelled with the wrong name.

```r
future::plan("multicore", workers = 2)
sequential <- LayerData(ScaleData(object, split.by = "group"), layer = "scale.data")   # plan sequential
parallel   <- LayerData(ScaleData(object, split.by = "group"), layer = "scale.data")

equal as-is:                                FALSE
equal after undoing the split permutation:  TRUE
```

## Reach

Three places bind splits: the two parallel paths (scaling, and regression), and the **sequential** regression path. So this is not only a multi-worker problem: with one worker,

```r
ScaleData(object, split.by = "sample", vars.to.regress = "nCount_RNA")
```

is affected too, since that path always assembles the result by `cbind`ing the splits. Plain `split.by` scaling is correct when `nbrOfWorkers() == 1`, because that path fills a preallocated matrix by name.

Anything downstream of the scaled data is affected: PCA, and everything that follows from it.

## Fix

Reorder the columns back to the object's order before the dimnames are restored, in all three places.

## Tests

`tests/testthat/test_scaledata_split.R`: parallel `split.by` scaling equals sequential; each cell's values equal what it gets when its own split is scaled alone; the same two for `vars.to.regress`; and the regression case with no workers at all, which needs no multicore support to catch the bug.

All four fail without the change and pass with it. Full suite `NOT_CRAN=true`: 1183 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- Fixed `ScaleData` giving cells another cell's values when `split.by` is used: the splits were bound back together in split order and the dimnames were then overwritten with the object's own order. This affected every cell whose position moved, with more than one worker, or with any plan when `vars.to.regress` is also given
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
 - Fixed bug in `PercentageFeatureSet` where layer data was incorrectly retrieved prior to finding features with requested pattern ([#10438](https://github.com/satijalab/seurat/pull/10438))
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

PR #10459, opened 2026-08-31
Send each worker its block of data, not the whole matrix

BenjaminDEMAILLE:
Fixes #10406.

## The bug

With more than one `future` worker, `NormalizeData()` fails on a moderately sized object:

```r
plan("multicore", workers = 4)
obj <- CreateSeuratObject(counts = <2000 x 14000 dgCMatrix>)
NormalizeData(obj)
#> Error: The total size of the 6 globals exported for future expression ('FUN()')
#>   is 1.09 GiB. This exceeds the maximum allowed size 500.00 MiB
```

The revealing part is what the globals are. `FUN` alone accounts for **697 MiB, for a function**.

## Root cause

The data is chunked, but `future_lapply()` iterated over block *indices* and the worker function referenced `object`:

```r
normalized.data <- future_lapply(
  X = 1:ncol(x = chunk.points),
  FUN = function(i) {
    block <- chunk.points[, i]
    data <- object[, block[1]:block[2], drop = FALSE]   # <- object
    ...
  }
)
```

So every worker was sent the entire matrix and then subset it locally. Memory scaled with the number of workers rather than staying flat, which is the report in #10406, and past a certain size `future`'s export limit rejects the job outright.

## The change

Two parts, because the first alone was not enough:

1. **Split before dispatching**, and iterate over the blocks, so each worker receives only its own. This dropped the globals from six to four and left the size unchanged at 836 MiB.
2. **Move the worker body to a package-level `.NormalizeBlock()`.** A closure defined inside `NormalizeData.default()` carries that frame as its environment, and the frame holds the matrix, so the data followed the function no matter how the iteration was written.

Measured on a 2000 x 14000 object with four workers:

| | globals | exported | result |
| --- | --- | --- | --- |
| `origin/main` | 6 | 1.09 GiB | fails |
| after step 1 only | 4 | 836 MiB | fails |
| this PR | 1 | **2.63 MiB** | completes |

The remaining 2.63 MiB is the function itself and does not grow with the data.

## Tests

New `tests/testthat/test_normalize_parallel.R`:

- `LogNormalize`, `CLR` and `RC` give results identical to the sequential path under four workers.
- **What is sent to workers does not grow with the data.** Rather than asserting a fixed number, this measures the reported export size for a small and an eight-times-larger object and requires them to be equal, and to be smaller than the matrix. That is the invariant this PR restores.
- The sequential path still matches a hand-computed log-normalization.

[...]

Issue #10420, opened 2026-06-29
BUG: Setting multiple cores in future causes IntegrateLayers to go into an infinite loop


chlige:
### Issue Description

We have noticed that if one uses plan() to set workers to any value greater than 1 IntegrateLayers will get stuck in an infinite loop.   
In the code example, if the plan() line is commented out or workers is set to 1, the script will finish in about 1 minute and 15 seconds.

When checking a process list (top), we can see the R process running using 1.7 GB of resident memory and 99% CPU.   If we do not kill the process it will continue to run like this for days.   We have waited up to 10 days and it still will not complete this step.   

We are running this on a HPC compute node with 64 cores and 1 TB of memory.  The R script is running via a SLURM batch job that was allocated 200 GB of memory and 4 cores.  So, there should not be any resource constraints that should cause the process to hang.

### Reproducing Code Example

```R
library(Seurat)
library(patchwork)

# Setting for parallelization
library(future)
# Set maxSize for future to 10 GB
options(future.globals.maxSize = 10 * 1e9)
plan("multicore", workers=4)

sessionInfo()

# Set the clustering resolutions
resolutions <- c(0.5, 1, 2)

# load dataset
ifnb <- readRDS("ifnb.rds")

# split the RNA measurements into two layers one for control cells, one for stimulated cells
ifnb[["RNA"]] <- split(ifnb[["RNA"]], f = ifnb$stim)

# run standard analysis workflow
ifnb <- NormalizeData(ifnb)
ifnb <- FindVariableFeatures(ifnb)
ifnb <- ScaleData(ifnb)
ifnb <- RunPCA(ifnb)

# Perform clustering of unintegrated PCA
ifnb <- FindNeighbors(ifnb, dims = 1:30, reduction = "pca")
ifnb <- FindClusters(ifnb, resolution = resolutions,
                     cluster.name = paste0("seruat_clusters.pca.", resolutions))

# Create UMAP reduction for unintegrated PCA
ifnb <- RunUMAP(ifnb,
                dims = 1:30,
                reduction = "pca",
                reduction.name = "umap.unintegrated")

# Perform RPCA integration
cat("Run Integration\n")
ifnb <- IntegrateLayers(object = ifnb,
                        method = RPCAIntegration,
                        orig.reduction = "pca",
                        new.reduction = "integrated.rpca",
                        verbose = TRUE)

# re-join layers after integration
cat("JoinLayers\n")
[...]

PR #10471, opened 2026-08-31
Send each worker its block of the data, not the whole matrix

BenjaminDEMAILLE:
`ScaleData()` exports the whole matrix to every worker, whatever the chunking.

**Based on #10470** (`ScaleData` mislabelling cells under `split.by`), which touches the same block and should go in first; this branch is that one plus one commit. If you would rather have them apart, say so and I will restructure.

## Cause

The worker functions are closures created inside `ScaleData.default()`, so they carry that frame as their environment, and the frame holds the full matrix. The chunking happens inside the worker, so the chunk boundaries are all that changes per task; the data itself is a global. Both parallel paths do this, scaling and regressing variables out.

The visible symptom is the run stopping before any work is done:

```
Error: The total size of the 5 globals exported for future expression ('FUN()')
is 1.09 GiB. This exceeds the maximum allowed size 500.00 MiB
```

and, when the limit is raised enough to get past it, every worker holding a copy of the whole dataset. This is the same failure I hit while trying to reproduce #10420, and the same shape as #10459 for `NormalizeData`.

## Fix

Split the data before dispatching and pass the blocks as `X`, with the work moved into package-level `.ScaleBlock()` and `.RegressOutBlock()` so that what is exported is a constant. Each worker then receives only its own block.

## Tests

`tests/testthat/test_scaledata_parallel.R`: parallel scaling equals sequential, with and without `split.by`; parallel regression equals sequential; the size `future` reports as exported does not grow with the data (measured by setting `future.globals.maxSize` to 1 KiB and reading the reported size out of the error, 8x data gives the same exported size); and the sequential result still matches `t(scale(t(data)))`.

The size tests fail without the change and pass with it. Full suite `NOT_CRAN=true`: 1192 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- `ScaleData` now sends each worker its own block of the data instead of exporting the whole matrix to every worker, which on a moderately sized object exceeds `future.globals.maxSize` before any work starts ([#10420](https://github.com/satijalab/seurat/issues/10420))
+
 - Fixed `ScaleData` giving cells another cell's values when `split.by` is used: the splits were bound back together in split order and the dimnames were then overwritten with the object's own order. This affected every cell whose position moved, with more than one worker, or with any plan when `vars.to.regress` is also given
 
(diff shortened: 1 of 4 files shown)


BenjaminDEMAILLE:

Issue #8812, opened 2024-04-23
Warning message with Nanostring FFPE NSCLC CosMx Data Set


Rayan21100:
Hello everyone !

I'm trying to load the Nanostring FFPE NSCLC CosMx Data Set as a Seurat Object in R and trying to follow the vignette on this dataset (https://satijalab.org/seurat/articles/seurat5_spatial_vignette_2#human-lung-nanostring-cosmx-spatial-molecular-imager).

I used the LoadNanostring function as in the vignette : 

`nano.obj <- LoadNanostring(data.dir = "/brahms/hartmana/vignette_data/nanostring/lung5_rep1", fov = "lung5.rep1")`

I have two problems : 

- First I have two warning messages: 
```
Warning messages:
1: In Polygon(coords = cx) : less than 4 coordinates in polygon
2: In Polygon(coords = cx) : less than 4 coordinates in polygon
```
I don't really know what it means and if it's really a problem 

- Then, when my Seurat object is created, my count matrix doesn't have any row/column name as you can see in the picture. This dataset is divided into FOVs and for each FOVs the cell IDs start again from 1, do you think the absence of cell names in the count matrix could be linked to this ? 

<img width="705" alt="Capture d’écran 2024-04-23 à 16 42 51" src="https://github.com/satijalab/seurat/assets/158488103/8f4041ca-98fd-4408-abc8-9d8b5ecd8d38">

For the absence of gene name it is really weird, in "molecules" I can find the list of all the genes as you can see in the picture below (AATK, ABL1, ABL2 etc...). Why is this list recognized under molecules and not as the row names of the count matrix ?

<img width="708" alt="Capture d’écran 2024-04-23 à 16 43 02" src="https://github.com/satijalab/seurat/assets/158488103/a12e74fe-8796-424b-a92e-9643985e0a89">

As the download link for the cell boundaries in the vignette is not working anymore, I asked Nanostring for them, so maybe the problem is from the file they gave me ? I can send the head of each csv file I have if needed,

Thank you in advance for your help !

BenjaminDEMAILLE:
On the first of your two problems, the warnings:

```
1: In Polygon(coords = cx) : less than 4 coordinates in polygon
2: In Polygon(coords = cx) : less than 4 coordinates in polygon
```

[...]

PR #10472, opened 2026-08-31
Keep the metadata column ReadNanostring was asked for

BenjaminDEMAILLE:
Two defects in `ReadNanostring`, found while working on #8812.

## A requested metadata column comes back empty

```r
> colnames(ReadNanostring(data.dir = d, metadata = "Area")$metadata)
character(0)
```

`md[, metadata]` drops to a vector when one column is asked for, so `df$cell <- ...` on the next line has nothing to attach to and the caller receives a metadata frame with no columns. `drop = FALSE`.

## Dense counts are always made sparse

```r
(sum(tx == 0) / length(x = tx)) > ratio
```

`tx` is a data frame at that point, so `length()` is its number of **columns**, not its number of values. The zero fraction comes out roughly `nrow` times too large, so the test passes whatever the data holds and dense counts are converted regardless. The two other readers run the same test on a matrix, where `length()` is correct, so only this one is affected.

## Tests

`tests/testthat/test_read_nanostring.R` writes a CosMx export to a temporary directory (expression matrix including the `cell_ID == 0` rows, per cell metadata, transcript coordinates, cell boundaries), so it needs no vendor data. It covers both defects plus the basic read and `LoadNanostring`, and pins the cell naming (`cell_ID_fov`) and the dropping of unassigned rows.

Three assertions fail without the change. Full suite `NOT_CRAN=true`: 1192 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

## On the rest of #8812

The reporter's other two symptoms are separate. The `less than 4 coordinates in polygon` warnings are addressed in satijalab/seurat-object#303, which names the cells instead of leaving two anonymous `sp` warnings. The missing dimnames I could not reproduce from a well-formed export; a fixture built to the documented layout reads back with the correct row and column names, which this test now pins.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- Fixed `ReadNanostring` returning nothing usable when a single `metadata` column is requested, and deciding whether the counts are sparse from the number of columns rather than the number of values ([#8812](https://github.com/satijalab/seurat/issues/8812))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

Issue #7117, opened 2023-04-03
Nanostring CosMx data error: "No compatible spatial coordinates present"


Ikey20:
Hello, I'm trying to make an ImageDimPlot of Nanostring CosMx data using v5, and keep getting the error "No compatible spatial coordinates present". I followed the vignette but used unsupervised clustering rather than Azimuth results.. Seems I'm missing a step to associate the coordinates data, but it's not clear to me how to do that? I figured when you Load Nanostring to create the object it would pull that info?

AustinHartman:
Hi @Ikey2, could you please provide some code and your `sessionInfo()`?

kcavagnero:
Hi @AustinHartman,

I believe I am having the same issue with my own nanostring cosmx data. ImageDimPlot works with the full object, but fails when I specify a FOV as in the vignette. Please see below some code, a few relevant lines from str(object), and session info. Thank you in advance for your help. 

Best,

Kellen

fullSeurat <- readRDS(fullSeuratPath)
fullSeurat

An object of class Seurat 
1012 features across 39992 samples within 1 assay 
Active assay: Nanostring (1012 features, 0 variable features)
 2 dimensional reductions calculated: pca, umap
 1 spatial field of view present: Skin


str(fullSeurat)

..@ meta.data   :'data.frame':	39992 obs. of  33 variables:
.. ..$ fov                : int [1:39992] 1 1 1 1 1 1 1 1 1 1 ...

  ..@ images      :List of 1
  .. ..$ Skin:Formal class 'FOV' [package "SeuratObject"] with 4 slots


ImageFeaturePlot(fullSeurat, fov = '1', features = "KRT14", max.cutoff = "q95")

Error in ImageFeaturePlot(fullSeurat, fov = "1", features = "KRT14", : No compatible spatial coordinates present
Traceback:

1. ImageFeaturePlot(fullSeurat, fov = "1", features = "KRT14", 
 .     max.cutoff = "q95")
2. stop("No compatible spatial coordinates present")


sessionInfo()

R version 4.2.3 (2023-03-15)
Platform: x86_64-conda-linux-gnu (64-bit)
Running under: CentOS Linux 7 (Core)

Matrix products: default
BLAS/LAPACK: /home/kecavagn/miniconda3/envs/seurat4/lib/libopenblasp-r0.3.21.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
[...]

PR #10473, opened 2026-08-31
Say which field of view is missing, and what is there

BenjaminDEMAILLE:
Addresses #7117, where four people in turn hit the same wall.

`ImageDimPlot()` and `ImageFeaturePlot()` raise

```
Error: No compatible spatial coordinates present
```

for two different situations, and the message distinguishes neither, names neither what was asked for, nor what the object actually holds:

```r
fov <- Filter(f = function(x) x %in% Images(object) && inherits(object[[x]], 'FOV'), x = fov)
if (!length(x = fov)) {
  stop("No compatible spatial coordinates present")
}
```

Either the name given is not an image in the object, or it names an image that is **not** a field of view, such as the `VisiumV1` and `SlideSeq` images that `SpatialDimPlot()` takes. The second is the trap: the object plainly has spatial coordinates, and the user is told it does not.

After:

```
No image named 'Test' in this object. Images present: 'test' (FOV)

'slide' is not a field of view. Images present: 'slide' (SlideSeq). None of them is a
field of view; images of this kind are plotted with SpatialDimPlot() and SpatialFeaturePlot()

This object has no images, so there are no spatial coordinates to plot
```

No behaviour changes: the same calls succeed and the same calls fail, they just say why.

## Tests

`tests/testthat/test_image_plot_fov.R` covers all three messages for both functions, plus a field of view that is present still plotting. Full suite `NOT_CRAN=true`: 1186 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- `ImageDimPlot` and `ImageFeaturePlot` now say why no field of view could be used, naming what was asked for and what the object holds, instead of \dQuote{No compatible spatial coordinates present} ([#7117](https://github.com/satijalab/seurat/issues/7117))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

Issue #9189, opened 2024-08-07
SCTransform with "variable.features.rv.th" cannot be run on split multi-sample data


jasonleongbio:
I have been exploring our own datasets with SCTransform, and recently found that instead of strictly setting 3,000 genes as variable features, using the residual variance cutoff (1.3) yielded results that were slightly different. (e.g. In one of our samples, only 2,592 genes were identified as variable by using the cutoff and somehow the clustering and DEG results seemed to be more reasonable in a biological sense; still exploring the data but the result is interesting to us).

```r
# One sample
seurat_obj <- SCTransform(
    seurat_obj, 
    vst.flavor = "v2",
    variable.features.n = NULL,
    variable.features.rv.th = 1.3,
    verbose = TRUE
)
```

However, when I tried to integrate multiple samples and run SCTransform with the cutoff again (now on the split data), an error occurred (here I use the `ifnb` dataset as a demonstration):

```r
# Split data

ifnb <- LoadData("ifnb")
ifnb[["RNA"]] <- split(ifnb[["RNA"]], f = ifnb$stim)
ifnb_SCT <- SCTransform(
    ifnb, 
    vst.flavor = "v2",
    variable.features.n = NULL,
    variable.features.rv.th = 1.3,
    verbose = TRUE
)
```

The error message suggests that the function `VariableFeatures()` does not accept more than one parameter (one for each individual sample):
```
Running SCTransform on assay: RNA
Running SCTransform on layer: counts.CTRL
vst.flavor='v2' set. Using model with fixed slope and excluding poisson genes.
Variance stabilizing transformation of count matrix of size 12747 by 6548
Model formula is y ~ log_umi
Get Negative Binomial regression parameters per gene
Using 2000 genes, 5000 cells
Found 91 outliers - those will be ignored in fitting/regularization step

Second step: Get residuals using fitted parameters for 12747 genes
Computing corrected count matrix for 12747 genes
Calculating gene attributes
Wall clock passed: Time difference of 10.71726 secs
Determine variable features
Centering data matrix
  |=========================================================================================================================================================================| 100%
Getting residuals for block 1(of 2) for CTRL dataset
Getting residuals for block 2(of 2) for CTRL dataset
Centering data matrix
  |=========================================================================================================================================================================| 100%
Finished calculating residuals for CTRL
Running SCTransform on layer: counts.STIM
vst.flavor='v2' set. Using model with fixed slope and excluding poisson genes.
[...]

PR #10474, opened 2026-08-31
Choose variable features from the ones that were scaled

### LINKED ISSUE #8880: SCTransform some variable features absent from scale.data
After using the `SCTransform` function, I found that some of my variable features are not in `scale.data`. This is best illustrated in the example below. From other posts (https://github.com/satijalab/seurat/issues/4877#issuecomment-894454625, https://github.com/satijalab/seurat/issues/3598#issuecomment-739111462, https://github.com/satijalab/seurat/issues/6987#issuecomment-1457736097), I thought that all of them should be. Shouldn't all variable features be in `scale.data`? Secondly, if they're not, does that mean that those genes aren't being used downstream for PCA, etc?

For more context, my dataset is an immune scRNA-seq dataset with ~23k genes and ~12k cells. Additionally, it has 3 layers from the 3 batches. I tried changing return.only.var.genes, but there were still variable features missing in `scale.data`. I suspect that this issue might be due to how SCTransform integrates multiple layers. Version information: R v4.3.1 and Seurat v5.0.2. 

```
big_obj = Seurat::SCTransform(big_obj, method="glmGamPoi", vars.to.regress=c("pct_mt", "pct_hb", "pct_plat"), variable.features.n=3000, verbose=verbose, return.only.var.genes=TRUE, ncells=5000, conserve.memory=FALSE, vst.flavor="v2")
hvg = VariableFeatures(big_obj)
length(hvg1) # -> 3000
length(which( !hvg %in% rownames(big_obj[["SCT"]]$scale.data) )) # -> 47, this is unexpected
```

BenjaminDEMAILLE:
Fixes #8880 and #9189, both about which variable features come out of `SCTransform` on a split assay.

## Variable features that were never scaled (#8880)

The features are ranked over the union of what the layers chose, but only features present in **every** layer are scaled:

```r
scale_data_features <- intersect(all_features_intersect, var_features_union)
```

So a feature that is variable in one layer and absent from another is returned as variable while having no residuals. `PrepDR` warns, `RunPCA` runs without it, and the object keeps claiming it as variable:

```
> s <- SCTransform(split.object)
variable features: 191   in scale.data: 155   missing: 36
```

That is `pbmc_small` split in two, not a contrived object; the reporter counted 47 on their own data, and others in the thread 149.

The fix selects the variable features from the rows of the residual matrix, which is whatever was actually scaled, on either branch of `return.only.var.genes`. Afterwards every variable feature has residuals, and the count reflects what is usable rather than overstating it.

## A residual variance threshold on split data (#9189)

`variable.features.n = NULL` selects by `variable.features.rv.th` instead of by count. That is documented and works on an unsplit object. On a split assay:

```r
> SCTransform(ifnb, vst.flavor = "v2", variable.features.n = NULL, variable.features.rv.th = 1.3)
Error: 'nfeatures' must be a single positive integer
```

because the missing count is passed to a ranking that requires one. With no count to rank down to, the features the layers selected by their own threshold are what the user asked for, so those are kept.

## Tests

`tests/testthat/test_sctransform_variable_features.R`: every variable feature is in `scale.data` and `RunPCA` does not warn that features were not scaled; a feature silenced in one layer is not returned as variable; and the threshold path runs and produces features that are all scaled.

Four assertions fail without the change. Full suite `NOT_CRAN=true`: 1179 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@
 
 ### Fixes
 
+- Fixed `SCTransform` on a split assay returning variable features that have no residuals, so that `RunPCA` warned they had not been scaled and ran without them; variable features are now chosen from the features that were scaled ([#8880](https://github.com/satijalab/seurat/issues/8880))
+- Fixed `SCTransform` failing on a split assay with `variable.features.n = NULL` and a `variable.features.rv.th` cutoff ([#9189](https://github.com/satijalab/seurat/issues/9189))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
 - Fixed bug in `PercentageFeatureSet` where layer data was incorrectly retrieved prior to finding features with requested pattern ([#10438](https://github.com/satijalab/seurat/pull/10438))
 - Updated `as.SingleCellExperiment` to address conversion case where an object has both original and sketched assay / reductions (differing numbers of cells) ([#10419](https://github.com/satijalab/seurat/pull/10419))
--- a/R/preprocessing5.R
+++ b/R/preprocessing5.R
@@ -1235,12 +1235,31 @@ SCTransform.StdAssay <- function(
   )
   LayerData(assay_out, layer = "scale.data") <- residuals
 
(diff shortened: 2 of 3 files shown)


BenjaminDEMAILLE:

PR #10475, opened 2026-08-31
Say why no variable features can be selected

BenjaminDEMAILLE:
Found while working through satijalab/seurat-object#191, on single-cell objects.

`FindVariableFeatures()` fits a loess to the features that vary:

```r
not.const <- hvf.info$variance > 0
fit <- loess(log10(variance) ~ log10(mean), data = hvf.info[not.const, ], span = span)
```

When nothing varies the fit is handed no rows and fails with

```
Error: invalid 'x'
```

which says nothing about the data, and comes from inside `loess`'s C code so the traceback does not help either.

Two ways to reach it:

- **a single cell**, where the variance of one value is `NA`. `NA` then slips into the subset as well, so `hvf.info[not.const, ]` carries rows of `NA` instead of dropping them, and on the way to `loess` you can also get `missing value where TRUE/FALSE needed`.
- **constant data**, where every variance is 0. Small subsets and control panels hit this.

## Fix

Treat `NA` as no variance, and stop with what is actually wrong:

```
Cannot select variable features: no feature has non-zero variance, as the data has 1 cell
Cannot select variable features: no feature has non-zero variance
```

Applied in all four places that fit this loess: `FindVariableFeatures.V3Matrix`, and the dense, sparse and BPCells paths of the v5 method.

No behaviour changes for data that has any variance at all: the call that used to fail still fails, and the call that used to work still works.

## Tests

`tests/testthat/test_variable_features_variance.R`: the single-cell message, the constant-data message, and a mixed object where five constant features are simply ranked out while the other twenty are selected normally.

Two assertions fail without the change. Full suite `NOT_CRAN=true`: 1183 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- `FindVariableFeatures` now says why it cannot select features when no feature has non-zero variance, such as on a single cell, instead of failing inside `loess` with \dQuote{invalid 'x'}; a variance of `NA` is also no longer treated as non-constant ([#191](https://github.com/satijalab/seurat-object/issues/191))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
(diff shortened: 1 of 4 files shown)


BenjaminDEMAILLE:

PR #10476, opened 2026-08-31
Keep the array coordinates a Visium run records

BenjaminDEMAILLE:
Addresses satijalab/seurat-object#229, where a spatial object loaded on 5.0.2 no longer carries what it did on 5.0.1.

## What is lost

A `VisiumV1` image kept all five columns of the tissue positions file in its `coordinates` slot:

```
                   tissue row col imagerow imagecol
AACAATGGAACCACAT-1      1  54  32     1293      657
```

A `VisiumV2` image is an `FOV`, which holds only the pixel position of each spot. `Read10X_Image` reads all five columns and passes two of them on:

```r
fov <- CreateFOV(coordinates[, c("imagecol", "imagerow")], type = "centroids", ...)
```

So the array row and column, and whether the spot is over tissue, are dropped at load and recorded nowhere in the object. Measured on the repository's own Visium fixture:

| | columns |
| --- | --- |
| tissue positions file | `tissue, row, col, imagerow, imagecol` |
| `VisiumV1@coordinates` | `tissue, row, col, imagerow, imagecol` |
| `GetTissueCoordinates(VisiumV2)` | `x, y, cell` |

with `x` equal to `imagecol` and `y` equal to `imagerow`, both verified against the file.

The array indices are not derivable from the pixel positions, so a tool that used them, such as the one in that issue, has nothing to fall back on.

## Change

`Load10X_Spatial` adds them to the object's meta data as `array_row`, `array_col` and `in_tissue`, named as spaceranger names them. The image is untouched, so plotting is unaffected, and this is additive: nothing that exists today changes.

An object built from a supplied `image =`, or from a directory whose tissue positions file is missing or unreadable, is left exactly as it was.

## Tests

`tests/testthat/test_visium_array_coordinates.R`, against the fixture already in the repository: the columns are added and match the file spot for spot, the image's own coordinates are unchanged, a directory with no positions file is a no-op, and `Load10X_Spatial` produces them end to end (skipped where no HDF5 reader is installed).

Full suite `NOT_CRAN=true`: 1185 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

## If you would rather not

The alternative reading is that this is by design and the answer to that issue is a documented recipe:

```r
positions <- Read10X_Coordinates(Sys.glob(file.path(dir, "spatial", "*tissue_positions*"))[1], filter.matrix = TRUE)
object <- AddMetaData(object, positions[colnames(object), c("row", "col")])
```

I am happy to close this and send a documentation note instead; the loss is real either way, and the recipe is not obvious from the manual pages.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- `Load10X_Spatial` now keeps the array coordinates from the tissue positions file as `array_row`, `array_col` and `in_tissue` meta data. A `VisiumV2` image stores only the pixel position of each spot, so these were dropped on load, while a `VisiumV1` image kept all five columns in its `coordinates` slot ([#229](https://github.com/satijalab/seurat-object/issues/229))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
 - Fixed bug in `PercentageFeatureSet` where layer data was incorrectly retrieved prior to finding features with requested pattern ([#10438](https://github.com/satijalab/seurat/pull/10438))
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

PR #10477, opened 2026-08-31
Name the Xenium column that is missing

BenjaminDEMAILLE:
Addresses #8265, and the class of failure behind several other Xenium loading reports.

`ReadXenium()` selects the columns it needs by name:

```r
cell_info <- cell_info[, names(col.use)]
```

so a cells or transcripts file that does not carry them fails with

```
Error in `[.data.frame`(cell_info, , names(col.use)) : undefined columns selected
```

which names neither the file nor the column. Depending on which of the three readers (`arrow`, `data.table`, `read.csv`) got there first, the same input instead reaches `CreateCentroids()` and stops with `argument is of length zero`, which is exactly the traceback in #8265:

```
3: CreateCentroids.default(data$centroids)
2: CreateCentroids(data$centroids)
```

Neither message gives the user anything to check, which is why that thread ends in a suggestion to `debug()` the reader.

## Change

Check before selecting, and name the file and the columns:

```
The Xenium cells file has no 'x_centroid', 'y_centroid' columns. Columns found: cell_id

The Xenium transcripts file has no 'feature_name' column. Columns found: x_location, y_location
```

No behaviour changes for well-formed outputs: the same files read, and the same files fail, they just say why.

## Tests

`tests/testthat/test_readxenium_columns.R` writes Xenium-layout `cells.csv.gz` and `transcripts.csv.gz` files to a temporary directory, so no vendor data and no HDF5 reader is needed: a cells file with no centroids, a transcripts file with no feature names, and a complete pair still reading with the expected columns.

Two assertions fail without the change. Full suite `NOT_CRAN=true`: 1184 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- `ReadXenium` now names the file and the columns it lacks, instead of failing with \dQuote{undefined columns selected} when a cells or transcripts file does not carry the expected column names ([#8265](https://github.com/satijalab/seurat/issues/8265))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

PR #10478, opened 2026-09-01
Plot a repeated feature once, and say when cells have no identity

BenjaminDEMAILLE:
Two ways `DotPlot()` fails on input it should handle. Both found by sweeping the plotting functions with degenerate but ordinary arguments; `VlnPlot()` and `FeaturePlot()` handle the same inputs.

## A feature named twice

```r
> DotPlot(object, features = c(markers, markers[1]))
Error in `levels<-`: factor level [4] is duplicated
```

From

```r
data.plot$features.plot <- factor(x = data.plot$features.plot, levels = features)
```

The features are there only to fix the order of the axis, so they are taken unique. This happens as soon as marker lists are combined, and with a named list (the grouped dot plot) it is unavoidable when a gene marks two groups.

## Cells with no identity

```r
> DotPlot(object, features = markers)
Error in `rbind`: duplicate 'row.names' are not allowed
```

`CellsByIdentities()` returns a group for `NA`, padded with `NA` cell names, and those reach `FetchData()`. The per-group frames then fail to bind. Such cells are now dropped, with

```
Warning: 3 cells have no identity and are not shown
```

Both the cell names and the identity column are filtered, so this holds whether or not the padding itself is fixed. With satijalab/seurat-object#302 (which fixes the padding) but without this change, the same call still fails, differently:

```
Error: replacement has 1 row, data has 0
```

so the two are complementary rather than duplicates.

## Tests

`tests/testthat/test_dotplot_features_idents.R`: a repeated feature plotted once, a feature repeated across named groups, cells with no identity warned about and excluded, and the ordinary case unchanged.

Three assertions fail without the change. Full suite `NOT_CRAN=true`: 1188 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- Fixed `DotPlot` failing with \dQuote{factor level [n] is duplicated} when a feature is named twice, and with \dQuote{duplicate 'row.names' are not allowed} when some cells have no identity; those cells are now left out with a warning
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

PR #10479, opened 2026-09-01
Say when ctrl is larger than the bins it draws from

BenjaminDEMAILLE:
`AddModuleScore()` on a small object fails with a message from base R:

```r
> AddModuleScore(pbmc_small, features = list(rownames(pbmc_small)[1:20]))
Error: cannot take a sample larger than the population when 'replace = FALSE'
```

## Cause

The control features are drawn from the expression bin each feature of the set sits in:

```r
sample(x = data.cut[which(x = data.cut == data.cut[features.use[j]])],
       size = ctrl, replace = FALSE)
```

so a bin smaller than `ctrl` cannot supply them. Roughly, any object with fewer than `nbin * ctrl` features hits this on the defaults. `pbmc_small` is 230 features over 24 bins, about 9 per bin against a default `ctrl = 100`, so **every** call with default arguments fails on it. Small panels, heavily filtered objects and targeted assays are all in that range.

The message names neither `ctrl`, nor `nbin`, nor the object's size, so there is nothing to act on.

## Change

Check the bins that will actually be drawn from, and say what to change:

```
Cannot sample 100 control features: the smallest expression bin holding a feature of
the set has 9. Lower ctrl to at most 9, or lower nbin (currently 24) so that the 230
features are spread over larger bins
```

Still an error rather than a silent adjustment: lowering `ctrl` or `nbin` changes what the score means, so it is the caller's decision. `CellCycleScoring()` goes through the same path and gains the same message.

## Tests

`tests/testthat/test_addmodulescore_ctrl.R`: the message on defaults, scoring with a `ctrl` the bins can supply, and scoring with a smaller `nbin`.

Two assertions fail without the change. Full suite `NOT_CRAN=true`: 1185 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- `AddModuleScore` (and `CellCycleScoring`) now says when `ctrl` is larger than the expression bins it draws control features from, instead of failing with \dQuote{cannot take a sample larger than the population}; this is what a small or heavily filtered object hits with the default `ctrl = 100`
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

PR #10480, opened 2026-09-01
Say when there are not enough dimensions to work with

BenjaminDEMAILLE:
Four failures that come out of the decomposition, or out of a subscript, rather than from Seurat. Found by sweeping the reductions with degenerate but ordinary arguments.

| call | today |
| --- | --- |
| `RunPCA(one.feature)` | `max(nu, nv) must be positive` |
| `RunPCA(two.features)` | `non-conformable arguments` |
| `RunUMAP(object, dims = 1:50)` | `subscript out of bounds` |
| `RunTSNE(object, dims = 1:50)` | `perplexity is too large for the number of samples` |

The last is the worst: the reduction has 19 dimensions, `dims` asks for 50, and the message sends the reader after `perplexity`, which has nothing to do with it.

After:

```
Cannot compute principal components from 2 features: at least 3 are needed
More dimensions specified in dims than have been computed: 50 requested, pca has 19
```

`FindNeighbors()` already checks its `dims` this way and says exactly that; this is the same check for `RunUMAP()` and `RunTSNE()`.

`PrepDR()` also now stops when nothing survives the `scale.data` and variance filters, instead of handing the decomposition a matrix with no rows:

```
None of the requested features are in the scale.data layer with non-zero variance.
Run ScaleData() on them first
```

`PrepDR5()` already stops in the same situation, so this brings the v3 path in line.

No behaviour changes for calls that work: the same inputs succeed, and the same inputs fail, they just say why.

## Tests

`tests/testthat/test_dimension_guards.R`: PCA on one, two and three features; features that were never scaled; `dims` past the end for UMAP and tSNE; and `dims` within the reduction still running.

Six assertions fail without the change. Full suite `NOT_CRAN=true`: 1186 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@
 
 ### Fixes
 
+- `RunUMAP` and `RunTSNE` now say when `dims` goes past what the reduction holds, as `FindNeighbors` does, instead of failing with \dQuote{subscript out of bounds} or a misleading complaint about perplexity
+- `RunPCA` now says when there are too few features to compute components, and `PrepDR` stops when none of the requested features are scaled, instead of the decomposition failing with \dQuote{max(nu, nv) must be positive} or \dQuote{non-conformable arguments}
+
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

PR #10469, opened 2026-08-31
Recorrect the features and cells the counts assay still has

BenjaminDEMAILLE:
Fixes #9112, `subscript out of bounds` from `PrepSCTFindMarkers()`.

## Cause

An SCT model records the features and cells it was fit on, and subsetting the object does not prune that record. `my.correct_counts()` then indexes the counts assay with the model's features:

```r
umi <- raw_umi[valid_genes, cells, drop = FALSE]
```

so as soon as a feature named by the model is no longer in the counts assay, indexing fails:

```r
> sub <- subset(obj, features = rownames(obj)[1:50])
> PrepSCTFindMarkers(sub)
Found 2 SCT models. Recorrecting SCT counts using minimum median counts: 143
Error in .subscript.2ary(x, i, j, drop = TRUE) : subscript out of bounds
```

That is the error in the report, and it arrives at the same place, just as the progress bar starts.

## Fix

Intersect the model's features and cells with the counts assay before indexing it, and subset `cell.attributes` to match so it still lines up with the umi matrix column for column.

Recorrected values do not change: `correct_counts()` works per gene from that gene's fitted parameters, so a gene corrects identically whether or not its neighbours are present. The test asserts this, comparing against the full-object recorrection on the retained genes.

Features that the **SCT** assay carries but the counts assay does not are a different matter, since those come back as zeros rather than being skipped. That case now warns and names them instead of passing silently.

## Tests

Three cases, all on the existing two-model `sct_merged` fixture: features removed (the reported failure, plus equality with the full-object recorrection), and a model naming a cell the object no longer has, which is the same failure through the other index.

Both fail without the change and pass with it. Full suite `NOT_CRAN=true`: 1184 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

## Note

#10468 fixes a second, independent way `PrepSCTFindMarkers()` fails after subsetting. The two do not overlap; this one is the hard error, that one is the silent skip.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10481, opened 2026-09-01
Rename the cells an SCT model still shares with its assay

BenjaminDEMAILLE:
Fixes the `RenameCells` half of satijalab/seurat-object#256, where `RunAzimuth()` fails renaming a query object.

## Cause

`RenameCells.SCTAssay()` builds the mapping from the **assay's** cells and applies it to each **model's** cell attributes:

```r
old.names <- Cells(x = object)
names(x = new.names) <- old.names
...
rownames(x = x) <- unname(obj = new.names[old.names])
```

A model records the cells it was fit on, and subsetting the object does not prune that record, so a model can name cells the assay no longer has. Those look up as `NA`:

```
Error in `.rowNamesDF<-`(x, value = value): missing values in 'row.names' are not allowed
```

which is exactly the traceback in that issue:

```
11: FUN(X[[i]], ...)
10: lapply(X = cell.attributes, FUN = function(x) { ... })
 9: RenameCells.SCTAssay(object = object[[i]], new.names = new.cell.names[colnames(x = object[[i]])])
```

It also explains why passing `make.unique(Cells(object))` did not help: uniqueness was never the problem.

## Fix

Rename the cells that are in the mapping and leave the rest as they are. Dropping them would discard the record; renaming them is impossible, since they have no new name.

## Related

The same model-versus-assay drift is behind #10469 (`PrepSCTFindMarkers` failing with `subscript out of bounds`), and this is the same shape of fix.

The other error in that issue, `duplicate 'row.names' are not allowed`, comes from renaming to names that are not unique, which is a SeuratObject-level check; I will send that separately.

## Tests

`tests/testthat/test_rename_sct_cells.R`: renaming an object whose model names a cell the assay lost, asserting the assay's cells are renamed, the stale one is left alone and no row name is `NA`; plus ordinary renaming unchanged.

Three assertions fail without the change. Full suite `NOT_CRAN=true`: 1186 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- Fixed `RenameCells` on an SCT assay failing with \dQuote{missing values in 'row.names' are not allowed} when a model records cells the assay no longer has, which is what `RunAzimuth` hits on a subset object ([#256](https://github.com/satijalab/seurat-object/issues/256))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
 - Fixed bug in `PercentageFeatureSet` where layer data was incorrectly retrieved prior to finding features with requested pattern ([#10438](https://github.com/satijalab/seurat/pull/10438))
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

PR #10483, opened 2026-09-01
Say that a tree needs more than one identity

BenjaminDEMAILLE:
```r
> BuildClusterTree(subset(object, idents = 0))
Error in hclust(d = data.dist) : must have n >= 2 objects to cluster
```

Nothing in that mentions identities, and it comes from `hclust`, so the traceback does not help either. The object in this situation is usually a subset, where the identity count is precisely what the user needs to be told.

```
Cannot build a tree from 1 identity: at least two are needed. Set Idents() to a
grouping with more than one level, or run FindClusters() first
```

Found by sweeping with degenerate but ordinary arguments, not from an issue.

## Tests

`tests/testthat/test_cluster_tree.R`: the message on a one-identity subset, and two identities still building a `phylo` tree.

Two assertions fail without the change. Full suite `NOT_CRAN=true`: 1182 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- `BuildClusterTree` now says that at least two identities are needed, instead of `hclust` reporting \dQuote{must have n >= 2 objects to cluster}
+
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

Issue #10406, opened 2026-06-11
BUG: Potential issue in parallel memory usage


rbutleriii:
### Issue Description

Typically I have been running a `future(multicore, workers = threads)` setup for our pipeline on our cluster, and just letting Seurat handle it. I will typically set memory from SLURM divided by `threads` for  `future.globals.maxSize`, For a 100k cell dataset that would run this with 32 cores & 96GB mem (so 3GB per thread).
```r
  # split by group
  sc[["RNA"]] <- split(sc[["RNA"]], f = sc$group)
  sc

  # Prep & PCA on each group
  sc %>%
    NormalizeData() %>%
    FindVariableFeatures() %>%
    ScaleData(vars.to.regress = c("batch", "orig.ident")) %>%
    RunPCA(npcs = pcs) %>%
    RunUMAP(dims = 1:pcs, reduction.name = "umap", n.neighbors = n_neigh) %>%
    # feature selection and integration prep
    IntegrateLayers(
      method = HarmonyIntegration,
      new.reduction = "harmony",
      verbose = TRUE
    ) %>%
    RunUMAP(
      dims = 1:pcs, 
      reduction = "harmony", 
      reduction.name = "umap.harmony",
      n.neighbors = n_neigh
    ) %>%
    FindNeighbors(reduction = "harmony", dims = 1:pcs) %>%
    FindClusters(resolution = resi) %>%
    # combine again
    JoinLayers() -> sc

``` 

Was there a change to how memory is handled in the latest version? After updating to 5.5.0, it seems like it is using much more memory than before. Even during a single sample run (so maybe 10k cells), the same pipeline was chewing through memory, 16 cores required like 256GB memory? 

### Reproducing Code Example

```R
# Running 2 cores, 4 GB memory

library(Seurat)
library(SeuratData)
library(future)
library(magrittr)

plan(multicore, workers = 2)

LoadData("bmcite")
sc <- UpdateSeuratObject(bmcite)
pcs <- 30
resi <- 0.8

  # split by group
  sc[["RNA"]] <- split(sc[["RNA"]], f = sc$donor)
  sc

  # Prep & PCA on each group
  sc %>%
    NormalizeData() %>%
    FindVariableFeatures() %>%
    ScaleData(vars.to.regress = c("lane")) %>%
    RunPCA(npcs = pcs) %>%
    RunUMAP(dims = 1:pcs, reduction.name = "umap") %>%
    # feature selection and integration prep
    IntegrateLayers(
[...]

Issue #10238, opened 2025-12-11
FindIntegrationAnchors Hangs When Run in Parallel


mfurla:
Dear developers,

I'm having trouble running FindIntegrationAnchors in parallel mode because the code hangs. This occurs even when running ScaleData within the my.lapply loop; a step that takes only about 7 seconds in serialized mode.

I have tried using 4-10 CPUs, up to 150 GB of RAM (which should not be limiting, according to top), and the future plans multisession and cluster.

I am working on an HPC with the Slurm scheduler, and I have tried allocating resources both through --ntasks and --cpus-per-task options.

Any suggestions would be extremely valuable.

Best regards,

Mattia

samuel-marsh:
Hi Mattia,

Not member of dev team but hopefully can be helpful.  Could you please post the full code that you are running and contents on `sessionInfo` to understand situation better?  Also when you mean running in parallel are you running on series of Seurat objects at once?  If just a single object then running in parallel will not add any benefit as future is no longer supported for Seurat as of V5 (see [here](https://github.com/satijalab/seurat/issues/9385#issuecomment-2486214970)).

Best,
Sam

BenjaminDEMAILLE:
There is a concrete reason `FindIntegrationAnchors()` behaves badly in parallel, and it is measurable.

The function that finds anchors for a pair of objects is a closure created inside `FindIntegrationAnchors()`, and it refers to `object.list`. `future` therefore exports **every object being integrated to every worker**, whichever pair that worker was actually given. Intercepting the dispatch, on four objects totalling 22.3 MiB:

| | globals sent per task |
| --- | --- |
| today | 25.5 MiB, of which `object.list` is 21.7 MiB |
| with #10484 | 4.6 MiB, all of it functions; the pair itself travels as the task's data |

With a real integration, each worker serialises and receives the entire object list for every pair it computes. On an HPC node that is the picture you describe: memory is not exhausted, `top` looks calm, and nothing finishes, because the time goes into copying rather than computing. It also explains why more CPUs made it worse rather than better.

https://github.com/satijalab/seurat/pull/10484 sends each worker the pair it was asked about. Anchors are unchanged (same count, same indices, same scores).

**Until it lands**, the reliable workaround is to run the anchoring sequentially:

```r
future::plan("sequential")
anchors <- FindIntegrationAnchors(object.list = objects, dims = 1:30)
```

which loses less than it sounds, since most of the parallel time was going into moving data.

[...]

PR #10484, opened 2026-09-01
Send a worker the pair of objects it was asked about

BenjaminDEMAILLE:
The third and largest instance of the pattern in #10459 and #10471: work dispatched to `future` as a closure, so the whole dataset travels to every worker.

## Cause

`anchoring.fxn` is created inside `FindIntegrationAnchors()` and refers to `object.list`, so `future` exports **every object being integrated** to **every worker**, whichever pair that worker was given. Intercepting the dispatch, on four objects totalling 22.3 MiB:

| | globals per task | X per task |
| --- | --- | --- |
| before | 25.5 MiB, of which `object.list` is 21.7 MiB | ~0 MiB over 6 elements |
| after | 4.6 MiB, all of it functions | 14.4 MiB over 6 elements, one pair each |

Before, each task carries all four objects; after, it carries the pair it needs, and that stops growing with the number of objects. For the twelve- and fourteen-sample integrations people run, the difference is the whole list against two of it.

This is the shape behind #10406 (memory climbing with workers) and a plausible explanation for #10238 (integration that never finishes on a cluster): what looks like a hang is workers serialising and copying the entire object list per task.

## Change

- each object is trimmed with `DietSeurat()` **once**, in the parent, rather than once per pair, which is also less work overall (an object in `k` pairs was dieted `k` times);
- the per-pair work moves to a package-level `.AnchorsForPair()`, so what is exported is a function rather than a frame;
- both the sequential and the parallel path use it, so there is one code path for the work.

## Anchors are unchanged

Same object list, same seed, three objects:

```
anchors: 2254 | checksum: 2738940 | score sum: 1301.871      (before)
anchors: 2254 | checksum: 2738940 | score sum: 1301.871      (after)
```

and `IntegrateData()` on those anchors gives the same object (360 cells, 80 features).

## Tests

`tests/testthat/test_anchors_parallel.R`: anchors from two workers equal to anchors from a sequential plan, and the dispatched function being a package function rather than a closure (a closure carries the frame, which is the defect).

Full suite `NOT_CRAN=true`: 1181 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- `FindIntegrationAnchors` now sends each worker the pair of objects it is asked about, rather than exporting every object being integrated to every worker ([#10238](https://github.com/satijalab/seurat/issues/10238), [#10406](https://github.com/satijalab/seurat/issues/10406))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
 - Fixed bug in `PercentageFeatureSet` where layer data was incorrectly retrieved prior to finding features with requested pattern ([#10438](https://github.com/satijalab/seurat/pull/10438))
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

Issue #8385, opened 2024-01-27
Error in SCTransform on merged Seurat v5 object


ctlongg:
Hi,

I am recently processing an scRNA-seq dataset containing 14 samples. For each sample, I performed the standard QC filtering based on nFeatures and mitochondrial gene percentage, followed by the identification of doublet cells with DoubletFinder. After that, I merged the 14 samples and joined the layers while preserving only the merged `counts` layer. However, when I try to perform SCTransform on this merged object, I get the error `Error in bw.SJ(genes_log_gmean_step1) : need at least 2 data points` . I tried looking into the source code of SCTransform but couldn't locate whether there was wrong with my merged object or it was due to an incompatibility issue with Seurat v5. Below is the code for merging and SCTransform, thanks for the help.

```r
data <- readRDS("D:/workspace/scRNA-seq/merged_filtered.RDS")
data <- JoinLayers(data)
data[["RNA"]]$scale.data <- NULL
data[["RNA"]]$data <- NULL
data <- SCTransform(data, vars.to.regress = c("percent.mt"), verbose = FALSE)
```

saketkc:
Hi @ctlongg, I would need to take a look at the object in this case. Can you email your object to schoudhary@nygenome.org? Thanks!

denvercal1234GitHub:
Hi @saketkc - I also merge my individual Seurat objects (setting (`merge.data=T`)) that were SCTranform-ed individually. Do I need to perform `SeuratObject::JoinLayers()` now before proceeding with setting my variable_features with `SelectIntegrationFeatures`? 

This is because when I did `SeuratObject::JoinLayers(Merged_Seurat_Obj)` right after merging the different objects, I encountered an error `Error in UseMethod(generic = "JoinLayers", object = object) : no applicable method for 'JoinLayers' applied to an object of class "c('SCTAssay', 'Assay', 'KeyMixin')"`

Here is my merged object atm: 

<img width="1048" alt="Screenshot 2024-08-02 at 15 33 47" src="https://github.com/user-attachments/assets/56d2e655-0929-4233-b2c3-4662ff700743">

Thank you for your help.

denvercal1234GitHub:
Hi @saketkc @ctlongg @chungsai - Have you by chance found any solution/clarification regarding whether we need to perform `SeuratObject::JoinLayers()` before setting my `variable_features` with `SelectIntegrationFeatures()` ?

PR #10485, opened 2026-09-01
Check what SCTransform is being given

BenjaminDEMAILLE:
Two ways `SCTransform()` fails from inside `sctransform`, with messages that name neither the cells nor the features involved. Addresses #8385, and the recurring first one.

## A cell with no counts

`sctransform` models log10 of the counts per cell, which is `-Inf` for a cell that has none among the features given:

```
Error: cell attribute "log_umi" contains NA, NaN, or infinite value
```

This is one feature subset away at any time: `pbmc_small[1:8, ]` has 20 such cells out of 80. Nothing in the message says that cells are the problem, let alone which.

```
20 of 80 cells have no counts for the features given, such as 'GAACCTGATGAACC',
'TGACTGGATTCTCA', 'TGGTATCTAAACAG'. Remove them first, for example with
subset(object, subset = nCount_RNA > 0)
```

## Too few detected features

The variance stabilisation fits a density over the expressed features, and with too few of them:

```
Error in bw.SJ(genes_log_gmean_step1) : need at least 2 data points
```

which is what #8385 reports after merging and joining layers. The check here uses the `min_cells` that will be passed to `vst()`, so it agrees with what `sctransform` itself filters on:

```
Only 0 features are detected in at least 5 cells, out of 230. sctransform cannot fit a
model to that; lower min_cells, or use NormalizeData() for an object this sparse
```

## Scope

Both are checks on the input, before any work: an object that transforms today transforms identically, and an object that fails today still fails, with a message that says what to change.

## Tests

`tests/testthat/test_sctransform_input.R`: the cells message on a feature-subset object, the features message on a near-empty one, and an ordinary object still transforming.

Three assertions fail without the change. Full suite `NOT_CRAN=true`: 1185 passing, the only 2 failures being the pre-existing `Read10X_Image` ones fixed in #10454.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- `SCTransform` now says when cells have no counts for the features given, or when too few features are detected for a model to be fit, instead of failing inside `sctransform` with \dQuote{cell attribute "log_umi" contains NA, NaN, or infinite value} or \dQuote{need at least 2 data points} ([#8385](https://github.com/satijalab/seurat/issues/8385))
+
 - Fixed `AddModuleScore` (and `CellCycleScoring`) on v5 objects with on-disk (e.g. BPCells) assays, where each layer was fully densified to an in-memory `dgCMatrix` before scoring; scoring now operates directly on the on-disk matrix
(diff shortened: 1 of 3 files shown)


BenjaminDEMAILLE:

PR #10465, opened 2026-08-31
Give each Visium modality its own assay

BenjaminDEMAILLE:
Fixes #9182. **Builds on #10464** (rhdf5 fallback), which is its base branch; without that I could not read a Visium `.h5` at all on a current system.

## The bug

On a Gene + Protein panel, `Read10X_h5()` returns one counts matrix per modality, and `Load10X_Spatial()` passes that list straight to `CreateSeuratObject()`, which stacks them into a single assay. `nCount_Spatial` and `nFeature_Spatial` are then sums across modalities, and antibody counts run orders of magnitude above gene counts:

```
true gene UMIs      123, 126, 118
true antibody UMIs  39642, 39903, 40374
nCount_Spatial      39765, 40029, 40492      <- their sum
```

That matches the reporter's figures exactly: they see 746,750 where the gene total is 4,127.

Nothing indicates it. The object looks ordinary, `Assays()` shows one assay, and a metric everyone uses for filtering is wrong by a factor of several hundred. Any `subset(obj, nCount_Spatial > x)` on such an object is filtering on antibody signal.

## The change

Build the object from the primary modality, `Gene Expression` when present, and add the others as their own assays, as `LoadAkoya()` already does for multimodal input.

| | before | after |
| --- | --- | --- |
| assays | `Spatial` | `Spatial`, `Antibody.Capture` |
| `Spatial` rows | 48 (40 genes + 8 antibodies) | 40 |
| `nCount_Spatial` | 39765 | **123** |

The antibody counts are kept in their own assay rather than dropped, so nothing is lost and the standard CITE-seq style workflow applies.

Single-modality capture areas take the same path as before and are untouched.

## Compatibility

This changes results for anyone loading a Gene + Protein panel today: the object gains an assay, and `nCount_Spatial` falls to the gene-only figure. That is the point of the fix, but it should be called out in release notes, since a user's saved filtering thresholds will no longer mean the same thing. Arguably they never meant what the user thought.

## Tests

New `tests/testthat/test_spatial_multimodal.R`, which writes a Visium HDF5 from the repository's own `tests/testdata/visium/spatial` files, so no vendor data is needed: each modality gets its own assay, `nCount_Spatial` and `nFeature_Spatial` match gene expression alone, the antibody counts are present and correct in their assay, and a single-modality capture area is unchanged.

Against the base branch: 4 failures and 2 errors. With this change: 8 passes. Suite 548 passing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10460, opened 2026-08-31
Apply the Xenium transcript quality threshold again

BenjaminDEMAILLE:
Fixes #10155.

## The bug

`ReadXenium()` accepts `mols.qv.threshold`, documents it as removing low-quality transcript molecules, and `LoadXenium()` passes it through. It has no effect.

The reporter's reading of the code is exactly right: the quality column is not among those read, and nothing filters on it.

```r
col.use = c(
  x_location   = letters[24 + flip.xy],
  y_location   = letters[25 - flip.xy],
  feature_name = 'gene'
)
```

`grep -rn "mols.qv.threshold" R/` finds it only in the two signatures, the two doc blocks, and the pass-through from `LoadXenium()`. Never in a function body.

Nothing warns, so the result looks like filtered data. Anyone who has been relying on the documented default of 20 since the regression has been analysing unfiltered transcripts.

## Demonstration

Synthetic Xenium output, 200 molecules of which 60 are below the threshold:

| threshold | `main` | this PR |
| --- | --- | --- |
| 0 | 200 rows | 200 rows |
| 20 (the default) | **200 rows** | 140 rows |
| 40 | **200 rows** | 0 rows |

## The change

Read `qv` alongside the coordinates, drop the molecules below the threshold, then drop the column again so the returned frame keeps its current shape of `x`, `y`, `gene`. Nothing downstream sees a different structure.

Outputs without a quality column are handled both ways they can fail, since the three readers behave differently: `arrow` and `data.table::fread` select columns and error on a missing one, while `read.csv` returns a frame without it. Either way the transcripts are re-read without `qv` and a warning says the threshold was not applied, rather than filtering silently or failing on a missing column.

## Tests

New `tests/testthat/test_read_xenium_qv.R`, which writes a small synthetic Xenium directory rather than needing real output: thresholds of 0, 20 and 40 give 200, 140 and 0 molecules, different thresholds give different results, the returned columns are unchanged and carry no `qv`, an output without a `qv` column still loads and warns, and the documented default of 20 does filter.

Against `origin/main`: 5 failures. With this change: 9 passes. Full suite 536 passing; the 2 failures (`test_load_10X.R`) are present unchanged on `main` and are addressed separately in #10454. `R CMD check` Status OK.

## Note

This restores the documented behaviour. It does not revisit whether 20 is the right default, or whether `LoadXenium` should report how many molecules were dropped, both of which seem worth a separate look.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #10155, opened 2025-10-23
Regression: `ReadXenium` does no longer apply transcript QV filtering


ppm1337:
### Issue Description

The function `ReadXenium` has the argument `mols.qv.threshold` to filter transcripts based on the QV value reported in `transcripts.csv.gz`, `transcripts.parquet`, etc. However, looking into the current Seurat code base (v5.3.0) reveals that `mols.qv.threshold` is not used within the function body of `ReadXenium` and also not passed to other methods. Digging deeper, the code to apply the QV filtering used to be there, but was removed in commit 3dad59d824c3cb810f96c45dbe6773f2f8094d65 as far as I can tell.

Furthermore, the columns read from the `transcripts.*` files  (see `cols.use` variable) don't include the `qv` column anymore.

Since `LoadXenium` passes the argument in question to `ReadXenium`, the filtering is also not working for that function.

To give a concrete reprex to show that different values for `mols.qv.threshold` yield the same result:
After downloading the Xenium demo data used in the vignette with
```
wget https://cf.10xgenomics.com/samples/xenium/1.0.2/Xenium_V1_FF_Mouse_Brain_Coronal_Subset_CTX_HP/Xenium_V1_FF_Mouse_Brain_Coronal_Subset_CTX_HP_outs.zip
unzip Xenium_V1_FF_Mouse_Brain_Coronal_Subset_CTX_HP_outs.zip
```

I can read the `transcripts.csv.gz` manually using `data.table::fread("transcripts.csv.gz")`, to confirm that the `qv` values are in range 5.88 to 40. But filtering with `mols.qv.threshold = 41` gives the same output than filtering with `mols.qv.threshold = 0`:

```
so_41 <- ReadXenium(".", mols.qv.threshold = 41, outs = "microns")
so_0 <- ReadXenium(".", mols.qv.threshold = 0, outs = "microns")

all.equal(so_41, so_0)
#[1] TRUE
```

### Reproducing Code Example

```R
see above
```

### Error Message

```shell
No error.
```

### Session Info

```shell
R version 4.4.3 (2025-02-28)
Platform: x86_64-conda-linux-gnu
Running under: Ubuntu 24.04.3 LTS

Matrix products: default
BLAS/LAPACK: [...]/lib/libopenblasp-r0.3.30.so;  LAPACK version 3.12.0

locale:
[...]

PR #10489, opened 2026-09-01
Spatial reader fixes: Xenium, Akoya, Nanostring, Visium (replaces seven PRs)

BenjaminDEMAILLE:
**Draft, and grouped on purpose.** This replaces seven separate PRs, all in the spatial readers, so that one review covers one subject. I have closed the ones it supersedes: #10460, #10463, #10465, #10466, #10472, #10476, #10477. The individual commits are preserved, so a single one can be dropped without losing the rest.

Marked draft so it does not sit in your review queue until you say you want it.

## What is in it

| Commit | Problem |
| --- | --- |
| Xenium QV filter | The transcript quality threshold stopped being applied, a regression ([#10155](https://github.com/satijalab/seurat/issues/10155)) |
| Xenium single-feature-type panels | A panel carrying one feature type fails to load ([#9142](https://github.com/satijalab/seurat/issues/9142)) |
| Xenium column names | The missing column is now named ([#8265](https://github.com/satijalab/seurat/issues/8265)) |
| Akoya / QuPath centroids | Centroid columns are matched as QuPath actually writes them, and text coordinates are handled ([#7792](https://github.com/satijalab/seurat/issues/7792)) |
| `ReadNanostring` | The requested metadata column is kept, and the sparse path is tested |
| Visium modalities | Gene Expression and Protein each get their own assay, instead of one assay holding both and reporting wrong `nCount` ([#9182](https://github.com/satijalab/seurat/issues/9182)) |
| Visium array coordinates | The array coordinates a run records are kept ([satijalab/seurat-object#229](https://github.com/satijalab/seurat-object/issues/229)) |

## Note on HDF5

Two of the readers here need `.h5` files, which is what #10464 and #10467 are about. Those two are alternatives to each other and stay separate from this PR: #10464 adds `rhdf5` as a fallback, #10467 replaces `hdf5r` with it. Take either and I will close the other.

## Verification

Each commit carries a test that fails without its fix, several building their own vendor-format files so no sample data is needed. Full suite on the merged branch: 1254 passing, the only failures being the two pre-existing `Read10X_Image` ones that #10454 fixes.


YoukaiFromAccounting (commented):
Hi @BenjaminDEMAILLE - we are sequentially addressing the changes proposed in this PR in an attempt to provide more narrower fixes as necessary. In regards to the Xenium QV filter issue, which I believe to be pressing, I've opened a minimal PR here which you can take a look at: https://github.com/satijalab/seurat/pull/10507

Thank you for your interest in the Seurat project!

PR #10493, opened 2026-09-01
Hold and move matrices past the dgCMatrix 2^31 limit (replaces two PRs)

BenjaminDEMAILLE:
**Draft, and grouped on purpose.** This replaces #10453 and #10452, which are the two halves of the same problem: holding a matrix too large for `dgCMatrix`, and being able to move the object that holds it. Both are closed in favour of this one; the individual commits are preserved.

Goes with [satijalab/seurat-object#320](https://github.com/satijalab/seurat-object/pull/320), which carries the `StitchMatrix` method this needs.

## Why

`dgCMatrix` indexes its non-zeros with 32-bit signed integers, so a matrix stops being representable at `2^31 - 1` entries:

```
invalid class "dgCMatrix" object: 'i' slot is too large (> 2^31 - 1 elements)
```

Real data already crosses that line: the 10x GEM-X Flex 1M human K562 CRISPR aggregate has 4,355,024,689 non-zeros. Today BPCells is the only way past it, which means a dependency, a specific on-disk format, and objects that stop working when the directory moves.

## What is in it

**A 64-bit backend** on `DelayedArray`/`HDF5Array`, mirroring the existing `IterableMatrix` one. `as.DelayedMatrix(x, ondisk = FALSE, file = NULL)` wraps a matrix for use as a layer: in memory it gets an `SVT_SparseMatrix` seed, which stores non-zeros in a per-column tree rather than one flat index vector, so a layer can hold more than `2^31` of them and still `saveRDS`; on disk it is realized to a 10x-style `TENxMatrix`, whose `indptr` dataset is written with an 8-byte native HDF5 type. Backend methods keep the pipeline block-wise: `LogNormalize`, `VST`, `.CalcN`, `as.sparse`, and `DelayedMatrix` branches in `RowMeanSparse`/`RowSumSparse`/`RowVarSparse`.

**A materialization fallback that cannot be silenced.** Operations with no block-wise form (CLR and RC normalization, `mvp`/`dispersion` selection) load the layer into memory. Since someone reaches for this backend precisely when the data may not fit, that always emits a `warning()` naming the operation, the dimensions and the block-wise alternative. Deliberately not gated on `verbose`: I found this by testing my own first version, which was silent under `verbose = FALSE`.

**Portability.** `SaveSeurat`/`LoadSeurat` keep an object and its on-disk stores together, and `AsInMemory()` converts out-of-memory layers back to `dgCMatrix` so a plain `saveRDS` produces one self-contained file. This is what makes an object with on-disk layers shareable at all ([#9798](https://github.com/satijalab/seurat/issues/9798)).

**Two rough edges fixed on the way**, both of which affect BPCells today: `PseudobulkExpression` multiplies block-wise instead of erroring with \"does not support DelayedArray objects\", and `CheckMatrixSize()` reports where a matrix stands against the limit.

## Verification

Suite on the merged branch: 1253 passing, the only failures being the two pre-existing `Read10X_Image` ones that #10492 fixes. The 8-byte `indptr` claim is checked by reading the written HDF5 type back, not assumed.


BenjaminDEMAILLE:
Closing in favour of a series cut small enough to review: #10494 (the backend, 522 lines), #10495 (the materialization warning), #10496 (SingleCellExperiment conversion) and #10497 (CheckMatrixSize). Tracking issue with the order and dependencies: #10498.

Only #10494 is ready for review; the rest are drafts so they do not sit in the queue. The portability half of this PR is held back entirely rather than folded in, since it is a separate feature.

This is the same work as #10017, which sat unreviewed for thirteen months at 62 files and 100 commits. That was my fault for shape, not yours for pace, and this is the attempt to fix the shape.

PR #10503, opened 2026-09-10
Make RunPCA, RunUMAP and RunTSNE depend only on their seed

### LINKED ISSUE #10502: RunUMAP/RunPCA reproducibility: embeddings depend on the ambient RNG stream, thread counts, and unpinned uwot defaults
## Summary

The same script, on the same data, does not give the same UMAP on a different machine or after a package upgrade. Part of that is unavoidable floating-point drift, but several avoidable sources sit in Seurat itself: the embedding is a function of the ambient R random stream rather than of `seed.use`, thread counts and epoch counts are left to whatever the dependency happens to default to, `RunPCA`/`RunUMAP`/`RunTSNE` clobber the caller's random stream, and `RunPCA` returns components whose sign is arbitrary.

Below is an audit with measurements, including the things that turned out **not** to be causes, so nobody has to re-check them.

Environment: Seurat 5.5.1, R 4.6.1, uwot 0.2.4, RcppAnnoy 0.0.23, irlba 2.3.7, macOS 15 arm64 (Accelerate BLAS).

## Avoidable causes

**1. `RunUMAP` drives uwot through the global random stream instead of passing a seed.**

[`RunUMAP.default`](https://github.com/satijalab/seurat/blob/main/R/dimensional_reduction.R) calls `set.seed(seed.use)` and then `uwot::umap()` without `seed=`. The embedding is therefore a function of the R RNG stream position, which makes two things part of the contract that shouldn't be:

- the number of variates uwot draws is load-bearing. A uwot release that draws a different number of variates before the optimization moves every embedding for the same `seed.use`, with no visible change in Seurat.
- `seed.use = NULL` is documented as supported, and under it the embedding depends on whatever random code ran earlier in the session.

uwot >= 0.2.0 has an explicit `seed=` argument for exactly this. It is free to adopt:

```r
X <- matrix(rnorm(3000 * 20), nrow = 3000)
a <- {set.seed(42); uwot::umap(X, n_neighbors = 30L, metric = "cosine", min_dist = 0.3,
                               init = "spectral", n_threads = 1, n_sgd_threads = 1)}
b <- {set.seed(7);  uwot::umap(X, n_neighbors = 30L, metric = "cosine", min_dist = 0.3,
                               init = "spectral", n_threads = 1, n_sgd_threads = 1, seed = 42)}
max(abs(a - b))
#> 0
```

**2. `n_sgd_threads` is never pinned.**

Seurat relies on uwot's default of `0`. Nothing in Seurat states the requirement, so a change to that default silently makes the embedding irreproducible. It already is irreproducible whenever `uwot.sgd = TRUE`, which Seurat documents without a reproducibility caveat:

```r
set.seed(1); X <- matrix(rnorm(5000 * 30), nrow = 5000)
run <- function() {
  set.seed(42)
  uwot::umap(X, n_neighbors = 30L, metric = "cosine", min_dist = 0.3,
             init = "spectral", n_threads = 8, fast_sgd = TRUE)
}
max(abs(run() - run()))
#> 5.139      # and 6.145 on a previous invocation
diff(range(run()))
#> 6.37       # i.e. points move by about the full width of the embedding
```

**3. `n_threads = future::nbrOfWorkers()` is hard-coded into the uwot call.**

A machine-dependent quantity (core count, active `future` plan) is baked into the call, and no argument lets a user pin it. It happens not to affect today's output, verified below, but that is a property of uwot's current neighbor search, not something Seurat states or tests.

**4. `n.epochs` is left `NULL`.**

The epoch count comes from uwot's rule (500 for <= 10,000 observations, 200 otherwise). If that rule changes, every embedding changes.

**5. `RunPCA`, `RunUMAP` and `RunTSNE` clobber the global random stream.**

Each calls `set.seed(seed.use)` and never restores it, so any random draw after them depends on which Seurat functions ran earlier and in what order. This makes the reproducibility of a whole script depend on its call order rather than on its seeds.

**6. `RunPCA` returns components with an arbitrary sign.**

Truncated SVD singular vectors depend on the random start vector, on the BLAS/LAPACK, and on how many iterations the solver needed:

```r
X <- matrix(rnorm(2000 * 300), nrow = 2000)
a <- {set.seed(42); irlba::irlba(X, nv = 10)}
b <- {set.seed(7);  irlba::irlba(X, nv = 10)}
max(abs(a$d - b$d))
#> 9.78843e-09                                    # same subspace
sign(diag(crossprod(a$u, b$u)))
#> 1 1 -1 1 1 -1 -1 -1 -1 -1                      # 6 of 10 components flipped
```

scikit-learn, and therefore scanpy, canonicalizes this in `svd_flip`. Seurat's own `RunCCA` already does the equivalent; `RunPCA` does not.

To be precise about the blast radius: a sign flip applied to a whole component preserves every pairwise Euclidean and cosine distance, so this does **not** change `FindNeighbors`, `FindClusters`, `RunUMAP` or `RunTSNE` (verified identical end to end on `pbmc_small`). It does change the sign of loadings, the orientation of `DimPlot(reduction = "pca")`, `VizDimLoadings`, and anything that reads a PC coordinate's sign.

## Measured non-causes

Recording these so they don't get re-investigated:

- **annoy index construction.** `AnnoyBuildIndex` never seeds the index, but RcppAnnoy's default RNG state is fixed. Two builds over a 4000 x 20 matrix give identical 20-NN lists, 0 differing entries out of 80,000.
- **`FindNeighbors` / SNN construction.** Identical across runs and across ambient RNG state, 0 differing entries out of 9,000,000 on a 3000-cell input.
- **BLAS thread count.** `irlba(nv = 50)` on a 3000 x 300 matrix gives an identical `u` with 1 vs 8 BLAS threads on this machine.
- **uwot `n_threads`.** No effect on the embedding: `n_threads` 1 vs 4 vs 8, `max(abs(diff))` = 0.
- **Small input perturbations.** Perturbing one matrix entry by a relative 1e-15 left the embedding bit-identical, so the optimizer is not as chaotic as its reputation suggests; the sources above are the ones that actually bite.

## What is out of reach

Bit-identical results across operating systems, CPU architectures, and BLAS/LAPACK implementations are not achievable for UMAP or t-SNE, in Seurat or anywhere else. Floating-point addition is not associative, so a different BLAS, a different vector width, or FMA contraction changes the last bits of a dot product; hundreds of SGD epochs amplify that into a visibly different layout. scanpy carries the same limitation. Any fix should say so plainly rather than imply a guarantee it cannot keep.

What *is* achievable, and what I think this issue should be closed against: for a given installation, the embedding depends only on the input and on `seed.use`, and on nothing else - not the core count, not the `future` plan, not the ambient random stream, not a dependency's default that nobody pinned.

## Proposal

- Pass `seed = seed.use` to `uwot::umap()` / `umap2()` / `umap_transform()`.
- Add an `n.sgd.threads` argument, pinned to 1 by default; warn when `uwot.sgd = TRUE` makes the result irreproducible.
- Add an `n.threads` argument so `nbrOfWorkers()` is a default rather than a hard-coded value.
- Pin `n.epochs` to the value uwot would currently pick, so a change to uwot's default cannot move existing embeddings.
- Restore the caller's `.Random.seed` on exit from `RunPCA`, `RunUMAP` and `RunTSNE`.
- Apply the scikit-learn `svd_flip` sign convention in `RunPCA`.
- Add regression tests for each, including the non-causes above so they stay non-causes.

Every one of these except the `RunPCA` sign convention is bit-for-bit output-preserving against Seurat 5.5.1 (verified). The sign convention flips the sign of some PCs; since downstream distances are invariant, nothing but PCA plots and loadings changes.

I have this implemented and tested and will open a PR against this issue.

BenjaminDEMAILLE:
Fixes #10502.

## What this changes

`RunUMAP`, `RunPCA` and `RunTSNE` now depend only on their input and on `seed.use`, and not on the machine, the `future` plan, the ambient random stream, or a dependency default nobody pinned.

**`RunUMAP`**
- Passes `seed = seed.use` to `uwot::umap()` / `umap2()` / `umap_transform()` instead of relying on `set.seed()` plus whatever the stream position happens to be. `set.seed(42); umap(...)` is bit-identical to `umap(..., seed = 42)` (verified, `max(abs(diff))` = 0), so this costs nothing today and stops a change in how many variates uwot draws from moving every embedding tomorrow.
- New `n.sgd.threads`, pinned to 1. Seurat previously relied on uwot's default of 0 without stating the requirement.
- New `n.threads`, defaulting to `future::nbrOfWorkers()` as before, so a user can pin a value that was previously hard-coded to a machine-dependent quantity.
- Pins `n.epochs` to the value uwot currently chooses (500 for <= 10,000 observations, 200 otherwise) so a change to uwot's rule cannot silently move existing embeddings. Verified a no-op today for both `umap()` and `umap2()`.
- Warns when `uwot.sgd = TRUE`, which genuinely is irreproducible: two runs under the same `set.seed(42)` moved points by 5.1 units on an embedding spanning 6.4.

**`RunPCA`**
- Applies the scikit-learn `svd_flip` sign convention (largest-magnitude entry of each loading vector forced positive), the same one scanpy inherits and the one `RunCCA` already uses in this file. Truncated SVD singular vectors have an arbitrary sign that depends on the random start vector, the BLAS, and the iteration count: on a 2000x300 matrix two `irlba` runs agreed to 9.8e-09 on singular values and disagreed on the sign of 6 of 10 components.

**All three**
- Restore the caller's `.Random.seed` on exit. Previously each called `set.seed(seed.use)` and left it, so any random draw after a reduction depended on which reductions had run before it and in what order.

New `deterministic` argument on all three, defaulting to `getOption("Seurat.deterministic", TRUE)`.

## Compatibility

| | vs Seurat 5.5.1 |
|---|---|
| `RunUMAP` | bit-identical, verified against the released build |
| `RunTSNE` | bit-identical (`num_threads = 1` is already Rtsne's default) |
| `RunPCA` | **signs of some PCs flip** |
| `FindNeighbors`, `FindClusters`, `RunUMAP`, `RunTSNE` downstream of `RunPCA` | identical |

The PCA change is the only visible one. A sign flip applied to a whole component preserves every pairwise Euclidean and cosine distance, so the neighbor graph, the clusters and both embeddings come out identical; what changes is the sign of loadings, the orientation of `DimPlot(reduction = "pca")` and `VizDimLoadings`, and anything reading a PC coordinate's sign. Tested end to end on `pbmc_small`: 5 of 10 PCs flip, SNN graph identical, cluster assignments identical, UMAP identical, t-SNE identical.

`deterministic = FALSE`, or `options(Seurat.deterministic = FALSE)`, restores 5.5.1 output exactly, PC signs included.

## What this does not claim

Bit-identical results across operating systems, CPU architectures and BLAS/LAPACK implementations. Floating-point addition is not associative, so a different BLAS or vector width changes the last bits of a dot product and hundreds of SGD epochs amplify that into a visibly different layout. scanpy has the same limitation. The `RunUMAP` docs now carry a `Reproducibility` section that says this plainly instead of implying a guarantee the package cannot keep.

## Tests

Nine new tests in `test_dimensional_reduction.R` covering the sign convention, that it only flips signs, that it does not change downstream results, that `RunUMAP` is independent of the ambient RNG and of the worker count, the `uwot.sgd` warning, random-stream restoration for all three functions, and a regression test that `FindNeighbors` is reproducible (it already is; #10502 records the measurement so it does not get re-investigated).

Full `testthat` suite passes locally (R 4.6.1, uwot 0.2.4, macOS arm64), as does `R CMD check` for code/documentation consistency.


BenjaminDEMAILLE:
Heads up on the second commit, since it edits expected values in an existing test file.

The three SCT `FindTransferAnchors` tests in `test_integration.R` hardcode `pcaproject` embeddings and loadings, and the new `RunPCA` sign convention flips PC 1 for the SCT reference. All ten failures were exactly that:

- magnitudes agree to 10 significant digits
- anchor matrices unchanged: 256, 290 and 249 rows, identical scores
- `options(Seurat.deterministic = FALSE)` reproduces the old constants bit for bit

So the sign is the only thing that moved. Those constants were latent cross-platform fragility rather than a property of the data, since the sign `irlba` returns depends on its start vector and iteration count and a different BLAS could flip them on its own. Pinning them to the canonical sign makes the assertions stable rather than lucky.

The block is gated on `NOT_CRAN`, so reproducing locally needs `NOT_CRAN=true`. Full suite under that flag: 1207 expectations, 0 failures across all 17 files. CI is green.

PR #10504, opened 2026-09-10
Fix `FindSpatiallyVariableFeatures` cell lookup for `FOV` spatial images

### LINKED ISSUE #7020: error in FindSpatiallyVariableFeatures

### LINKED ISSUE #10099: BUG: Segfault in FindSpatiallyVariable Features

anashen:
This PR proposes changes based on https://github.com/satijalab/seurat/pull/10457, but with a narrower scope.

# Summary

`FindSpatiallyVariableFeatures()` assumed `rownames(spatial.location)` were cell names. For FOV-backed data such as CosMx/Xenium/Vizgen-style objects, GetTissueCoordinates() can return coordinates like:

|     | x   | y | cell |
|-----|-----|---|------|
| 1   | ... |   |      |
| 2   | ... |   |      |
| ... |     |   |      |

(That is, with integer row names and the real cell IDs in the cell column.) The `Assay`/`StdAssay` method then used row names like 1, 2, 3 to pull expression data, found no matching cells, produced a matrix with zero columns, and passed that to RowVar(), which could segfault.

Both this PR and 10457 address the core bug by using the cell column as coordinate row names before moving down the method dispatch. There are also two small fixes in both PRs that address other issues:  resolving the `selection.method` argument with `match.arg()` and fixing the wrong `nfeatures = nfeatures` default in `FindVariableFeatures.Assay/StdAssay` to `nfeatures = 2000`.

This branch proposes an additional safeguard on the cell column prior to using it as rownames: we check for duplicates, which occur when the default boundary for the selected image as segmentation-based. In that case, multiple coordinate rows can refer to the same cell (where each row represents coordinates for one vertex of the segmentation polygon). If this is true, users are directed to choose a centroid-based boundary or provide `spatial.location`.

The test added to `test_spatial.R` covers the assay-specific FOV case and also explicitly checks the segmentation boundary path.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: Seurat
-Version: 5.5.1.9004
+Version: 5.5.1.9005
 Title: Tools for Single Cell Genomics
 Description: A toolkit for quality control, analysis, and exploration of single cell RNA sequencing data. 'Seurat' aims to enable users to identify and interpret sources of heterogeneity from single cell transcriptomic measurements, and to integrate diverse types of single cell data. See Satija R, Farrell J, Gennert D, et al (2015) <doi:10.1038/nbt.3192>, Macosko E, Basu A, Satija R, et al (2015) <doi:10.1016/j.cell.2015.05.002>, Stuart T, Butler A, et al (2019) <doi:10.1016/j.cell.2019.05.031>, and Hao, Hao, et al (2020) <doi:10.1101/2020.10.12.335331> for more details.
 Authors@R: c(
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,9 @@
 - Fixed naming of combined p-value column in output of `FindConservedMarkers` when a non-default `meta.method` is specified ([#10429](https://github.com/satijalab/seurat/pull/10429))
 - Updated minimum required `uwot` version to `0.2.1` ([#10447](https://github.com/satijalab/seurat/pull/10447))
 - Registered `Radius.VisiumV2` as an S3 method to fix `Radius` returning `NULL` for `VisiumV2` images ([#10454](https://github.com/satijalab/seurat/pull/10454))
+- Fixed `FindSpatiallyVariableFeatures()` to correctly match FOV coordinates to cells in the requested assay and handle various edge cases, including fewer than two matched cells, a single varying feature, or no varying requested features ([#10504](https://github.com/satijalab/seurat/pull/10504))
+- Updated argument handling in `FindSpatiallyVariableFeatures()` by resolving `selection.method` with `match.arg()` and restoring the `FindSpatiallyVariableFeatures.Assay()` default `nfeatures` value to `2000` ([#10504](https://github.com/satijalab/seurat/pull/10504))
+- Fixed `RunMarkVario()` to return one named mark variogram result per feature for single-feature inputs and parallel execution chunks ([#10505](https://github.com/satijalab/seurat/pull/10505))
 
 # Seurat 5.5.1
 
--- a/R/preprocessing.R
+++ b/R/preprocessing.R
@@ -3650,15 +3650,28 @@ RunMarkVario <- function(
       f = ceiling(x = seq_along(along.with = features) / (length(x = features) / chunks))
     )
     mv <- future_lapply(X = features, FUN = function(x) {
-      pp[["marks"]] <- as.data.frame(x = t(x = data[x, ]))
-      markvario(X = pp, normalise = TRUE, ...)
+      # drop = FALSE so that a chunk holding a single feature stays a matrix
+      # and t() keeps one row per cell
+      pp[["marks"]] <- as.data.frame(x = t(x = data[x, , drop = FALSE]))
+      chunk <- markvario(X = pp, normalise = TRUE, ...)
+      # markvario() hands back a bare 'fv' for a single mark and a named list
+      # of them for several. Keep every chunk a list, so that the unlist()
+      # below contributes one entry per feature rather than one per column of
+      # an 'fv' whenever a chunk holds a single feature.
+      if (inherits(x = chunk, what = 'fv')) {
+        chunk <- list(chunk)
+      }
+      return(chunk)
     })
     mv <- unlist(x = mv, recursive = FALSE)
-    names(x = mv) <- rownames(x = data)
   } else {
     pp[["marks"]] <- as.data.frame(x = t(x = data))
     mv <- markvario(X = pp, normalise = TRUE, ...)
+    if (inherits(x = mv, what = 'fv')) {
+      mv <- list(mv)
+    }
   }
+  names(x = mv) <- rownames(x = data)
   return(mv)
 }
 
@@ -4756,14 +4769,22 @@ FindSpatiallyVariableFeatures.Assay <- function(
   data <- LayerData(object, layer = layer, cells = cells, features = features)
   data <- as.matrix(x = data)
   # RowVar() is C++ and aborts the session on an empty matrix rather than
-  # raising an R error, so catch any remaining path that produces one.
-  if (!ncol(x = data)) {
+  # raising an R error, so catch any remaining path that produces one. One
+  # column is caught here too: RowVar() divides by 'ncol - 1', so a single
+  # cell yields NaN for every feature, and the NA subscript that follows
+  # fails further down with a message that says nothing about the cause.
+  if (ncol(x = data) < 2) {
     stop(
-      "No cells were returned from the '", layer, "' layer.",
+      "Fewer than two cells were returned from the '", layer, "' layer; ",
+      "spatial variability is measured across cells and needs at least two. ",
+      "Check that the row names of 'spatial.location' are cell names.",
       call. = FALSE
     )
   }
-  data <- data[RowVar(x = data) > 0, ]
+  # Keep a single surviving feature as a one-row matrix; without drop = FALSE
+  # it becomes a vector and the call fails on 'attempt to set colnames on an
+  # object with less than two dimensions'.
+  data <- data[RowVar(x = data) > 0, , drop = FALSE]
   if (nrow(x = data) != 0) {
     svf.info <- FindSpatiallyVariableFeatures(
       object = data,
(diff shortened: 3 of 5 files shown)


BenjaminDEMAILLE (commented):
Thanks @anashen, and no objection to the narrower scope. Moving the `cell` column to row names before dispatch is the right place for the fix, and the duplicate check for segmentation boundaries is a good addition that #10457 did not have.

I checked out the branch and found three things worth a look. All three are pre-existing on `main` rather than regressions, but the first and third interact directly with the guard and the message this PR adds, so they seem in scope.

### 1. The `!ncol(data)` guard lets a one-column matrix through

`!any(cells %in% Cells(...))` only requires a single match, so a `spatial.location` with one real cell name and the rest bogus passes both new guards and then fails opaquely:

```r
n <- 12; cells <- paste0("cell", seq_len(n))
counts <- matrix(rpois(4 * n, 5), nrow = 4, dimnames = list(paste0("gene", 1:4), cells))
obj <- CreateSeuratObject(counts = as.sparse(counts)) |> NormalizeData(verbose = FALSE)

bad <- data.frame(x = runif(n), y = runif(n),
                  row.names = c(cells[1], paste0("bogus", 2:n)))
FindSpatiallyVariableFeatures(obj[["RNA"]], layer = "data", features = rownames(counts),
                              spatial.location = bad, selection.method = "moransi")
#> Error: argument is of length zero
```

`RowVar` divides by `ncol - 1`, so one column gives `NaN` for every feature, `NaN > 0` is `NA`, and the subset degenerates before `nrow()` is called. `ncol(x = data) < 2` instead of `!ncol(x = data)` would catch it and produce the clear message the guard is there to give.

### 2. `data[RowVar(x = data) > 0, ]` needs `drop = FALSE`

If exactly one feature survives the variance filter, `data` becomes a numeric vector:

```r
counts <- matrix(0, nrow = 4, ncol = 12, dimnames = list(paste0("gene", 1:4), cells))
counts["gene1", ] <- 1:12   # the only feature with variance
# ... same call as above with valid row names
#> Error: attempt to set 'colnames' on an object with less than two dimensions
```

One word on the line just below the new guard.

### 3. The new message points at an argument the `Seurat` method does not have

> Use `DefaultBoundary()` to select a centroid-based boundary or provide `spatial.location`.

A user who reaches this through `FindSpatiallyVariableFeatures.Seurat` cannot act on the second half:

```r
"spatial.location" %in% names(formals(getS3method("FindSpatiallyVariableFeatures", "Seurat")))
#> FALSE

FindSpatiallyVariableFeatures(obj, assay = "RNA", layer = "data", spatial.location = coords, ...)
#> Error: formal argument "spatial.location" matched by multiple actual arguments
```

`.Seurat` always passes `spatial.location = tc` itself, so routing one through `...` collides. Since the check only fires from `.Seurat`, dropping that clause, or pointing at `object[[assay]]` instead, would keep the advice followable.

### On the `RowVar` guard from #10457

#10457 also returned `NA_REAL` from `RowVar` for fewer than two columns. I could not reproduce the abort locally: on my build (macOS arm64, clang) a zero-column matrix returns zeros rather than crashing, so I cannot claim the segfault in #10099 from here. Grepping the R sources, the only caller that can reach zero columns is this one, so the R-level guard does cover the known path and the C++ change is not a blocker. Worth considering as separate hardening, since `RowVar` is the one place here that can take the session down rather than raise an error.

Happy to open a small follow-up PR against your branch for 1 and 2 if that is useful.

BenjaminDEMAILLE:
Opened #10505 against this branch with the follow-up I offered, so you can merge it in or ignore it.

One correction to my review: I called the `drop = FALSE` item "one word on the line just below the new guard." That was wrong. It is necessary but not sufficient, and writing the reproduction turned up two more shapes that reach the same code, so the follow-up covers five rather than two. The extra ones:

- `markvario()` returns a bare `fv` for a single mark and a named list for several, and `ComputeRMetric()` iterates, so a single surviving feature failed with `$ operator is invalid for atomic vectors` even with `drop = FALSE`.
- The same variable shape corrupts `RunMarkVario()` under a parallel plan, silently: a chunk holding one feature contributes its four `fv` columns to `unlist()` instead of one entry, so three features over two workers come back as length 6 with names `g1, g2, g3, NA, NA, NA`. Five features over two workers split 3 and 2 and are fine, which is presumably why it went unnoticed.
- When no feature survives the variance filter, `svf.info` is `c()` and both selection branches index into it: `attempt to set colnames on an object with less than two dimensions`.

All five are reachable on `main`; only the first is downstream of your guard. Happy to split the parallel one out against `main` instead if you would rather keep this branch tight.

I left the third review point alone, the "provide `spatial.location`" half of the segmentation message pointing at an argument `FindSpatiallyVariableFeatures.Seurat` does not have, since that is wording on your new code.

anashen:
Thank you for your input @BenjaminDEMAILLE, I've folded in #10505 with some extra changes.

